PR #75, opened 2015-03-10
Allow you to expand/collapse tree elements

### LINKED ISSUE #63: Can't inspect state or props
See around 18.50 https://www.youtube.com/watch?v=kNVatRjnU7U . It used to expand with state and props information but does not work anymore.

jhollingworth:
In newer versions of Chrome you can't expand any tree element (e.g. I can't expand/collpse the state props). The cause was [this line of code which I'm guessing is using a depreciated API](https://github.com/sebmarkbage/blink-devtools/blob/master/Source/devtools/front_end/treeoutline.js#L943).

I'm guessing the correct thing to do would be to update blink-devtools to the latest however as a stop gap we can update ``TreeElement.prototype.isEventWithinDisclosureTriangle`` with its [latest
implementation](https://chromium.googlesource.com/chromium/blink/+/master/Source/devtools/front_end/ui/treeoutline.js). It's far from ideal but it solves the problem.

I also made the sections non-editable to avoid accidental double clicks

Fixes #63.

### DIFF
--- a/views/components/ObjectSidebarPane.js
+++ b/views/components/ObjectSidebarPane.js
@@ -46,7 +46,7 @@ ReactPanel.ObjectSidebarPane.prototype = {
 
     var section = new WebInspector.ObjectPropertiesSection(object, '', '', this._emptyPlaceholder, false, null, ReactPanel.EditableObjectPropertyTreeElement.bind(null, this.onedit.bind(this)));
     section.expanded = true;
-    section.editable = true;
+    section.editable = false;
     section.headerElement.addStyleClass("hidden");
     body.appendChild(section.element);
   },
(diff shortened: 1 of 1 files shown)


zpao:
cc @sebmarkbage

jhollingworth:
Any update on this, its a really annoying bug

facebook-github-bot-7:
Thank you for reporting this issue and appreciate your patience. We've  notified the core team for an update on this issue. We're looking for a response within the next 30 days or the issue may be closed.

jaredly:
Sorry we didn't act on this before :/ it's now superseded by the new version of the devtools which is a total rewrite.
http://facebook.github.io/react/blog/2015/08/03/new-react-devtools-beta.html

jhollingworth:
No worries, the new version is much better! Congrats on the release :)

PR #80, opened 2015-03-25
Avoid deprecated getDOMNode call

### LINKED ISSUE #78: Warning: getDOMNode(...) is deprecated
I'm getting lots of the following warning: `Warning: getDOMNode(...) is deprecated in plain JavaScript React classes.`

voideanvalue:
Fixes #78

### DIFF
--- a/injected/Overlay.js
+++ b/injected/Overlay.js
@@ -32,7 +32,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
-(function() {
+(function(ReactHost) {
 
 var inspectorIframe;
 var inspectorOverlayPage;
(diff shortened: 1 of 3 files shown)


zpao:
Hmm, I wish we had exposed `React` in the `__REACT_DEVTOOLS_GLOBAL_HOOK__`. We really should go to using `findDOMNode`. Otherwise we should probably just take this and fix it better later. cc @sebmarkbage

injected/Overlay.js
```diff
@@ -160,7 +160,7 @@ function getHighlightColor(color) {
 
 var Overlay = {
 
-  highlightComponentInstance: function(instance, name, config) {
```
sebmarkbage: This is intentionally abstracted this way so that other environments like React Native can override this whole thing and implement highlighting differently.

voideanvalue: Makes sense. Updated the diff.

Daniel15:
When will this update be in the Chrome Web Store? I'm still seeing the getDOMNode warning with the latest released version.

voideanvalue:
@Daniel15: @sebmarkbage can probably answer that. But in the meanwhile you can clone this repository and use `Load unpacked extension...` in `chrome://extensions/`.

spicyj:
Just deployed as version 0.13.1; should be at https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi within an hour or so if you're impatient, and autoupdate sometime after that.

Daniel15:
Thanks! :smile:

PR #81, opened 2015-04-02
Add plugin to monitor the updates of react components 

hedgerwang:
Add plugin (codename: bananaslug) to monitor the updates of react components

The plugin renders a checkbox inside the current React panel. Clicking the checkbox would toggle the the plugin.

Most plugin codes lives inside `plugins/bananaslug/` and they're built with separate modern stacks (webpack, eslint, babel...etc) to ensure better developer experiences.


hedgerwang:
ping :)

hedgerwang:
ping~

hedgerwang:
ping :+1:

spicyj:
(@sebmarkbage)

sebmarkbage:
Sorry for letting this sit. This is really high value. We don't need to treat it as a plugin.

I think we need to update the copyright headers to use the new format.

What can we do to integrate the injected hooks with the main react-devtools hook? Currently we are often blocked on updating React because we also need to update all the places things introspecting internals here. I would prefer not to fork that code since it might mean we have to disable this plugin to be able to ship React. Integration would be better.

hedgerwang:
Hi @sebmarkbage,

`plugin` is just the name for this feature, we could use whatever name that you prefer.

What really matter is that this shall set up an example to hook into the existing code based with its own build system. I need this because unfortunately the main extension itself isn't modernized enough.

hedgerwang:
legal notes updated:
https://github.com/hedgerwang/react-devtools/commit/3ae0e17c3798769295e9d2932f94b339576a699e

hedgerwang:
ping :)

sebmarkbage:
If we update the hook passively, it is likely that it will be continuously broken internally because we run off of master. We would like to on a better solution for H2. Let me see what is the smallest viable thing I can come up with to get this in.

plugins/bananaslug/src/content_scripts/content-script-main.js
```diff
@@ -0,0 +1,60 @@
...
+
+  ScriptInjector.subscribe(
+    MessageType.ENABLED_STATE_CHANGE,
+    onEnabledStateChange
+  );
+
+  ScriptInjector.inject('injected-prelude');
+}
+
+main();
```
sebmarkbage: This file is much too heavy to execute on every page. Anyone that has the devtools installed will execute this on every page they ever visit.

plugins/bananaslug/src/content_scripts/ScriptInjector.js
```diff
@@ -0,0 +1,116 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * The ScriptInjector file.
+ */
+
+var Fetcher = require('./Fetcher');
```
sebmarkbage: We shouldn't be using XHR requests on pages that we inspect. This creates origin problem, potential security issues and screws up debugging of the page. (E.g. network requests popping up in the network tab.)

plugins/bananaslug/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+clear;
+
+echo 'build plugin "bananaslug"';
+
+# clean up old files
+rm build/*
```
stopachka: the first time, build is not a directory -- we would need to create if it didn't exist

manifest.json
```diff
@@ -15,12 +15,19 @@
...
+    "plugins/bananaslug/build/*"
+  ],
 
   "content_scripts": [
     {
       "matches": ["<all_urls>"],
-      "js": ["injected/GlobalHook.js"],
+      "js": [
+        "injected/GlobalHook.js",
+        "plugins/bananaslug/build/content-script.bundle.js"
```
stopachka: if we add this, then we would need to update the main README, to say that they must run ./build.sh, or content-script.bundle will not exist

stopachka:
+1 for this plugin, looking forward to using it! :D

plugins/bananaslug/src/content_scripts/content-script-main.js
hedgerwang: I'll see if there is a way to disable it when the inspector is hidden.

PR #85, opened 2015-05-09
Prevent DOM explorer from flickering

### LINKED ISSUE #73: Flickering when opening
Hi there,
I have been experiencing a issue with the DevTools flickering when the Reactis selected.  I have tried the plugin in both standard Chrome Version 41.0.2272.76 (64-bit) and Chome Canary Version 43.0.2327.5 canary (64-bit) with exactly the same results.

It looks like others are experiencing this issue as well:
https://code.google.com/p/chromium/issues/detail?id=248966

The video on post #3 looks pretty much identical to what I am seeing.

Casey

dariocravero:
This PR fixes #73.

Thanks to @LeZuse [for the hint](https://github.com/facebook/react-devtools/issues/73#issuecomment-95767146). :)

### DIFF
--- a/injected/Overlay.js
+++ b/injected/Overlay.js
@@ -183,7 +183,10 @@ var Overlay = {
       element = null;
     }
     if (element == null) {
-      DOMHost.hideHighlight();
+      try {
+        DOMHost.hideHighlight();
+      } catch(x) {
+      }
       return;
     }
 
@@ -220,7 +223,9 @@ var Overlay = {
   },
 
   hideHighlight: function() {
-    inspectorIframe.style.display = 'none';
+    if (inspectorIframe) {
+      inspectorIframe.style.display = 'none';
+    }
     inspectorIframeVisible = false;
   }
 


zpao:
Thanks!

Cc @sebmarkbage.

nw:
This does remove the flicker.

Still have the following issues:

* Selections inside the render tree (dark blue highlight) does not work. Can't see Props, State & Component context.
* Expansion inside Props, State & Component views doesn't work.
* Arrows to expand / collapse children nodes only expand after a double click (single click prior). Collapsing doesn't work.

I noticed that the blink dependencies are old. I brought in the latest files from the blink project, this does not break the extension build. However, the api for the inspector and the runtime that loads it have changed drastically.

I don't have time to further investigate / fix at the moment. Will take a look later this week if no one else gets to it.

Tested on:

* 44.0.2397.0 canary (64-bit)
* 43.0.2357.52 beta (64-bit)

spicyj:
Went with an alternate approach in #87. Thanks for sending this in though.

PR #87, opened 2015-05-12
Fix flickering

### LINKED ISSUE #73: Flickering when opening

spicyj:
Blink (or V8?) changed the exception format so this old check broke. This should be more stable. Fixes #73.

### DIFF
--- a/agents/ReactInspectorAgent.js
+++ b/agents/ReactInspectorAgent.js
@@ -38,6 +38,8 @@ var ReactInspectorAgent;
 
 (function() {
 
+var MISSING_RUNTIME =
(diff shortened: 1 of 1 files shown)


sebmarkbage:
:thumbsup:

Issue #8, opened 2014-01-04
React tab is blank when third-party cookies are blocked


ivan:
If I use React Developer Tools in a Chrome profile that has "Block third-party cookies and site data" enabled in Content settings, the contents of the React tab are completely blank.  Inspecting the inspector shows these errors:

```
Uncaught SecurityError: Access to 'localStorage' is denied for this document. Settings.js:131
Uncaught TypeError: Cannot read property 'domWordWrap' of undefined ReactPanel.js:67
```

The errors are thrown from these lines:

https://github.com/sebmarkbage/blink-devtools/blob/120a03721f07ecab689e5aa2444ed26094e5ede0/Source/devtools/front_end/Settings.js#L131

https://github.com/facebook/react-devtools/blob/9879bdc7a6483bca77d83c6e93e6768db1d93dc6/views/components/ReactPanel.js#L67

The relevant Chrome bug is https://code.google.com/p/chromium/issues/detail?id=319328#c4

If it isn't worth fixing, it might be worth detecting the problem and telling the user to disable third-party cookie blocking.

irae:
I found out a workaround. I added the chrome extension ID to the list of exceptions:

![screenshot 2015-07-01 at 11 29 50 am](https://cloud.githubusercontent.com/assets/98254/8461892/9c5145f0-1fe4-11e5-918a-1b87a20b46c6.PNG)

Looks like this will make `localStorage` available to the extension, without the need to change your privacy settings.

I agree with Sebastian that a better startup page would be great, and this workaround can actually be included as a warning message if such page exists in the future.

spicyj:
@irae Thanks, good find.

james-henderson:
The (copy-pasteable ;) ) chrome extension ID is `fmkadmapgofadopljbjfkapdkoienihi`

jaredly:
the newly released [beta version](http://facebook.github.io/react/blog/2015/08/03/new-react-devtools-beta.html) works just fine w/ third-party cookies blocked.
Let us know if you run into any other issues!

PR #100, opened 2015-07-08
Readme: add workaround to blocked 3rd-party cookies

denis-sokolov:
Per suggestion in the issue #8:
https://github.com/facebook/react-devtools/issues/8#issuecomment-117789730


jaredly:
The 3rd-party cookies issue (should be) now solved by the new beta version
http://facebook.github.io/react/blog/2015/08/03/new-react-devtools-beta.html
Let us know if you run into further issues

Issue #111, opened 2015-08-04
Global `Map` component breaks the backend


SoonDead:
I have installed the extension for a chrome and a firefox.

Whenever I try to open the react tab for testing my application I get the javascript error stated in the title for both browsers.

Here is the stack trace:

```
onMount	@	attachRenderer.js?dda8:123
walkNode	@	attachRenderer.js?dda8:163
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
(anonymous function)	@	attachRenderer.js?dda8:161
walkNode	@	attachRenderer.js?dda8:161
walkRoots	@	attachRenderer.js?dda8:153
extras.walkTree	@	attachRenderer.js?dda8:126
(anonymous function)	@	inject.js?9d71:30
(anonymous function)	@	VM63:46
emit	@	VM63:45
setupBackend	@	backend.js?a404:55
module.exports	@	inject.js?9d71:38
(anonymous function)	@	VM166:56
g	@	events.js?b340:169
EventEmitter.emit	@	events.js?b340:74
(anonymous function)	@	VM167:135
(anonymous function)	@	Bridge.js?1afd:284
(anonymous function)	@	Bridge.js?1afd:284
_handleMessage	@	Bridge.js?1afd:277
listener	@	VM166:33
```

jaredly:
Huh. Are you using any polyfills that might be interfering (rootNodeIDMap is an es6 Map)?

jaredly:
A quick way to test this: in your console, what do you get for `Map + ''`?

SoonDead:
I get the following output:

```
function (props, context) {
[...]

PR #113, opened 2015-08-04
fixes #104; makes the chrome ext immune to Object.create polyfills

### LINKED ISSUE #103: Firefox doesn't reload React tools when navigating

### LINKED ISSUE #104: [New Devtools] restrictive Object.create polyfill breaks the devtools

jaredly:
Here is a fairly simple way to insulate the chrome extension from `Object.create` polyfills.

The only place `Object.create` is used is in the babel helpers, so we render them externally and replace `Object.create` with `__REACT_DEVTOOLS_GLOBAL_HOOK__.nativeObjectCreate`, which is set by the content script *before* any page js has run, so before any polyfills.

There's certainly some hackiness to it, but I think it will be robust... does anyone have a more elegant solution?

What I'm least happy about is the need to have two separate webpack config files. It makes development a little more cumbersome because you need two webpack processes.
This could be fixed by having a node script that runs two webpacks at once though.

fixes #104


Daniel15:
Perhaps you could use a Babel plugin to tell it to not use Object.assign? cc @sebmck

Is there anything else in Babel that relies on native polyfills? Generally we can't assume anything about the environment while in the context of an extension.

sebmck:
I'm not sure I understand why `Object.create` can't just be used verbatim?

Daniel15:
The issue is that the page we're injecting scripts to could be loading an `Object.create` polyfill that doesn't behave properly. On some pages at Facebook we override `Object.create` and add an explicit check that only the first parameter is used, to ensure people don't use functionality that won't work in the IE polyfill. It has something like this in it:
```
if (arguments.length > 1) {
  throw new Error(
    'Object.create implementation supports only the first parameter'
  );
}
```

This is throwing when the React devtools are open, as it's calling this polyfilled code.

In general browser extensions can't really assume anything about a page's environment, as the page could be doing *anything* in its JavaScript.

spicyj:
> In general browser extensions can't really assume anything about a page's environment, as the page could be doing anything in its JavaScript.

You can say that but it's not actually practical to insulate your code against a hostile environment. What if Number.prototype.toString is changed? What if `document` is deleted from `window`? You can't win…

sebmck:
Opened #119 as an alternative.

jaredly:
re: hostile environment; the issue in https://github.com/facebook/react-devtools/issues/111 is that the page has a react class named `Map` in the global scope, which replaces the es6 `Map` class (－‸ლ)

Daniel15:
I don't think that'd be uncommon; lots of people would have Map components :P

Aren't extensions meant to be sandboxed? Can't you somehow get a "real" Map from the extension's context?

sebmck:
So this is kinda crazy but you _should_ be able to do:

```javascript
var iframe = document.createElement("iframe");
document.body.appendChild(iframe);
iframe.contentWindow.Object.create;
iframe.contentWindow.Map;
... // etc
```

Might be some weird cross realm issues though, worth a shot though maybe. (As bizarre of an idea as it is)

jaredly:
We need to be in the window's context (not sandbox) to mess w/ react objects/runtime.
We could get the "real" Map class in the same way I get the "real" object.create in this PR `GlobalHook.js`

jaredly:
@sebmck haha neat idea :)

spicyj:
@sebmck Better hope no one [overwrites `document.createElement`](https://github.com/facebook/react-native/blob/9f81d6726c507da0d647781d79e1c3415e844bc8/Libraries/JavaScriptAppEngine/polyfills/document.js#L28). :)

jaredly:
You can actually get around that (and I do in React Native) with `document.prototype.createElement` :)

Daniel15:
Until someone overwrites `document.prototype.createElement` :laughing:

sebmarkbage:
I actually think @sebmck's idea might be the most long term stable solution here. It is the whole point of the "Realms" proposal of ES6 to specify and make this easier to work with.

I wouldn't extract the methods though. I would run all the devtools code inside of the iframe's realm. That way you even protect against `Number.prototype.toString` and your array literals gets their own prototype chain etc.

Daniel15:
>  I would run all the devtools code inside of the iframe's realm.

This sounds like a great idea.

jaredly:
> I wouldn't extract the methods though. I would run all the devtools code inside of the iframe's realm.

This means we would have to leave an iframe hanging around somewhere attached to the DOM :( at least in the current chrome, the iframe's context (including timers, etc) is garbage collected when the iframe is removed from the DOM.
So we could add it to the `head` or the `documentElement` or something, but that feels a bit messy

sebmarkbage:
Seems like Babel could avoid using defineProperties in loose mode, or this polyfill could die with IE8.

If this is the only fix, just go for the hack in this PR for now.

@jaredly I'd be create if you could document the dependencies on the native environment so that we can predict if other features are going to cause problems. E.g. the document.prototype.createElement one is a bit crazy. What happens if we (have to) change it back? Do we break older versions of React Native debugging?

jaredly:
For the chrome extension, we only need to support chrome :) getting IE8 debugging would be interesting though.

Current versions of React Native set `document.createElement = null`, so in order to support them we would need to keep the `createElement` weirdness.
I'll add info to the Readme

PR #116, opened 2015-08-04
[chrome] hide highlight when switching devtools tab

jaredly:
The Elements pane will hide its highlight if you go to a different devtools tab, and I think that behavior makes more sense.

This is what happens w/ this PR:

chrome-highlight

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -288,11 +288,19 @@ class Store extends EventEmitter {
       this.emit('hover');
       this._bridge.send('highlight', id);
     } else if (this.hovered === id) {
-      this.hovered = null;
-      this.emit(id);
-      this.emit('hover');
-      this._bridge.send('hideHighlight');
+      this.hideHighlight();
     }
+  }
+
+  hideHighlight() {
+    this._bridge.send('hideHighlight');
+    if (!this.hovered) {
+      return;
+    }
+    var id = this.hovered;
+    this.hovered = null;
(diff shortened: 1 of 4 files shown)


frontend/Store.js
```diff
@@ -288,13 +288,21 @@ class Store extends EventEmitter {
...
-      this.hovered = null;
-      this.emit(id);
-      this.emit('hover');
-      this._bridge.send('hideHighlight');
+      this.hideHighlight();
     }
   }
 
+  hideHighlight() {
+    this._bridge.send('hideHighlight');
```
spicyj: What does this line do?

jaredly: It sends a message to the agent (living on the page) to hide the highlight

spicyj: Got it, of course. Does anything listen to the `emit('hover')`?

jaredly: Yeah -- [here's the file](https://github.com/facebook/react-devtools/blob/devtools-next/frontend/Highlighter/setup.js) used to connect a `Highlighter` to an `Agent`

spicyj: I think you misread my question?

jaredly: oh yes! I totally did. No, nothing listens for hover at the moment. Would you rather I remove it?

spicyj: Either way. (accepted)

PR #117, opened 2015-08-04
Firefox load fix

### LINKED ISSUE #103: Firefox doesn't reload React tools when navigating

jaredly:
It works! fixes #103

firefox-load


jaredly:
@frantic could you check this out?

PR #118, opened 2015-08-04
fix "show elements pane" in chrome extension #109

### LINKED ISSUE #109: [New DevTools] Show in Elements Pane not working in Opera
Hi,

I’m using the new beta Chrome extension in Opera (v30). When I right-click on a node within the React DevTools, it shows the option "Show in Elements Pane", but clicking it doesn’t appear to do anything. The other options, such as "Show Source" work okay though.

jaredly:
and add more flowtypes such that similar errors would be automatically detected

fixes #109

### DIFF
--- a/shells/chrome/src/panel.js
+++ b/shells/chrome/src/panel.js
@@ -13,6 +13,8 @@
 var checkForReact = require('./checkForReact');
 var inject = require('./inject');
 
+import type {Props} from '../../../frontend/Panel';
+
 type Listenable = {
   addListener: (fn: (message: Object) => void) => void,
 }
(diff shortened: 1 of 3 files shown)


shells/chrome/src/panel.js
```diff
@@ -56,7 +59,7 @@ var config = {
     chrome.devtools.inspectedWindow.eval('window.__REACT_DEVTOOLS_GLOBAL_HOOK__.$0 = $0');
     bridge.send('checkSelection');
   },
-  selectElement(bridge, id) {
+  selectElement(id, bridge) {
```
jaredly: this is where the error was

jaredly:
@frantic could you check this out?

PR #119, opened 2015-08-04
Use babel-plugin-object-create-multiple-to-single-and-define-propery

### LINKED ISSUE #104: [New Devtools] restrictive Object.create polyfill breaks the devtools

sebmck:
_Should_ fix #104.

This plugin replaces:

```javascript
Object.create(foo, {});
```

with:

```javascript
var _temp;
_temp = Object.create(foo), Object.defineProperties(_temp, {}), _temp;
```

This is pretty hacky, it only just gets around the error for `Object.create`.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,1 +1,4 @@
-{"stage": 0}
+{
+  "stage": 0,
+  "plugins": ["object-create-multiple-to-single-and-define-property:after"]
+}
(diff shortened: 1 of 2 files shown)


sebmck:
re the plugin name: I'm terrible at names and decided to summon my inner Java.

jaredly:
haha that's excellent

jaredly:
opted to use the other workaround

sebmck:
It was worth it just for the package name!

PR #121, opened 2015-08-05
fix styling of "looking for react" message

### LINKED ISSUE #31: Design a default screen when no React instance is detected
Currently we don't show the React tab if we can't detect React. We should have a better default screen that explains how you can get React and how to make it work with DevTools on your page.

jaredly:
image

Fixes #31


spicyj:
Did you mean #31?

jaredly:
yeah

PR #137, opened 2015-08-10
run the backend in an iframe in the <head>

### LINKED ISSUE #104: [New Devtools] restrictive Object.create polyfill breaks the devtools

jaredly:
an alternative to #113, fixes #104 

This runs all the of the backend code that needs to be in the same context as the React runtime in an `iframe` that is added to the `<head>` of the page.

#### Why does the iframe need to be on the DOM?
At least in chrome, iframes that are not attached to the DOM are garbage collected - such that all listeners, timers, etc. from code executed in the iframe are invalid.

See #113 for other possible solutions

### DIFF
--- a/frontend/Highlighter/setup.js
+++ b/frontend/Highlighter/setup.js
@@ -14,8 +14,8 @@ var Highlighter = require('./Highlighter');
 
 import type * as Agent from '../../agent/Agent';
 
-module.exports = function setup(agent: Agent) {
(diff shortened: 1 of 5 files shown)


spicyj:
I'd guess that if you hold a reference in JS then it won't be GCed?

jaredly:
References stick, but listeners and timers are killed

Issue #120, opened 2015-08-04
[dt-next] Move to the search input to the top, and highlight it when a search query is entered


nt1m:
- Moving it to the top will make it more discoverable
- Highlighting the search box will make the UX less confusing if the markup view is filtered. See Chrome/Firefox DevTools' filter styles input for example.

jaredly:
A wanted to keep it out of people's faces b/c it seems like a secondary use case.
Open to discussion though.
Highlighting it is a great idea :+1:

jaredly:
@nt1m what do you thing of #141?

nt1m:
@jaredly Unfortunately, I'm getting a bunch of "permission denied" errors trying to `npm install`

kassens:
We highlight the search box now (thanks for the idea) and I think the argument of secondary functionality for filtering is convincing.

PR #141, opened 2015-08-11
highlight the search bar when active, and make the font bigger. re #120

jaredly:
In response to #120

When active:
image

When inactive:
image


spicyj:
Sure?

PR #142, opened 2015-08-12
Fix search results

jaredly:
Before this change, both `Native` and `NativeWrapper` components were showing up in listings under 0.13.
We want to only display NativeWrappers in 0.13, but in 0.14 there are no native wrappers, so we just show the native components.

Before:
image

After:
image

(half as many native components shown)

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -353,11 +353,7 @@ class Store extends EventEmitter {
     if (up) {
       return this._parents.get(id);
     }
-    var children = node.get('children');
-    if ('string' === typeof children) {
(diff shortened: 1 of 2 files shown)


spicyj:
Why can't we always show the Native component, sorry? I'm failing to remember what NativeWrappers are at all.

jaredly:
It's the 0.13 wrapper around divs, etc.
We can't just show the native component, b/c they have immutable props, and the current devtools allow props mutation for native components in 0.13

spicyj:
Oh, on web. All right.

PR #143, opened 2015-08-12
add a breadcrumb trail to fix #133

### LINKED ISSUE #133: Show breadcrumbs like in Elements pane
In complex structures it's sometimes hard to navigate around, would be nice to have breadcrumbs at the top/bottom like in the Elements pane like `Component 1 > Component 2 > etc`

jaredly:
fixes #133.
We also have a breadcrumb trail in the official version, so it makes sense to bring it back.
However, this breadcrumb trail reflects the **parent** hierarchy, whereas the old one showed the **owner** hierarchy. Owners are going away, so I used parents.
breadcrumbs

### DIFF
--- a/frontend/Breadcrumb.js
+++ b/frontend/Breadcrumb.js
@@ -1,0 +1,94 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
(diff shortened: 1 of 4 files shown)


kassens:
Looks good!

PR #145, opened 2015-08-13
Add a tabbed pane

jaredly:
It doesn't look super interesting yet, but I figured it would be better to do this in small chunks.
This is in preparation for the `Relay` integration.

I modeled the styling after chrome's "sources" tab, but it's whatever

image

### DIFF
--- a/frontend/Container.js
+++ b/frontend/Container.js
@@ -15,6 +15,7 @@ var PropState = require('./PropState');
 var React = require('react');
 var SearchPane = require('./SearchPane');
 var SplitPane = require('./SplitPane');
+var TabbedPane = require('./TabbedPane');
 
 import type MenuItem from './ContextMenu';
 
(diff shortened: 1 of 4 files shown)


kassens:
How do you think this would work inside of Chrome or Firefox's tools? I'm actively looking into the Relay integration now and it seemed better to create a second Plugin and potentially communicate with the React devtools via postMessage.

jaredly:
Chrome's "sources" pane has a second tab bar, and this is styled after that. I admit it's not the prettiest :)
I felt like I wanted tighter integration powers -- the ability to add menu items, side panes, etc. and it would be a nightmare to try to version multiple extensions in parallel.

That said, we could just do multiple tabs from the same plugin. I think they might even all live within the same process, and we could have pretty simple interop API.

btw check out the [relay branch](https://github.com/jaredly/react-devtools/tree/relay) on my fork -- that's where I developed the relay plugin MVP

kassens:
Oops, I guess this was closed because I cleaned up the `facebook:devtools-next` branch. An updated version can be found in #186, so I'm leaving this closed.

Issue #129, opened 2015-08-08
Poor Performance in Chrome


timoxley:
It appears that after opening react-devtools, react-devtools itself becomes the most significant, negative influence on rendering performance:

![image](https://cloud.githubusercontent.com/assets/43438/9147877/45a809b6-3d33-11e5-8c15-c230057520a8.png)

In this case, react-devtools is blocking on the UI thread for 157ms which makes a very perceptible dent in the app's frames per second:

![image](https://cloud.githubusercontent.com/assets/43438/9147686/23a2d5ba-3d30-11e5-94fc-f6cfac6fa76d.png)

Even with the react-devtools tab closed there appears to be no way to stop this from happening without closing Chrome devtools entirely, thus losing any other state one has currently configured in devtools.

Perhaps a solution would be to stop collecting data while react-devtools tab is not focussed, or provide a means to disconnect this polling mechanism without restarting Chrome devtools entirely.

jaredly:
Yeah. There is some redundancy in the data we send over - so we can probably streamline things a bit.
Not sure about what we want to do about "stop collecting data"

jaredly:
actually, I can think of a way to pause the sending of data. I'll investigate that.

hummlas:
When I open up the React tab in chromes dev-tools chrome becomes incredibly slow, almost unresponsive. Even trying to close the dev-tools might take a few seconds, but as soon as the dev-tools closes the speed/responsiveness goes back to normal. 

If I look in Chrome task-manager I can see that the tab where React dev-tools are open increase in memory usage by the second, going up from ~200MB to before switching to the React-tab in the dev-tools and within a minute of switching increasing to 900+ and the it starts going back and forth between 500-ish to 900-ish.

Using Chome version 45.0.2454.85 m and React dev-tools version 0.14, still using React 0.12 though. But a friend with same version of chrome, React dev-tools and the same site does not have the same perf-issues..

vasco3:
Having the same issue with Version 46.0.2490.80 (64-bit)  React devtools 0.14.4

ansonkao:
+1

...also... https://www.change.org/p/github-inc-add-voting-functionality-for-github-issues

ayozebarrera:
Having the same issue here Version 51.0.2704.84 (64-bit) React devtools 0.14.11

gaearon:
#146 made it significantly better. I also made some changes later to address this.
[...]

PR #146, opened 2015-08-14
pause the data transfer if the tab is not selected

jaredly:
Changes
- when the devtools panel is not visible, don't send data from the backend to the panel (this slows down pages)
- once the devtools panel is shown, send over all the buffered events
- dedup events sent from the store so that all the buffered events don't kill the panel UI

addresses #129

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -85,6 +85,8 @@ class Store extends EventEmitter {
   _nodes: Map;
   _parents: Map;
   _nodesByName: Map;
+  _eventQueue: Array<string>;
+  _eventTimer: ?number;
 
   // Public state
   contextMenu: ?ContextMenu;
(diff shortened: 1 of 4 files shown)


frontend/Store.js
```diff
@@ -141,6 +143,24 @@ class Store extends EventEmitter {
...
+        this._eventQueue.forEach(evt => EventEmitter.prototype.emit.call(this, evt));
+        this._eventQueue = [];
+        this._eventTimer = null;
+      }, 50);
+      this._eventQueue = [];
+    }
+    if (this._eventQueue.indexOf(evt) === -1) {
+      this._eventQueue.push(evt);
+    }
+    // to appease flow
```
spicyj: Eh? What _should_ this function return?

spicyj:
Looks good I think, thanks.

frontend/Store.js
jaredly: imo it "should" return void. But flow wants a bool to align with the signature of EventEmitter. The bool indicates whether the emit "succeeded"

Issue #52, opened 2014-11-12
Tools should display an additional field for the 'native js' version of Immutable objects


twisterghost:
While working on a React/Flux project and having Immutables here and there, I've noticed that the dev tools become pretty silly when inspecting the state of an element that uses Immutable.

While first looking through the codebase, I noticed that the state of an object is being pulled and pushed to the sidebar [here](https://github.com/facebook/react-devtools/blob/master/views/components/ReactPanel.js#L1001). With a bit of hacking around, I began to notice that there is no place where the actual values are read or passed around, but instead, only references to the values.

I'm not sure how possible this may be, but I feel that with the deep ties between React, Flux and Immutable, the dev tools should be able to detect an immutable and provide either an extra property on the state panel or a new panel altogether with the `.toJS()`'d version of the immutable to help developers work with these utilities.

BinaryMuse:
Agreed, this would be really nice. Not sure if others agree, but I feel like in an ideal world, we could integrate application code into the React devtools (e.g. to make them understand arbitrary data structures)

jiyinyiyong:
After reading this new post I really want this to be fixed
http://facebook.github.io/react/docs/advanced-performance.html
Feeling bad seeing this:
![data](https://cloud.githubusercontent.com/assets/449224/6188001/1008996e-b3c8-11e4-9a70-e45e050af12f.png)

0x80:
+1

cassus:
+1

cesarandreu:
+1

danbruder:
+1

spicyj:
"+1" comments do not help us. Please stop.

cbenz:
Just asked myself the same question and found that with the [new devtools extension](https://facebook.github.io/react/blog/2015/08/03/new-react-devtools-beta.html), the `$r` global variable helps a lot: even if the sidebar still shows the internals of the immutable variables, I can select the component and type in the console:

```
$r.props.myImmutableProp.toJS()
```

or even use right-click "Store as a global variable" on the immutable prop/state then:

```
$tmp.toJS()
```

ntucker:
+500

dubert:
:+1: 

This is so tedious to open every time:

![screen shot 2015-08-28 at 12 49 15](https://cloud.githubusercontent.com/assets/1843672/9552057/377713ea-4d83-11e5-8604-5fceecd6a7fc.png)

neoziro:
[...]

PR #149, opened 2015-08-16
Support for inspecting Immutable.js objects

garetht:
This pull request makes changes meant to support the functionality requested for in https://github.com/facebook/react-devtools/issues/52.

**Changes Made**
- When an object is an Immutable object, it is now expandable so that its properties can be examined
- The display name for the object notes that it is an Immutable object
- Additional tests for the dehydrator and Immutable utilities have been added

### DIFF
--- a/agent/immutableUtils.js
+++ b/agent/immutableUtils.js
@@ -14,8 +14,7 @@ var Immutable = require('immutable');
 
 var types = ['OrderedMap', 'OrderedSet',
              'Map', 'Set', 'List', 'Stack',
-             'Seq',
-             'Iterable'];
+             'Seq'];
 
 module.exports = {
   isImmutable: Immutable.Iterable.isIterable,
--- a/agent/immutableUtils.js
+++ b/agent/immutableUtils.js
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+'use strict';
+
+var Immutable = require('immutable');
+
+var types = ['OrderedMap', 'OrderedSet',
+             'Map', 'Set', 'List', 'Stack',
+             'Seq',
+             'Iterable'];
+
+module.exports = {
+  isImmutable: Immutable.Iterable.isIterable,
+  getImmutableName(data) {
+    if (Immutable.Iterable.isIterable(data)) {
+      var name = 'Immutable.',
+          typelen = types.length;
+      for (var i = 0; i < typelen; i++) {
+        if (Immutable[types[i]]['is' + types[i]](data)) {
+          name += types[i];
+          break;
+        }
+      }
+      return name;
+    } else {
+      return data.constructor.name;
+    }
+  },
+  shallowToJS(dataStructure) {
+    if (!Immutable.Iterable.isIterable(dataStructure)) {
+      return dataStructure;
+    }
+
+    if (Immutable.Iterable.isKeyed(dataStructure)) {
+      return dataStructure.toObject();
+    } else if (Immutable.Iterable.isIndexed(dataStructure) ||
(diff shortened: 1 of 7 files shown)


jaredly:
Hmmm. So that looks reasonable.
My inclination would be to instead implement this as a plugin, and either:
1 - make it so that plugins can alter the representation of props, or
2 - show immutable objects in an extra section in the sidebar, separate from props

Thanks for putting this together!
@spicyj do you have an opinion on how this is implemented?

garetht:
That makes sense to me. Do you have a sample plugin I can consult?

agent/immutableUtils.js
```diff
@@ -0,0 +1,50 @@
...
+          name += types[i];
+          break;
+        }
+      }
+      return name;
+    } else {
+      return data.constructor.name;
+    }
+  },
+  shallowToJS(dataStructure: any): any {
```
jackwanders: May I ask your rationale behind providing a `shallowToJS` here but not a `deepToJS` that calls [`Iterable.toJS()`](https://facebook.github.io/immutable-js/docs/#/Iterable/toJS)? For example, if a particular prop is an `Immutable.KeyedIterable` with a property value that is itself an Immutable object, then it seems that `shallowToJS` would not convert the property value to a mutable object.

garetht: Was just about to write that. The reason in edit 2 is precisely why.

jackwanders:
@garetht, I believe @jaredly is referring to the idea of `extraPanes`, as implemented in [`frontend/Panel.js`](https://github.com/facebook/react-devtools/blob/3948128bf12696e9c2dc0fd20c84dc8ffa5b30c6/frontend/Panel.js#L231). Currently, it looks like the `rnStyle` capability is used to conditionally add an extra pane for `ReactNative` styles.

I'm not sure what the most appropriate way to implement these utils as a conditional functionality is, but one option might be to render an `ImmutableInspector` pane underneath the existing component inspection pane when the user inspects an immutable object. Some kind of top level toggle would be nicer, in my opinion, but may require more work as there doesn't appear to be an existing UI element in which to render something like a checkbox that, when checked, calls `.shallowToJS` on all immutable objects.

jackwanders:
@garetht, @jaredly, I took a pass at implementing the "inspect immutable objects as js" functionality as an optional setting.

https://github.com/garetht/react-devtools/pull/1

This pull request creates a new checkbox at the top of `PropState` which, when checked, will toggle a property in `Store` and force an update of `Container`. The change wound up touching more modules than I had hoped, and someone with a better understanding of the devtools architecture may be able to clean things up.

jaredly:
Yeah -- I'm hesitant to bake in Immutable support; it would be awesome to see this as a plugin; I'll take some time this weekend to figure out what that would look like

Daniel15:
:+1:  I'm looking forward to this!

zackargyle:
I decided against using Immutable for this very reason. It really renders the dev tools useless in inspecting data. As long as immutability is being encouraged through projects like redux, I think the few extra lines of code should definitely be baked into the devtools.

kristian-puccio:
You can always select an object in the dev tools and query it using $r I
think. Then just use .toJS() in the console to print out the value.

But yeah it's definitely not as quick as it could be

Daniel15:
@zackargyle - You could always clone @garetht's version of the React devtools (ie. the branch this pull request points to) and try it out :)

gaearon:
As for the feature, I think it is really necessary we get something like this merged.
However I’m not sure about this implementation.

I think we should strive to support any iterables, and not just Immutable.js specifically. As an example, [react-json-tree](https://github.com/alexkuz/react-json-tree) checks whether [something has an iterator](https://github.com/alexkuz/react-json-tree/blob/7b1ccfa1e641ca2005afb2e6968f5ead114b5db0/src/objType.js#L3) and [handles iterable nodes in a specific way](https://github.com/alexkuz/react-json-tree/blob/7b1ccfa1e641ca2005afb2e6968f5ead114b5db0/src/JSONIterableNode.js). I think we should do the same here.

Would you like to look into reworking this PR to support arbitrary iterables? Thanks.

gaearon:
Closing as there is no response from the author.

gaearon:
A more generic support for this landed in https://github.com/facebook/react-devtools/pull/459.

PR #154, opened 2015-08-18
Allow selecting component name in right sidebar

### LINKED ISSUE #153: Can't easily copy component name
You used to be able to double-click the component name and copy it in the tree view, now you can't. The text in the right pane also isn't selectable.

spicyj:
Fixes #153.


jaredly:
looks awesome

PR #167, opened 2015-08-28
Make ART work

### LINKED ISSUE #160: Integration with non DOM render targets.
I'm doing something similar to react-canvas and react-pixi where the render target is not actually DOM nodes although it somewhat hijacks the react-dom renderer. In my case it's a scene graph that renders to a canvas element using WebGL. When using these devtools with my project I don't see anything below my "Surface" component (the component that contains the canvas element) other than the canvas element. Is there a way to fake it so that my components will show up?

spicyj:
Fixes #160.

image

### DIFF
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -81,6 +81,13 @@ function getData(element: Object): DataType {
       setInContext: inst.forceUpdate && setInContext.bind(null, inst),
     };
     publicInstance = inst;
+
+    // TODO: React ART currently falls in this bucket, but this doesn't
+    // actually make sense and we should clean this up after stabilizing our
+    // API for backends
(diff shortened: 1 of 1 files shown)


jaredly:
Oh interesting. Thanks for fixing this.

fsbdev:
Thanks, any idea what release this will be in?

spicyj:
Planning to push it to the Chrome store this week.

fsbdev:
I just pulled master and it works for my app too but is it also supposed to/did it ever support highlighting ReactART elements when you hover over the node in the treeview, like it does with ReactDOM elements?

spicyj:
No, sorry.

PR #169, opened 2015-09-01
Add files for Chrome and Firefox releases

spicyj:


shells/chrome/manifest.json
```diff
@@ -18,8 +20,7 @@
 
   "permissions": [
     "http://*/",
-    "https://*/*",
-    "file://*"
```
kassens: Why did you remove `file://` permission? Seems like this would allow debugging on local files that don't go through http?

spicyj: Chrome Web Store doesn't allow it – the extension was automatically rejected. :(

shells/firefox/webpack.config.js
```diff
@@ -9,8 +9,10 @@
  */
 'use strict';
 
+var __DEV__ = process.env.NODE_ENV !== 'production';
+
 module.exports = {
-  // devtool: 'cheap-module-eval-source-map',
+  devtool: __DEV__ ? 'cheap-module-eval-source-map' : false,
```
jaredly: Firefox actually has rather poor sourcemap support generally, which is why I disabled source maps even in dev.

spicyj: Oh, I missed this. I'll comment it back out.

PR #170, opened 2015-09-01
Fix some $FlowFixMe by subclassing React.Component

kassens:
Is there a reason we should avoid extending `React.Component`?


spicyj:
No, in fact we'll probably require it (https://github.com/facebook/react/pull/4663).

PR #175, opened 2015-09-03
Use BSD-3-Clause in Firefox package

zpao:
We're using the 3 clause, not the 2 clause. This shows up on AMO where it links off-site to OSI. We have the right license in the root, so it's just a matter of correctness.


spicyj:
The license shown on AMO might just be based on what's configured in the admin interface. Not sure.

PR #176, opened 2015-09-03
Naturally sort data keys

jayphelps:
Before
image
After
image

The most notable difference is sorting of numbers, like in array keys. (i.e. 10 after 9, instead of after 1) but other things are sorted better as well. See the [javascript-natural-sort docs for more examples](https://www.npmjs.com/package/javascript-natural-sort).

### DIFF
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -37,7 +37,7 @@ class DataView extends React.Component {
     }
     var names = Object.keys(data);
     if (!this.props.noSort) {
-      names.sort();
+      names.sort(alphanumericSort);
     }
     var path = this.props.path;
     if (!names.length) {
@@ -188,6 +188,17 @@ class DataItem extends React.Component {
       </li>
     );
   }
+}
+
+function alphanumericSort(a, b) {
+  if ('' + (+a) === a) {
+    if ('' + (+b) !== b) {
+      return -1;
+    }
+    a = +a;
+    b = +b;
+  }
+  return (a < b) ? -1 : 1;
 }
 
 var styles = {


jaredly:
I definitely want array indices to sort write, but I looked at [the source](https://github.com/Bill4Time/javascript-natural-sort/blob/master/naturalSort.js) and it seems like overkill for this instance, given that what we are sorting are attribute names.

My inclination is to just mirror chrome's behavior here, which I would assume is just alphabetical sort with a special case for numbers.

jayphelps:
@jaredly Okiedoke. I disagree, so I'll let someone else take it from here. Just because a similar app doesn't sort keys as naturally as this would, doesn't mean we should strive to lower things to that level. If microperf of filesize was a concern, I'd be a little more inclined to be okay with less, but since it's not really an issue in extensions there's just no excuse except that Chrome doesn't do it.

jaredly:
:)
don't let me shut you down --- I'm open to more discussion
and actually I looked deeper into it and it looks like a good thing

spicyj:
It would be highly surprising to sort dates in d/m/y format "naturally" (or is it m/d/y? can't tell from their examples).

jaredly:
I don't think it actually knows anything about dates. It just splits sequences of numbers and sorts by chunk.

spicyj:
https://github.com/Bill4Time/javascript-natural-sort/blob/a56cd90d4a2491a8e1423b2cc5d2d50aef7bb364/naturalSort.js#L22 :(

jaredly:
oh right.

jayphelps:
@spicyj Just to clarify, you're referring to this:
```js
let dates = ['10/12/2008','10/11/2008','10/11/2007','10/12/2007'];

dates.sort(naturalSort)
// ['10/11/2007', '10/12/2007', '10/11/2008', '10/12/2008']

// vs.

date.sort();
// ['10/11/2007', '10/11/2008', '10/12/2007', '10/12/2008']
```

Right? I can see arguments both ways. If you really are storing dates as keys, having them by true logical date could certainly be helpful in debugging a large range of dates spanning years.

jaredly:
I feel like that gets rather beyond the scope of what we have here. You're
not going to want to visually inspect large ranges of anything -- just
bring it into the console and inspect it w/ JS.

kassens:
What about a simple sort function that sorts purely numeric values first and everything else alphabetically?

```javascript
    function sort(a, b) {
      if ('' + (+a) === a) {
        if ('' + (+b) !== b) {
          return -1;
        }
        a = +a;
        b = +b;
      }
      return (a < b) ? -1 : 1;
    }

**Comment by jayphelps** [comment]
@kassens I believe that's what @jaredly was recommending initially. I went ahead and committed it on your behalf, thanks!

**Comment by spicyj** [comment]
Not quite that simple because `a1, a2, a10` should be sorted in that order.

**Comment by spicyj** [comment]
How about this?

```
function sort(a, b) {
  var zeroCode = '0'.charCodeAt(0);
  var nineCode = '9'.charCodeAt(0);
  var ai = 0;
  var bi = 0;
  while (ai < a.length && bi < b.length) {
    var ac = a.charCodeAt(ai);
    var bc = b.charCodeAt(bi);
    if (zeroCode <= ac && ac <= nineCode && zeroCode <= bc && bc <= nineCode) {
      var an = 0;
      var bn = 0;
      do {
        an = 10 * an + ac - zeroCode;
      } while (
        (ac = a.charCodeAt(++ai)) &&
        zeroCode <= ac && ac <= nineCode
      );
      do {
        bn = 10 * bn + bc - zeroCode;
      } while (
        (bc = b.charCodeAt(++bi)) &&
        zeroCode <= bc && bc <= nineCode
      );
      if (an !== bn) {
        return an - bn;
      }
    } else if (ac !== bc) {
      return ac - bc;
    } else {
      ai++;
      bi++;
    }
  }
  return a.length - b.length;
}
```

**Comment by jayphelps** [comment]
@spicyj I'll give it a test here shortly. Here's basically what Chrome's dev tools is doing: (I basically just reformatted their code cause it used very different coding styles)

```js
function alphaNumberericSort(a, b) {
  if (a === '__proto__') {
      return 1;
  } else if (b === '__proto__') {
      return -1;
  }

  let splitRegexp = /^\d+|^\D+/;

  while (1) {
    if (a) {
      if (!b) {
        return 1;
      }
    } else {
      if (b) {
        return -1;
      } else {
        return 0;
      }
    }

    let aPart = a.match(splitRegexp)[0];
    let bPart = b.match(splitRegexp)[0];
    let aNumber = !isNaN(aPart);
    let bNumber = !isNaN(bPart);

    if (aNumber && !bNumber) {
      return -1;
    } else if (bNumber && !aNumber) {
      return 1;
    } else if (aNumber && bNumber) {
      let diff = aPart - bPart;
      if (diff) {
        return diff;
      } else if (aPart.length !== bPart.length) {
        if (!+aPart && !+bPart) {
          return aPart.length - bPart.length;
        } else {
          return bPart.length - aPart.length;
        }
      }
    } else if (aPart !== bPart) {
      return (aPart < bPart) ? -1 : 1;
    }

    a = a.substring(aPart.length);
    b = b.substring(bPart.length);
  }
}
```

One thing to note is that they also include `Symbol`s, sorted and placed at the bottom of the normal key list. We could does this simply with:

```js
var names = Object.keys(data);
names.sort(alphanumericSort);
var symbols = Object.getOwnPropertySymbols(data);
symbols.sort(alphanumericSort);
names.push.apply(names, symbols);
```

jaredly:
I think we can accept this as-is, and maybe continue discussion about fancier sorting in a separate PR/Issue.
@kassens do you want to merge this? Given that it's your code :)

kassens:
Cool, was concerned about the Google code regarding license. Will merge the simple one from this PR now.

spicyj:
You can use the one I posted. I wrote it myself.

jayphelps:
@kassens rebased, in case you just wanna merge this PR.

kassens:
@jayphelps Sweet, thanks!

Issue #172, opened 2015-09-02
Chrome devtools don't work on file: URLs


spicyj:
This is presumably because I removed it from the list of permissions… but the old devtools didn't have any permissions listed in manifest.json and did work on file URLs, so maybe we can fix this regression.

arm22:
Sorry, totally new here, but I think that your presumption is correct. Here is what my error looks like if it helps:
`Unchecked runtime.lastError while running tabs.executeScript: Cannot access contents of url "file:///Path/To/React/app". Extension manifest must request permission to access this host.
at installContentScript (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/background.js:91:16)
at chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/background.js:67:6`

spicyj:
@arm22 Thanks. I guess chrome.tabs.executeScript doesn't work on file: URLs?

kassens:
According to [this answer](https://stackoverflow.com/questions/19493020/adding-file-permission-to-chrome-extension/19493206#19493206) on StackOverflow users have to tick the checkbox in the settings in order to allow access to `file://`. This seems to work as expected for the `content-script` that we're injecting on every page, but doesn't seem to work for `chrome.tabs.executeScript`. Sounds like a Chrome bug to me. (filed https://crbug.com/529548)

A quick test seems to confirm that doing the work from `contentScript.js` directly in `GlobalHook.js` would fix this issue. This would mean we do slightly more work on every page though.

CC @paulirish

dbashford:
FWIW, allowing access to `file://` gets React dev tools started, but once started, it, for me, gets stuck on "Connecting to react..." as per #179.

syndbg:
@dbashford Confirming that I have the same issue. No errors, but stuck on `Connecting to React`

ErikSchierboom:
@dbashford Same here.

kumthekar:
Same problem on Chrome Version 46.0.2490.71 (64-bit) on Mac OS 10.9.5

Stuck on 'Connecting to react ...'

wzpan:
Same problem.

fuchao2012:
@wzpan open the [react](http://facebook.github.io/react/) to test the dev tools. And do not test the page in file URL

Bnaya:
Same problem, even with "allow access to file urls"

darknoon:
[...]

PR #189, opened 2015-09-05
add support for local file debugging #172

arm22:
The file permissions were removed in the update, this adds local file
debugging


ngyikp:
https://github.com/facebook/react-devtools/pull/169/files#r38485951 :/

arm22:
Wow it really sucks that the web store won't allow that permission! Thanks for the update @ngyikp

PR #192, opened 2015-09-08
Set minimum_chrome_version for Chrome extension

### LINKED ISSUE #159: Add 'minimum_chrome_version' for Chrome extension
To avoid similar issues like #151 because of an old Chrome version, the manifest should set `minimum_chrome_version` to the current stable version, which is `44`.

https://developer.chrome.com/extensions/manifest

kassens:
43 is 2 versions behind the current stable to cover people lagging behind in the
updater.

Closes #159


spicyj:
Sure.

PR #193, opened 2015-09-08
Fix lint warning

kassens:
Fixes "warning: evt is already declared in the upper scope" since `evt` was
shadowed.


spicyj:
Can we rename the outer to `event` instead? Generally I like variables for larger scopes to have longer names. Also would be nice if we got lint set up on Travis.

kassens:
Updated.

Travis runs lint (see the console output), but I guess lint warning currently don't cause travis to fail the build.

PR #194, opened 2015-09-08
Update eslint

kassens:
Ran this to confirm we're not seeing new lint warnings:

    npm run lint

### DIFF
--- a/frontend/flash.js
+++ b/frontend/flash.js
@@ -22,7 +22,7 @@ function flash(node: DOMElement, flashColor: string, baseColor: string, duration
   node.style.transition = 'none';
   node.style.backgroundColor = flashColor;
   // force recalc
-  node.offsetTop;
+  node.offsetTop; // eslint-disable-line no-unused-expressions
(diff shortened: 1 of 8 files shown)


spicyj:
eslint 1 disables all the formerly-default rules. Can we copy from fbjs? Like #4746.

kassens:
Added `fbjs-scripts` as a dependency and use its eslintrc as the base similar to React. I disabled some lint rules that need a bigger cleanup which I might do in a followup.

frontend/flash.js
```diff
@@ -22,7 +22,7 @@ function flash(node: DOMElement, flashColor: string, baseColor: string, duration
   node.style.transition = 'none';
   node.style.backgroundColor = flashColor;
   // force recalc
-  node.offsetTop;
+  node.offsetTop; // eslint-disable-line no-unused-expressions
```
spicyj: Can also write `void node.offsetTop;`.

spicyj:
:+1:

PR #196, opened 2015-09-09
Move flow declarations to flow libs

kassens:
Instead of having the flow type declarations duplicated in many places and
distracting from the file contents, define them once in a flow lib file.

A small downside is that flow now assumes the `chrome` global is available
everywhere. This is mitigated by only whitelisting the global for lint in
selected filed.


kassens:
@spicyj

Issue #187, opened 2015-09-05
Add support for Relay


kassens:
It would be great if we could show details about what's going on in Relay and make the inspection of Relay components better.

- Show network activity similar to the Network panel, but specialized for Relay.
- Inspection of the local data that Relay knows about.
- Show composed queries of RelayContainers potentially with a configurable link to GraphiQL.

PR #186, opened 2015-09-04
Add a tabbed pane [updated]

kassens:
This is an update to @jaredly's PR #145. Not sure how to do that with github correctly…

Small changes from me on top:
- fix flow issue
- lint fix
- hide the sub tab if there's only a single tab

PR #200, opened 2015-09-10
Add RelayPlugin

kassens:
This is mostly @jaredly's work to add an initial plugin for Relay.

It currently includes #186, but I'll rebase once that's landed to clean this PR up.

I've fixed lint and flow issues as well as actually disabled this plugin by default. Since we don't have options yet, it's just commented out in `Panel.js`.

It's the beginning of #187.

### DIFF
--- a/plugins/Relay/Store.js
+++ b/plugins/Relay/Store.js
@@ -16,12 +16,13 @@ var {EventEmitter} = require('events');
 var {OrderedMap, Map} = require('immutable');
 var assign = require('object-assign');
 var consts = require('../../agent/consts');
+var invariant = require('../../frontend/invariant');
 
 function getDataIDs(obj, collector) {
   for (var name in obj) {
-    if (name === 'id' && 'string' === typeof obj[name]) {
+    if (name === 'id' && typeof obj[name] === 'string') {
       collector.push(obj[name]);
-    } else if (typeof obj[name] == 'object') {
+    } else if (typeof obj[name] === 'object') {
       getDataIDs(obj[name], collector);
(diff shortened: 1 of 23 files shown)


kassens:
@yungsters @steveluscher @spicyj Thoughts on moving forward with this?

plugins/Relay/Store.js
```diff
@@ -0,0 +1,187 @@
...
+      dataIDs.forEach(id => {
+        if (!this.queriesByDataID[id]) {
+          this.queriesByDataID[id] = [data.id];
+        } else {
+          this.queriesByDataID[id].push(data.id);
+        }
+      });
+    });
+    bridge.on('relay:success', ({id, response, end}) => {
+      console.log('response', id, response);
```
yungsters: We should remove these.

kassens:
@yungsters and I thought it's best to land this (behind the URL toggle) and then iterate.

jaredly:
At one point I had a settings tab [see branch here](https://github.com/jaredly/react-devtools/tree/settings-tab), where I planned to allow users to enable/disable various plugins. That way you wouldn't need the location.hash thing.

kassens:
@jaredly: Yeah, I saw that but it didn't seem worth it for a single option. Once, we evolve the plugin a bit more, I think we can enable most things just when we see it's being used on the page.

jaredly:
Other things that a settings page would be useful for:
Switching off display of `<RelayWrapper/>` components, see #211 

I agree that relay should only appear when it's detected though

PR #205, opened 2015-09-12
Enable more lint rules! (Fixes #195)

### LINKED ISSUE #195: Enable more lint rules
The lint rules on the bottom of [`.eslintrc`](https://github.com/facebook/react-devtools/blob/master/.eslintrc) should probably all be enabled. Not sure we want the `const` one.

flipstewart:
shocked


agent/Bridge.js
```diff
@@ -323,16 +323,13 @@ class Bridge {
     if (!this._callers[name]) {
       return console.warn('unknown call');
     }
-    var args = args;
-    if (!Array.isArray(args)) {
-      args = [args];
-    }
+    var preparedArgs = !Array.isArray(args) ? [args] : args;
```
kassens: You could just overwrite `args` here when you remove the `var ` prefix.

.eslintrc
```diff
@@ -46,9 +46,9 @@ rules:
   react/wrap-multilines: [2, {declaration: false, assignment: false}]
 
   # TODO: we might want these, but currently have multiple warnings
```
kassens: You can remove this now :smile:

frontend/decorate.js
```diff
@@ -130,14 +130,14 @@ module.exports = function (options: Options, Component: any): any {
 function arrayDiff(array, oldArray) {
   var names = new Set();
   var missing = [];
-  for (var i = 0; i < array.length; i++) {
-    names.add(array[i]);
+  for (var ai = 0; ai < array.length; ai++) {
```
kassens: If flow supported `let` already, we could use that here...

Otherwise, I'd prefer `i`, `j`, `k` as more traditional.

flipstewart: :+1:

.eslintrc
flipstewart: :joy:

agent/Bridge.js
flipstewart: :+1:

kassens:
This is great. Thanks!

flipstewart:
@kassens happy to help!

PR #207, opened 2015-09-14
Add support for Firefox private browsing

ngyikp:
Without this, the React tab would still show up, but it stays stuck on "Connecting to devtools"

screen shot 2015-09-14 at 2 45 08 pm

Documentation for the "permissions" key can be found at: https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json#permissions


kassens:
Great, thanks!

PR #213, opened 2015-09-17
Make sure a Bridge always has a Wall

kassens:
The flow type annotation was already promising that there would always be a wall
set. This ensures it by making the wall part of the constructor.


spicyj:
Seems like this should've been a flow error.

kassens:
It looks like flow doesn't validate that all fields are initialized in constructors (probably because existing code often initializes fields using methods and it's difficult/impossible to track).

PR #218, opened 2015-09-21
Update Firefox shell for path change in Firefox 44 and later

jryans:
In Firefox 44, we are [moving DevTools files][1] to a new location. This impacts some add-ons that import those files at their old paths.

For React, there's only one, as the rest go through the Add-on SDK APIs which are already updated correctly.

I've updated the add-on to use the new path when available (Firefox 44 and later) and fallback to the old one if not.

I tested this add-on build in Firefox 43 and 44.

At the moment, the old path also works in 44, but we'll soon start logging deprecation warnings for the old paths, so now you'll already be updated to avoid such warnings.

[1]: https://bugzil.la/912121

### DIFF
--- a/shells/firefox/main/trackSelection.js
+++ b/shells/firefox/main/trackSelection.js
@@ -10,7 +10,16 @@
 'use strict';
 
 const { Cu } = require('chrome');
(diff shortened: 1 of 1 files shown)


spicyj:
Thank you!

PR #222, opened 2015-09-22
Upgrade to flow 0.16.0

kassens:
Upgraded the references and fixed a type error caused by the upgrade.

@spicyj, @jaredly: Was there a reason to re-define the `Bridge` type? This broke for some reason with the upgrade.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -13,9 +13,9 @@ suppress_type=$FlowIssue
 suppress_type=$FlowFixMe
 suppress_type=$FixMe
 
-suppress_comment=\\(.\\|\n\\)*\\$FlowFixMe\\($\\|[^(]\\|(\\(>=0\\.\\(1[0-3]\\|[0-9]\\).[0-9]\\)? *\\(site=[a-z,_]*react_native[a-z,_]*\\)?)\\)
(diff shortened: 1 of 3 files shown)


spicyj:
No idea. What was the error?

kassens:
If I try to keep using the type definition for `Bridge`, I get some obscure flow error. Importing the type works find and I don't see that it would violate some isolation, so this seemed to be a good fix?

spicyj:
What was the error?

kassens:
If I revert the change on `Agent.js`, I get the following error.

```
$ flow

/Users/jkassens/Code/react-devtools/shells/chrome/src/backend.js:64:3,25: call of method addBridge
Error:
/Users/jkassens/Code/react-devtools/agent/Agent.js:20:15,24:1: object type
This type is incompatible with
global object

/Users/jkassens/Code/react-devtools/shells/firefox/src/backend.js:65:3,25: call of method addBridge
Error:
/Users/jkassens/Code/react-devtools/agent/Agent.js:20:15,24:1: object type
This type is incompatible with
global object

/Users/jkassens/Code/react-devtools/shells/plain/backend.js:31:1,23: call of method addBridge
Error:
/Users/jkassens/Code/react-devtools/agent/Agent.js:20:15,24:1: object type
This type is incompatible with
global object

Found 3 errors

spicyj:
Weird. Sure, this is fine. Thanks for pasting in the error.

PR #228, opened 2015-09-25
Show `{…}` instead of `{}` for displaying objects

MichelleTodd:
Displaying some objects as `{}` in the devtools and others as `{…}` makes it
look like the objects displayed as `{}` are empty, when in reality they are
just as likely to contain things.

This updates objects displayed in the PropPane and the SearchPane to be
displayed as `{…}`.


spicyj:
Not sure when `data[consts.type]` in previewComplex is falsey, but this seems more consistent with that too. Thanks!

PR #231, opened 2015-09-29
Improvements and fixes Relay queries tab.

kassens:
This fixes a problem caused by a wrong data format that prevented queries from correctly showing up in the Relay plugin and adds a formatter to display the queries in a readable format.


wincent:
Looks good to me.

PR #233, opened 2015-09-30
Factor out side panel components

kassens:
I want to re-use the styles of the side panel in the Relay plugin and they might also be useful in other places.


wincent:
Looks good.

PR #234, opened 2015-09-30
Replace deprecated eslint rule

kassens:
Also update the version of the linter and fbjs-scripts.


spicyj:
I have no context here but lint still passes so… sure?

kassens:
It was just a warning that `react/jsx-quotes` was deprecated. Figured I use the chance to also update lint.

PR #235, opened 2015-09-30
Fix bug in Relay store

kassens:
When the node was `null`, the Relay store would error on the change line. This
prevented the store from initializing correctly.

Test Plan:
Load the Relay todo example and open the React elements panel, see it's loaded
correctly.


wincent:
Looks good.

PR #237, opened 2015-10-01
Move all guid generation into a util

kassens:


kassens:
This needs an update as the injected code currently cannot `require` modules.

jaredly:
GlobalHook is the one place where I think we should leave it; it's really nice for the GlobalHook fn to be `toString()`able so you can inject it anywhere without fuss.

kassens:
Reverted the global hook file.

PR #238, opened 2015-10-01
Cleanup Relay query display

kassens:
This uses the generalized components from the React elements panel to achieve a
more consistent UI.

screenshot 2015-10-01 11 16 22


kassens:
Merging this, as I don't think there's anything controversial here. If there's concerns, feel free to request a fixup.

PR #241, opened 2015-10-02
Instrument Relay on page load time instead of when the devtools are opened

kassens:
This diff includes 2 easy changes:

- Rename `GlobalHook.js` to `installGlobalHook.js` since it exports a function
- Reduce indentation in `installGlobalHook` by adding an early return.

The last revision is the interesting part. It instruments Relay as soon as Relay registers itself with React devtools. This means that we instrument even if the developer panel is not open. This is necessary when we want to inspect GraphQL queries that are made for the initial page load. It's a bit ugly that we there's a bunch of Relay only code in `installGlobalHook` now, but we can't easily move it out since we `toString` that function to inject it into the page.

References #187

### DIFF
--- a/plugins/Relay/backend.js
+++ b/plugins/Relay/backend.js
@@ -13,18 +13,6 @@
 import type Bridge from '../../agent/Bridge';
 import type Agent from '../../agent/Agent';
 
-import guid from '../../utils/guid';
-
-function decorate(obj, attr, fn) {
-  var old = obj[attr];
-  obj[attr] = function() {
-    var res = old.apply(this, arguments);
-    fn.apply(this, arguments);
-    return res;
-  };
-  return () => (obj[attr] = old);
-}
-
 module.exports = (bridge: Bridge, agent: Agent, hook: Object) => {
   var shouldEnable = !!(
     hook._relayInternals &&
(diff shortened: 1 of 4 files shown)


kassens:
@jaredly, @spicyj thoughts on this approach or alternatives?

A way to make the injected script support requires would be to compile it with webpack in a first step and in a second step include the raw file content with the webpack `raw-loader` (`require('raw!build/globalHook.js')`)

jaredly:
Another thought:
Chrome devtools doesn't instrument network requests unless it's open (when you open devtools, the network tab is blank).
Is there a way for relay devtools to follow that lead? e.g. only instrument if you're reloading w/ devtools open?

kassens:
- The problem is that the communication if the devtools are open happens asynchronously. I couldn't find a way to find out early enough if the devtools are open.

- Moved the Relay hook to a separate file now. The downside is that this now has to be added to every place where we install the global hook in order to enable the specific plugin. I'll investigate if we can load the header script from a file instead to allow us to use `requires` there. Created #245 to investigate improvements here.

jaredly:
Fantastic! I like this much better.

gaearon:
I'm confused about this. When did `setRequestListener` ever exist? I can't seem to find it in the [history for Relay internals](https://github.com/facebook/relay/commits/db2e1815416b845ffd404582722520411748cfbd/src/RelayPublic.js).

gaearon:
Answering to myself: [the owls are not what they seem](https://github.com/facebook/react-devtools/pull/241/files#diff-65771b3fb309c23cc705df9f7476f434R111).

PR #242, opened 2015-10-02
Do not run global hook in Chrome panel context

kassens:
The file `shells/chrome/src/GlobalHook.js` already runs `backend/GlobalHook.js`
in the context of the page. I don't think there's a reason to also execute it in
the context of the Chrome extensions developer panel.

### DIFF
--- a/shells/chrome/src/panel.js
+++ b/shells/chrome/src/panel.js
@@ -94,8 +94,6 @@ var config: Props = {
   },
 };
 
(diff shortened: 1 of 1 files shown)


kassens:
@jaredly Let me know if I'm missing anything here? I tested the devtools and couldn't find anything that doesn't work.

jaredly:
Yeah that looks fine -- it was left over from when I had a "debug devtools using devtools" setup that isn't there anymore.

PR #246, opened 2015-10-06
Remove global-hook.js file from plain shell

kassens:
This file was no longer used. The `container` entrypoint sets up the global hook
instead.


spicyj:
:+1:

PR #249, opened 2015-10-07
Upgrade to flow 0.17.0

kassens:
Also simplify the `.flowconfig` as we don't make use of these extra options.


kassens:
Since nothing fails, I'm merging this so future PRs use flow 0.17 to check.

PR #252, opened 2015-10-08
Fix #183 - [firefox] Add invertIconForLightTheme: true

### LINKED ISSUE #183: [firefox] Icon is invisible on light theme
Right now, the react tool icon is white, and in light theme, it looks invisible since the background is too light. A simple fix would be adding `invertIconForLightTheme: true` in the tool definition.

nt1m:


spicyj:
Do you know when your patch will hit stable? I guess adding this doesn't hurt anything until it does though since it already looks wrong…

nt1m:
@spicyj The patch landed in Firefox 43, which is currently in Developer edition, and which will land in stable on 2015-12-15. It's still nice to have this PR merged for Firefox Developer Edition users using React DevTools.

spicyj:
Got it. I was thinking that everyone on developer edition used the dark theme but maybe not…

nt1m:
@spicyj Most people do, but they can still switch on the light theme if they like.

PR #253, opened 2015-10-08
Chunk up Relay query test to prevent cutoff

kassens:
The dehydration for sending data over the bridge cuts off strings at 500
characters. This is good to prevent overly large strings inside of elements. For
Relay queries which can often get longer, we don't want to have them shortened
though as it makes them useless.

Simple solution is to chop them up into multiple chunks and rebuild the string
when we receive it.

#187


josephsavona:
:+1:

PR #255, opened 2015-10-09
[relay] Switch timing information to relative times

kassens:
I don't think absolute start times are really interesting. Relative times since
page load are a lot more interesting when optimizing page load time.

#187


josephsavona:
:+1:

PR #257, opened 2015-10-09
[relay] update RelayStore tab when the changes happen

kassens:
Previously, we were only showing the state of when the devtools were opened. Now
it updates when changes happen. We could probably improve the transport of the
data by sending a delta over the bridge, but that can be further optimization.

#187


josephsavona:
:+1:

Issue #91, opened 2015-05-20
Not showing up when inspecting a Chrome app


gurdiga:
I’m using React to develop the UI for a [Chrome app](https://developer.chrome.com/apps).

I have installed the [React Chrome Developer Tools](https://facebook.github.io/react/blog/2014/01/02/react-chrome-developer-tools.html), it shows up in Crome dev tools in the browser, but it doesn’t show up when I’m inspecting the Chrome app’s UI.

I imagine this use case may have not been expected/popular, but I’m still wondering if it’d be a way around it.

![screen shot 2015-05-20 at 11 21 50 am](https://cloud.githubusercontent.com/assets/53922/7721718/802bd1c4-fee2-11e4-94fc-4526c09d78a0.png)

spicyj:
I don't know offhand why this wouldn't work but if you figure it out, I'd be happy to take a patch to fix it.

gurdiga:
I have added a print statement in `injected/GlobalHook.js` and it seems that it’s not run within the other extension’s¹ pages, although `content_scripts.matches` in manifest is `<all_urls>`.

I’m wondering if you guys have seen this kind of issue before.

¹ and a packaged Chrome app like mine is essentially a Chrome extension because it’s `location.href` is `chrome-extension://fcfaacoiceonjcbkphlmimnmngcalndp/index.html`

rafales:
Same issue here. Is it possible that's because of CSP enforced for all chrome packaged apps?

Adding this code:
```
Object.defineProperty(
  window, '__REACT_DEVTOOLS_GLOBAL_HOOK__', {
    value: {
      inject: function(runtime){ this._reactRuntime = runtime },
      getSelectedInstance: null,
      Overlay: null,
    }
  }
)
```

to a separate script which gets loaded via the `script` tag makes it work for me. Hovering over components in the react inspector causes error because of the open() being not available for packaged apps (see screenshot).
![screenshot 2015-06-05 03 42 40](https://cloud.githubusercontent.com/assets/919812/7997922/32b5d30e-0b35-11e5-9c1d-6fff7c66214a.png)

gurdiga:
@rafales That is interesting insight, thank you for sharing! :+1:

spicyj:
Maybe https://code.google.com/p/chromium/issues/detail?id=77440.

rafales:
[...]

PR #263, opened 2015-10-23
Resurrect electron shell

frantic:
This reverts commit 3771a1e54aed9121cb0e0aeea130fd01850682ee and brings it up to date with recent devtools changes.

Test Plan:
```
npm run lint
flow
webpack
electron .
```
Opened RN app, made sure it connects. Seems like there are still a couple of edge case bugs and sometimes I need to reload either devtools or react app. Will work on tests for those later.

@jaredly @spicyj


spicyj:
What did you change? This would be easier to review as two commits (one straight revert, one with fixes).

jaredly:
I've done some more work on the electron shell in [this branch here](https://github.com/facebook/react-devtools/compare/master...jaredly:inspection-server?expand=1). It also allows you to inspect chrome apps and the like (re #91)

It's mostly the same as this PR, but with some more organization & misc fixes, as well as the "inspect chrome apps / webviews" stuff.

frantic:
@spicyj @jaredly what do you guys think about the approach in last 4 commits? I'm trying to isolate the electron UI into self-contained package that can be re-used in Atom/Nuclide and other Electron-based apps.

`shells/electron/node/index.js` still needs some cleanup and customization options (like port to listen, etc.)

@jaredly I like the approach you take with serving `embed.js` file, I'm happy to integrate it here too. In RN we could easily change `setupDevtools` to `eval(fetch('/embed'))`.

jaredly:
I like it

darrrk:
Any moves here?

matthewwithanm:
I think `run.js` needs to be updated to setup the Relay backend too. Otherwise, the `relay:check` call triggers a yellow box ("unknown call").

shells/electron/src/backend.js
```diff
@@ -0,0 +1,63 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+
+'use strict';
+
+declare var FOR_BACKEND: {
```
spicyj: Where does this come from? I don't see this being set anywhere.

shells/electron/src/backend.js
```diff
@@ -0,0 +1,63 @@
...
+  wall: Object;
+  resolveRNStyle: () => void;
+};
+
+if (!window.performance) {
```
spicyj: Doesn't electron provide this?

spicyj:
I don't have all the pieces in my head to review this properly, but let's just get this in.

shells/electron/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "main": "./index.js",
+  "scripts": {
+    "build": "webpack --config webpack.backend.js && webpack --config webpack.config.js"
```
iamdustan: what does this expect your path to be? Or is webpack expected to be installed globally?

shells/electron/package.json
```diff
@@ -0,0 +1,9 @@
...
+  },
+  "dependencies": {
+    "ws": "^0.7.2"
```
iamdustan: why 0.7.2? 1.1.0 is latest.

0.7.2 seems like this requires an older version of node. Should that be added to the engines field so consumers know of that requirement?

frantic: `1.1.0` is great (and doesn't require compiling C++ modules)! This PR is rather old, that's why it had `0.7.2`

shells/electron/src/backend.js
iamdustan: I removed it locally and everything was still okay. Running latest electron-prebuilt.

shells/electron/src/ui.js
```diff
@@ -0,0 +1,175 @@
...
+ *
+ */
+'use strict';
+
+var ws = require('ws');
+
+var installGlobalHook = require('../../../backend/installGlobalHook');
+installGlobalHook(window);
+var Panel = require('../../../frontend/Panel');
+var React = require('react');
```
iamdustan: s/react/react-dom ?

shells/electron/src/backend.js
frantic: This code is injected and `eval`ed inside another context that might not have `windows.performance`

shells/electron/src/backend.js
frantic: Will add link to https://github.com/facebook/react-native/blob/b00c77af8066cf74f05ccaca2b08c8952e1ae8a6/Libraries/Devtools/setupDevtools.js#L19

shells/electron/package.json
frantic: I think we can re-use `webpack` that was installed via root `react-devtools/package.json`

shells/electron/src/backend.js
iamdustan: ah yes. I copied that for here: https://github.com/iamdustan/react-hardware/blob/516d2a8854028577a7d7928729042c8738181bb0/src/devtools/setupDevtools.js

shells/electron/package.json
iamdustan: yeah. What you changed it to was what I had to run locally to reuse that webpack. :thumbsup:

iamdustan:
🎉

PR #265, opened 2015-10-26
Use performance.now() shim in the backend

empyrical:
I've been trying to get the inspector backend working in a barebones JS environment that doesn't have the `performance` object.

Other than this and `setTimeout` I don't seem to see any other HTML API dependencies

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -4,6 +4,7 @@
 .*node_modules/json-loader.*
 .*node_modules/node-libs-browser.*
 .*node_modules/webpack.*
+.*node_modules/fbjs/flow.*
 
 [libs]
 flow
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -13,6 +13,7 @@
 var consts = require('./consts');
 var hydrate = require('./hydrate');
 var dehydrate = require('./dehydrate');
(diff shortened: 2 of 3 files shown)


spicyj:
It is not okay to depend on react internals like that. That module should be available in fbjs though…

empyrical:
Whoops - looks like it is! Switched it over to the `fbjs` version and rebased.

In the Travis test, Flow doesn't seem to be seeing `fbjs/lib/performanceNow` - do you know what's going on there? Things built fine with webpack for me, so webpack was able to resolve it

spicyj:
Probably because you've ignored it. :) I'm actually not sure if we have a good way to make it all work together… Relay has some annoying hacks to make Flow work but it would also require changing the webpack configs. Maybe if you ignore just flow/include but not flow/lib then it works?

empyrical:
Yeah, that node module seems to be intended to be used with a module rewriting system like the one React uses and Flow didn't seem to like a lot of the stuff in there. The stuff in `lib` was already built with their `require()` calls rewritten so it works fine, but the stuff in the flow folder is unbuilt.

I misunderstood the `ignore` setting and didn't realize it caused it to not see the folder at all when resolving modules - just ignoring the `flow` folder did do the trick!

kassens:
Thanks!

PR #266, opened 2015-10-26
[relay] enable selection of query in QueryViewer

kassens:
It can be useful to copy parts of the query, we shouldn't block selection here.


spicyj:
:+1:

PR #269, opened 2015-11-04
Update loading screen for Firefox review

### LINKED ISSUE #243: Fix up minor problems from Firefox addon review
We got preliminarily approved with this feedback:

> 1) Please reduce your use of innerHTML. There are multiple instances where it should be easy to replace with other DOM methods that are faster and safer.
>
> 2) I get "Connecting to devtools" when I'm on a website that doesn't support React. You should show a clearer message in this case, since the current message suggests that something else is going to happen if I wait long enough.

#1 probably refers to a couple uses in frontend/Highlighter/ and shells/firefox/panel/run.js. Haven't looked at all at #2 but I imagine it's not hard either.

spicyj:
Fixes #243: don't use innerHTML, make messaging clearer.

### DIFF
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -206,6 +206,10 @@ class Panel extends React.Component {
 
   render() {
     if (this.state.loading) {
+      // TODO: This currently shows in the Firefox shell when navigating from a
+      // React page to a non-React page. We should show a better message but
+      // properly doing so probably requires refactoring how we load the panel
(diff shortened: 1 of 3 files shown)


kassens:
Is React internally not using `innerHTML` for the initial render?

This is nicer either way.

spicyj:
It is, but it sanitizes it so it's hard to introduce security holes, which was (I believe) their concern here.

PR #277, opened 2015-11-10
Tweak Firefox shell again for 44 and later

jryans:
In #218, I updated some internal Firefox DevTools paths for 44 and later.

On the Firefox DevTools team, we made [one more change][1] that also landed in 44, so here's another update to cover that. No other changes like this are anticipated in the near future.

[1]: https://bugzil.la/1203159

### DIFF
--- a/shells/firefox/main/trackSelection.js
+++ b/shells/firefox/main/trackSelection.js
@@ -12,10 +12,10 @@
 const { Cu } = require('chrome');
 
 // The path to this file was moved in Firefox 44 and later.
-// See https://bugzil.la/912121 for more details.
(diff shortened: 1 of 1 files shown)


spicyj:
Thanks! Will addons on AMO will be updated automatically or do I need to cut a new release?

jryans:
> Thanks! Will addons on AMO will be updated automatically or do I need to cut a new release?

A new release is needed. The previous path value is still supported by a shim in 44 and later. Eventually we will turn on console warning for the old values, and even later remove the shims, but I expect it will be a while before any of that.

I am also an AMO add-on reviewer, so I can check out your next release when it's up for review if you ping me somewhere.

spicyj:
Thank you!

PR #281, opened 2015-11-12
[relay] Disable RelayStore updates when tab is not open

kassens:
This stops sending every Relay node update over the wire when the Relay tab is not open.


yungsters:
:+1:

PR #282, opened 2015-11-12
React components should extend React.Component

kassens:
I don't see anymore warnings about this after this PR.


josephsavona:
:+1:

PR #283, opened 2015-11-12
[relay] Group queries by request

kassens:
This displays Relay queries sent in the same network request with a header. It also sets the start time to be the same. It's otherwise slightly confusing to display times with a few ms difference.

### DIFF
--- a/plugins/Relay/QueryList.js
+++ b/plugins/Relay/QueryList.js
@@ -27,15 +27,34 @@ class QueryList extends React.Component {
     if (!this.props.queries.count()) {
       return <div style={styles.empty}>No Relay Queries logged</div>;
     }
-    var rows = this.props.queries.valueSeq().map((q, i) => (
-      <Query
-        data={q}
-        isSelected={q.get('id') === this.props.selectedQuery}
-        key={q.get('id')}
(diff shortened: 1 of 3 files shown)


plugins/Relay/QueryList.js
```diff
@@ -27,15 +27,34 @@ class QueryList extends React.Component {
...
-        onSelect={() => this.props.selectQuery(q.get('id'))}
-      />
-    )).toArray();
+
+    const rows = [];
+    let odd = false;
+    let lastRequestNumber = -1;
+    this.props.queries.forEach(query => {
+      const requestNumber = query.get('requestNumber');
+      if (lastRequestNumber !== requestNumber) {
```
yungsters: Would it not be better to first partition `this.props.queries` by `requestNumber` before rendering them? (What if queries from different `requestNumber` values are interleaved?)

kassens: Good point. I looked at it and makes the code more complex when I don't assume the right order. (For example, how should they be ordered?)

All queries of a request are added when the request is sent and shouldn't be interleaved. If this should happen, the interface will display the same request header twice and we can take a look then how to resolve that case.

PR #284, opened 2015-11-12
Upgrade to Babel 6

kassens:


spicyj:
None of the webpack configs needed the stage?

kassens:
This is now moved into a preset and I pushed the config to the babelrc that is shared for all of them (including jest).

spicyj:
I meant: before, .babelrc specified stage 0 but so did the webpack configs. Now only .babelrc does. Were the webpack configs redundant?

kassens:
They might've been redundant. Alternatively, it might be an update that came with `babel-loader` v6 that it reads the `.babelrc`. I tested the plugin and it worked and everything compiled (it didn't without the `.babelrc`)
¯\\\_(ツ)\_/¯

iamdustan:
I think the webpack config was redundant. I’m still using Babel 5.x witg webpack and the only config is in .babelrc.

PR #285, opened 2015-11-17
Fix mktemp on Linux

camabeh:
From Linux man page: TEMPLATE must contain at least 3 consecutive 'X's in last component.


kassens:
Thanks for the fix, I'll add a short note about it being needed for GNU mktemp.

Issue #268, opened 2015-10-29
Incorrect URL in warning message when using React 0.12 or below


peterjmag:
When I use react-devtools on a site with an outdated version of React (0.12 in this case), I get the following warning in my console:

```
You are using a version of React with limited support in this version of the devtools.
Please upgrade to use at least 0.13, or you can downgrade to use the old version of the devtools:
instructions here https://github.com/facebook/react-devtools/tree/devtools-next#how-do-i-use-this-for-react--013
```

This URL gives me a 404: https://github.com/facebook/react-devtools/tree/devtools-next#how-do-i-use-this-for-react--013. Perhaps the devtools-next branch was deleted? I was hoping to submit a pull request with an updated URL, but I can't find a "How do I use this for React 0.x" section in the current readme.

I'm using the latest stable version of the extension (0.14.4).

jaredly:
Good catch! Looks like it got removed from the FAQ section accidentally. We just need to have a link to an old `.crx` extension file (not sure where to dig that up)

peterjmag:
Looks like the [Chrome Web Store doesn't have an archive](http://stackoverflow.com/a/10744580/349353) (surprise, surprise), but perhaps we could just build one from an older version of the repo and then just host it somewhere else? If anyone still has their older clone lying around, that is.

jaredly:
Here's the last commit of the old version:
https://github.com/facebook/react-devtools/tree/329363301da8fad70d1b49f830886c55ed53b1ef
A build off of that should do the trick

peterjmag:
Sure, I built the version at 329363301da8fad70d1b49f830886c55ed53b1ef and I've added it to a personal repo here: https://github.com/peterjmag/react-devtools-issue-228. I'd like to be able to delete this repo later if necessary, so perhaps the actual packed extension should be hosted somewhere else (i.e. a branch on react-devtools, or FB's CDN)?

For now, here's a direct download link to the older version without the warning message in React 0.12 and below: https://github.com/peterjmag/react-devtools-issue-228/raw/master/react-devtools-329363301da8.crx

Once it's hosted somewhere with a "stable" URL, we can update the readme and the warning message URL.

peterjmag:
Ah, and I should probably also add a note that this version needs to be installed by dragging / dropping it on [chrome://extensions](chrome://extensions), since it can't be installed automatically via download due to Chrome's extension signing stuff (I assume): http://stackoverflow.com/a/14454311/349353

[...]

PR #290, opened 2015-11-20
Fix incorrect URL in warning message for old versions of React

peterjmag:
Addresses facebook/react-devtools#268.

## Todo

- [ ] Tag https://github.com/facebook/react-devtools/tree/329363301da8fad70d1b49f830886c55ed53b1ef as `v0.13.1` (assuming that's the right commit)
- [ ] Create a new (well, old) release
- [ ] Attach the packed extension from https://github.com/peterjmag/react-devtools-issue-228/blob/master/react-devtools-329363301da8.crx

Ping @zpao @jaredly (since they have access to the FB GitHub org)

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,18 @@ or by inspecting [Facebook](https://www.facebook.com/).
 
 Currently iframes and Chrome apps/extensions are not inspectable.
 
+### How do I use this for React < 0.13?
+
+Download and manually install [an older version of the Chrome extension](https://github.com/facebook/react-devtools/releases/tag/v0.13.1).
+
+If you try to open the downloaded `crx` file directly in Chrome, you will see
+the following message:
+
+> Apps, extensions and user scripts cannot be added from this website.
+
+Instead, follow [these instructions](http://stackoverflow.com/a/11879334/349353)
+to install the extension manually.
+
 ### ProTips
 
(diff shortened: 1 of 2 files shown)


peterjmag:
Ah, forgot about Firefox. I assume we can't build an older version for Firefox? If we can, then maybe I should add some more instructions for FF users, but if not, then I can just clarify in the readme that the older version's for Chrome only.

spicyj:
Right, the old version wasn't compatible with Firefox.

peterjmag:
Thanks @spicyj, I've added a note. I'll test all the links once the old release is added and this PR is merged.

peterjmag:
@kassens @spicyj Just going through some of my older open issues. Any updates on this one? I believe it's ready to merge, it just needs a bit of followup from a maintainer once it's in master (see [the checklist above](https://github.com/facebook/react-devtools/pull/290#issue-118015366)).

gaearon:
I’m going through the PRs, and it feels like this has been stale for a while. I appreciate the effort but I’ll close unless somebody is ready to jump on it and provide more context. We’ve recently shipped 15.0, so 0.12 is a really old version by now, and I don’t think it’s worth spending time correcting the mistakes we’ve made with regards to it. Cheers!

PR #293, opened 2015-12-01
Surface non-enumerable properties

tquetano-r7:
Use Object.getOwnPropertyNames instead of traditional for-in loop so that non-enumerable property names are surfaced. This can provide additional metadata for debugging purposes that will not get applied to the React element itself.


spicyj:
Did you test this? I wouldn't expect this to work as-is.

tquetano-r7:
No, I didn't do a full test with the rest of devtools, I just tested the implementation of it on a standard object to ensure that it looped correctly, and that non-enumerables were coming through when not a for loop. Why don't you think this will work as-is? Is there a downstream impact I'm not seeing? Cuz it looks like a pretty straightforward change.

spicyj:
I was thinking that that's not the right loop, but maybe I was mixing up hydrate and dehydrate. This also might ignore non-own properties now? I don't recall what the current behavior is there.

I am guessing this also won't show symbol keys since the UI would need to use a similar loop instead of a for/in. If this works for the non-enumerable string key case then it's okay as long as it doesn't regress anything.

tquetano-r7:
Per MDN:

tquetano-r7:
Per MDN:

Object.getOwnPropertyNames() returns an array whose elements are strings corresponding to the enumerable and non-enumerable properties found directly upon obj. The ordering of the enumerable properties in the array is consistent with the ordering exposed by a for...in loop (or by Object.keys()) over the properties of the object. The ordering of the non-enumerable properties in the array, and among the enumerable properties, is not defined.

Based on the verbiage, I took it basically as "everything that for-in does plus non-enumerable properties". That said, the build failed because of a possible undefined on path, which I didn't touch (that section was copy pasta), so unsure how you want to proceed with that. Everything else was successful in the build.

spicyj:
getOwnPropertyNames returns only the own properties:

image

tquetano-r7:
Alright I'll do a quick mod and update to include non-own props.

tquetano-r7:
Just to make sure I don't cause any more unnecessary builds ...
```
var res = {},
    names = Object.getOwnPropertyNames(data);
    
  for (var name in data) {
    if (names.indexOf(name) === -1) {
      names.push(name);
    }
  }
  
  names.forEach((name) => {
    res[name] = dehydrate(data[name], cleaned, path.concat([name]), level + 1);
  });
  
  return res;
```

Per MDN again, it looks like the only way I could grab both is a combination of a for loop and getOwnPropertyNames.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties

If you're cool with that, I'll re-merge.

tquetano-r7:
Alright I did it right this time, verified with ur lint script that it passes all that jazz, but still expecting it to fail because of that possible undefined on path that existing prior to my change.

jaredly:
+1 lgtm

spicyj:
You probably want to walk through the prototype chain so you get non-enumerable properties on the prototype chain too?

spicyj:
Also, can you please squash your commits when you're done?

tquetano-r7:
Not sure if I agree with getting non-enums on the prototype, it will produce a lot of noise:

```
["constructor", "toString", "toLocaleString", "valueOf", "hasOwnProperty", "isPrototypeOf", "propertyIsEnumerable", "__defineGetter__", "__lookupGetter__", "__defineSetter__", "__lookupSetter__", "__proto__"]
```

That is output of a simple object, and that output is likely to be different on a browser-to-browser basis so I think it'd be really hacky to try and filter them out or something. I can squash the commits, but I wanna make sure you guys are cool with not traversing prototype, basically keeping the PR as-is, before I do anything else.

spicyj:
Chrome shows each prototype separately:

image

I thought we already had code to do this…

jaredly:
we do get the proto separately

jaredly:
And it's dehydrated separately, see [here](https://github.com/facebook/react-devtools/blob/master/agent/Bridge.js#L354). This PR should be good as-id

spicyj:
In what scenario do we want all own properties plus all enumerable properties (excluding non-enumerable, non-own properties)? That's what this PR does. Perhaps we don't need any non-own properties here at all?

jaredly:
@spicyj methods of an instance are frequently non-own, and I imagine people would expect them to show up.

tquetano-r7:
Agreed, prototype methods would be expected. The original use case was merely to surface non-enumerables for the object itself, and while there is a valid use case for that, I don't think that use case extends for non-enumerables that exist on the prototype. I think that is far more niche.

The only thing this PR adds compared to the original setup is the ability to see own non-enumerables. Personally I would say that if someone wants to see non-enumerables on the prototype, it can be added in as enhancement. Just my two cents, though.

spicyj:
I'm sorry for all the back and forth on this. I haven't had time to try this out myself so I am relying on your descriptions in order to understand the behavior here.

It sounds like currently, the devtools show the properties from each prototype separately, like Chrome's does. So if I have an object like:

```
var o = Object.create({px: 1, py: 2});
o.x = 3;
o.y = 4;
```

then `o` looks like this in the devtools:

```
- x: 3
- y: 4
- __proto__
  - px: 1
  - px: 2
  - __proto__
    - __defineGetter__: fn
    - __defineSetter__: fn
    - ...
```

Right now, we are missing nonenumerable properties. Moving to Object.getOwnPropertyNames fixes this, but here is my confusion: If we are examining each prototype separately, then at each level we should only look at own properties, but it seems like we're doing a for/in loop.

What am I missing here?

jaredly:
@spicyj oh yeah you're totally right. +1 for dropping the `for in`

tquetano-r7:
I'll mod my PR again ... le sigh ... :P

gaearon:
I’d be happy to review and test this, but it currently fails Flow check. Can you please fix it?
You can run `npm run typecheck` to figure out what fails.

tquetano-r7:
So the update has been pushed, turns out that doing the `.forEach` messed with the scope and tyecheck failed on the `path.concat` declaration. I switched it to a traditional `for` loop and it passed.

gaearon:
There’s something wrong with the PR, as it lists 119 commits and removes recent changes. Can you take another look at what’s up with the branch and fix it?

tquetano-r7:
Sorry man, this PR is like ... months old, and it was like 106 commits behind. I'm gonna close this and open a new one off of fresh master, it will be way less problematic.

Issue #275, opened 2015-11-09
Unable to resize smaller than widest component


wesbos:
If a component has a couple of props, React Dev Tools makes the DOM pane as wide as the widest one. This makes it really hard to use them when docked to the side. 

Ideally something like this would just cause wrapping or horizontal scroll. I don't think this was an issue until about the last week. 

![Can't see the props / state pane with regular sizing](http://wes.io/dlNk/content)

![I have to expand it this wide just to see the props and state pane](http://wes.io/dlES/content)

spicyj:
Weird, it should do a horizontal scroll.

spicyj:
This is what I see on the React site:

![image](https://cloud.githubusercontent.com/assets/6820/11043161/3ac6274e-86d0-11e5-8d19-f28e837f9ae8.png)

jaredly:
flexbox implementations are evolving (well, coming more in line with the spec), and result in issues like this. I think adding a `min-width: 0` or two would fix this problem. What version of Chrome are you running?

wesbos:
I'm on Chrome Canary - I think I've figured out why its doing it. When I try resize, either to the left or the right, the width of the props/state panel just grows indefinitely. 

If I take the inline width off, it snaps to overflow scroll again. But upon resizing it, the inline width comes back.

Here is a video: http://wes.io/dl31

FutureKode:
I am also seeing this issue Chrome Version 48.0.2552.0 dev (64-bit)

wesbos:
Anyone else having this issue? It still persists for me.

ESWAT:
Yup, still see this in 49.0.2568.0 canary.

sillero:
Still an issue on Chrome (Mac) Version 49.0.2572.0 canary (64-bit)

revolunet:
Same here :)

The panel shows a resize cursor but doesnt work at all, stucked on the longer line width.

Ajwah:
Same here as well.
The panel doesn't allow me to resize.

<img width="1435" alt="screenshot 2015-12-14 08 10 25" src="https://cloud.githubusercontent.com/assets/11848116/11782767/1d801276-a241-11e5-82d5-e40498bf7a38.png">

wesbos:
@spicyj Do you know where in the codebase the resizing code is? This is still an issue for me and a pretty big pain on smaller screens.

revolunet:
[...]

PR #296, opened 2015-12-14
Fix resizing issue on Chrome

sillero:
This fixes the panel resizing issue on Chrome 48+ described on #275

### DIFF
--- a/shells/chrome/panel.html
+++ b/shells/chrome/panel.html
@@ -16,10 +16,12 @@
                 padding: 0;
                 flex: 1;
                 display: flex;
+                width: 100%;
             }
             #container {
                 display: flex;
                 flex: 1;
+                width: 100%;
             }
         </style>
     </head>


spicyj:
Can you test if `align-self: stretch` on body and #container work equally well?

sillero:
`align-self` and `align-items` are related to the cross-axis. MDN explains it better
> The CSS align-items property aligns flex items of the current flex line the same way as justify-content but in the perpendicular direction.

Tested, didn't work.

spicyj:
I assumed flex: 1 would work for the main axis and align-self: stretch would work for the cross axis… weird.

sillero:
Removing `white-space: nowrap` would make things work without `width: 100%`. But I'm not sure you guys would want that.

spicyj:
Okay, I don't really understand this but if it works it works. :(

PR #303, opened 2015-12-21
Update dependencies and pin versions

kassens:
Using fixed versions to make builds more reproducible and updates of dependencies more transparent.

[node-check-updates](https://www.npmjs.com/package/npm-check-updates) makes it pretty easy to explicitly update dependencies.


spicyj:
I guess this is fine but maybe you want to just commit a shrinkwrap file instead so that transitive dependencies are also pinned?

PR #304, opened 2015-12-21
Update flow to 0.20.1

kassens:
The biggest change required here was that `Map.prototype.get` now returns nullable values and we have to guard agains that in code. I've opted to just silence these errors but could also change these to `invariant`. Invariant seemed less desirable for a plugin.


spicyj:
:+1:

PR #305, opened 2015-12-22
Add permissions for file://

### LINKED ISSUE #172: Chrome devtools don't work on file: URLs

kassens:
Fixes #172

### DIFF
--- a/shells/chrome/manifest.json
+++ b/shells/chrome/manifest.json
@@ -20,7 +20,8 @@
   },
 
   "permissions": [
-    "http://*/",
+    "file:///*",
+    "http://*/*",
     "https://*/*"
   ],
 


spicyj:
Did you test that this works locally?

kassens:
Yep, it seems to work and it also doesn't appear to ask for extra permissions when updating. Hopefully, that will not be different on the app store.

spicyj:
Sweeeet. Thank you.

jimfb:
@spicyj Does this mean we now want to back out https://github.com/facebook/react/pull/5610 ?

spicyj:
If this works, then yeah.

spicyj:
If this works when it's in the store, then yeah.

kassens:
Only seems to work when it's installed locally and blocked on the store install.

jimfb:
Well, depending on how you define "work": https://github.com/facebook/react-devtools/issues/172#issuecomment-166741417

It "works" if you explicitly enable it, which is a complex interaction that most people will neglect to do.

PR #311, opened 2016-01-11
Debouncing search text change

ncthis:
When using the search box in devtools, if my keystrokes are too quick, then they are dropped. For an example, typing "Table" would amount to:

image

Therefore, I propose that the `onChangeSearch` function be `debounce`d.

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -16,6 +16,7 @@ var assign = require('object-assign');
 var nodeMatchesText = require('./nodeMatchesText');
 var consts = require('../agent/consts');
 var invariant = require('./invariant');
+var debounce = require('../utils/debounce');
 
 import type Bridge from '../agent/Bridge';
 import type {DOMEvent, ElementID} from './types';
@@ -83,6 +84,7 @@ class Store extends EventEmitter {
   _nodesByName: Map;
   _eventQueue: Array<string>;
   _eventTimer: ?number;
+  _emitSearchRootsChange: Function;
 
   // Public state
   bananaslugState: ?Object;
@@ -150,6 +152,7 @@ class Store extends EventEmitter {
     this._establishConnection();
     this._eventQueue = [];
     this._eventTimer = null;
+    this._emitSearchRootsChange = debounce(this.filterSearchRoots, 100);
   }
 
   emit(event: string): boolean {
@@ -199,10 +202,18 @@ class Store extends EventEmitter {
 
   changeSearch(text: string): void {
     var needle = text.toLowerCase();
+
     if (needle === this.searchText.toLowerCase()) {
       return;
     }
-    if (!text) {
+
+    this.searchText = text;
+    this.emit('searchText');
+    this._emitSearchRootsChange(needle);
+  }
+
+  filterSearchRoots(needle: string): void {
+    if (!needle) {
       this.searchRoots = null;
     } else {
       if (this.searchRoots && needle.indexOf(this.searchText.toLowerCase()) === 0) {
@@ -225,9 +236,9 @@ class Store extends EventEmitter {
         }
       });
     }
-    this.searchText = text;
-    this.emit('searchText');
+    
     this.emit('searchRoots');
+
     if (this.searchRoots && !this.searchRoots.contains(this.selected)) {
       this.select(null, true);
     } else if (!this.searchRoots) {
--- a/utils/debounce.js
+++ b/utils/debounce.js
@@ -1,0 +1,31 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+'use strict';
+
+function debounce(fn: Function, delay: number): Function {
+  var timeout;
+  return function() {
+    var context = this;
+    var args = arguments;
+    var later = function() {
+      timeout = null;
+      fn.apply(context, args);
+    };
+
+    clearTimeout(timeout);
+    timeout = setTimeout(later, delay);
+    if (!timeout) {
+      fn.apply(context, args);
+    }
+  };
+}
+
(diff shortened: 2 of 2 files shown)


jaredly:
@ncthis thanks for this! looks great to me

jaredly:
although a quarter-second might be a bit long -- how does it feel w/ a 100ms debounce?

ncthis:
@jaredly Sounds good. I've reduced it to 100. Thanks for the review!

andrei-picus-hs:
@ncthis @jaredly this is not the correct fix. You've debounced the callback that emits the input's content. What that means is that if I type really fast then only the last letter will be emitted.

Here's trying to type `SearchPane`:

debounced

jaredly:
oh good call. yeah, we need the debouncing at the place that we filter things

ncthis:
Ok, will look into it.

On Mon, Jan 11, 2016 at 22:34 Jared Forsyth <notifications@github.com>
wrote:

> oh good call. yeah, we need the debouncing at the place that we filter
> things
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/facebook/react-devtools/pull/311#issuecomment-170618410>
> .
>
-- 
- Nadeesha

ncthis:
I have changed the implementation to debounce at store, before filtering and after `emit`ting `searchText`. This improves the situation greatly. 

But for superhuman typing speeds, still, there remains a problem. And, it's to do with the `emit` function's `setTimeout`.
```js
setTimeout(() => {
        this._eventQueue.forEach(evt => {
          EventEmitter.prototype.emit.call(this, evt);
        });
        this._eventQueue = [];
        this._eventTimer = null;
}, 50);
```

Basically, when keystrokes happen in between that 50ms, they tend to be lost. (Because store state had not changed and the changed value has not propagated to the component through `props`). You can verify this if you try typing "asdf" really fast, and note the missing key(s). If you move the timeout to a lower value, it progressively becomes better.

I know this would hardly ever make an impact for the vast majority of users, but I'm just noting it here because it impacts what I set out to do. If a solution for this can be recommended, I'd be more than happy to work on it as well.

frontend/Store.js
```diff
@@ -198,10 +200,14 @@ class Store extends EventEmitter {
...
 
   changeSearch(text: string): void {
+    this.searchText = text;
+    this.emit('searchText');
+    this._emitSearchRootsChange(text);
+  }
+
+  filterSearchRoots(text: string): void {
     var needle = text.toLowerCase();
-    if (needle === this.searchText.toLowerCase()) {
```
jaredly: why remove this?

ncthis: By the time it hits this line, `this.searchText` is guaranteed to be updated. Therefore `needle` will always be equal to `this.searchText`.

jaredly: right -- it should be at the top of `changeSearch` (where it started)

ncthis: Understood. Will change that.

frontend/Store.js
```diff
@@ -150,6 +151,7 @@ class Store extends EventEmitter {
     this._establishConnection();
     this._eventQueue = [];
     this._eventTimer = null;
+    this._emitSearchRootsChange = debounce(this.filterSearchRoots, 100);
```
Timothee: From the error message in [Travis build](https://travis-ci.org/facebook/react-devtools/builds/101731744), it seems you should declare that function on `Store` as an empty function.

RaitoBezarius:
Heads up: still an annoying issue on big React tree, could I lend a hand, maybe?

ncthis:
@RaitoBezarius I will try to push this within the next few hours.

frontend/Store.js
ncthis: @Timothee Thanks for the heads up. I actually needed to add the type check function with `Function` in `Store`.

Timothee: Got it. Cool because that means we might get this thing merged soon :simple_smile:

ncthis:
@jaredly Any chance on getting this merged?

ncthis:
Sorry about the persistent pestering. But can someone from the org let me know if this could be merged at all?

spicyj:
I took a brief look just now, sorry for not commenting earlier. I was under the impression that it was laggy because we were waiting for some async operation over the bridge to return, but it looks like you're only moving the searchText assignment and event to execute earlier in the same event loop – which would seem to help only if we're CPU-bound in this interaction which I didn't think was the case. I didn't get to try this out yet locally though.

ncthis:
I think I found the issue to be CPU bound. It fixed the issue for me
locally. Please try it locally. I will be here if it needs more work.

On Wed, Mar 30, 2016, 23:24 Ben Alpert <notifications@github.com> wrote:

> I took a brief look just now, sorry for not commenting earlier. I was
> under the impression that it was laggy because we were waiting for some
> async operation over the bridge to return, but it looks like you're only
> moving the searchText assignment and event to execute earlier in the same
> event loop – which would seem to help only if we're CPU-bound in this
> interaction which I didn't think was the case. I didn't get to try this out
> yet locally though.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/facebook/react-devtools/pull/311#issuecomment-203550874>
>
-- 
- Nadeesha

spicyj:
I just patched this locally and it still drops keystrokes in my testing. If I type "div" quickly I get "iv" or "v".

ncthis:
@spicyj I think this relates to my [comment above](https://github.com/facebook/react-devtools/pull/311#issuecomment-170750309). If there's a recommendation to improve that emit function, I'm more than happy to work on it.

spicyj:
Thanks, I totally missed that comment.

ncthis:
@spicyj Would you have an update on this for me? I can fix the emit function if I get some direction.

ianks:
This is an important feature. This extension is almost unusable without it.

spicyj:
Sorry, I haven't had time to think more about this. (If I knew exactly what to fix I probably would have done it myself…)

gaearon:
Closing as the proposed fix doesn’t fix the underlying issue.

gaearon:
Follow up fix: #412.

Issue #313, opened 2016-01-16
Uncaught TypeError: Cannot assign to read only property 'prototype' of function Map()


huang47:
repro steps

1. install react-devtools(built from latest master) on chrome (lastet stable)
2. open https://facebook.github.io/react/
3. open inspector and see following error and stuck at `Connecting to React...`

![screen shot 2016-01-16 at 3 42 16 pm](https://cloud.githubusercontent.com/assets/157390/12375099/be155a80-bc67-11e5-985e-28525979b64b.png)

dckesler:
I'm also having this issue. Both with Map and Set. When I remove
```javascript
Map: 'window.__REACT_DEVTOOLS_GLOBAL_HOOK__.nativeMap',
Set: 'window.__REACT_DEVTOOLS_GLOBAL_HOOK__.nativeSet',
```
from `webpack.backend.js` it seems to fix it for now.
Both errors seems to be related to `immutable.js`.

huang47:
@dckesler fixed in #317

PR #317, opened 2016-01-18
Revert "use webpack DefinePlugin to dramatically simplify things"

huang47:
This reverts commit d748a51b9562dd6253f20b8795f765a9b1e5000c.

To address the issue raised in #313 

I am not a webpack expert and seems like this is caused by using DefinePlugin which overrides all `Map` call site to `window.__REACT_DEVTOOLS_GLOBAL_HOOK__.nativeMap` (same to `Set`)

rollback to ProvidePlugin which unbreak BananaSlug plugin release. @jaredly do you see any potential problems by reverting this commit?

### DIFF
--- a/shells/chrome/helpers/map.js
+++ b/shells/chrome/helpers/map.js
@@ -1,0 +1,12 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ */
+'use strict';
+
+module.exports = window.__REACT_DEVTOOLS_GLOBAL_HOOK__.nativeMap;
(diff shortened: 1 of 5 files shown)


spicyj:
Does this not have the same replacement problem as the DefinePlugin way? I would expect them to be about the same.

huang47:
according to the test cases [DefinePlugin](https://github.com/webpack/webpack/tree/master/test/configCases/plugins/define-plugin) / [ProvidePlugin](https://github.com/webpack/webpack/tree/master/test/configCases/plugins/provide-plugin) which looks pretty similar to me, need to investigate more why it only works for `ProvidePlugin`

jaredly:
What changed to cause #313? Is it a change in React core?

spicyj:
No, Hedger pulled in immutable which uses Map as a local variable name.

jaredly:
Ok yeah so it looks like `ProvidePlugin` only steps in when the name (e.g. Map) is *undefined* in the current context, whereas `DefinePlugin` just overwrites everything.
looks good to me

huang47:
@jaredly DefinePlugin seems to override all Map to nativeMap and ProvidePlugin does similar things. However, when I test it with ProvidePlugin which doesn't trigger following error message, any insights?

```shell
Uncaught TypeError: Cannot assign to read only property 'prototype' of function Map
```

PR #318, opened 2016-01-24
Fix scroll to selected element

### LINKED ISSUE #312: Scroll to selected element
When switching from the Elements panel, the corresponding component is automatically selected in the React pane. But it's often offscreen if a page has many roots (e.g., facebook.com). We should automatically scroll the element to be visible.

Shraymonks:
Fixes #312


spicyj:
Thank you!

PR #324, opened 2016-02-13
fixed tabbed resizing bug on Chrome

flipside:
This fixes a tabbed panel resizing issue on Chrome 48+ based on #275 and #296. This bug is present when there is a tabbed panel and can be seen on any site with Relay.


kassens:
Thanks for finding this and sending a fix :+1:

flipside:
No problem

PR #326, opened 2016-02-17
Update jest to 0.9.0-fb2

kassens:


spicyj:
Travis is failing.

kassens:
I re-started the run a few times and it's not a transient issue. Jest is watching too many files. Created facebook/jest#714 to see if it's best fixed in Jest directly.

PR #328, opened 2016-02-18
Add element's `key` and `ref` to tree and prop views

### LINKED ISSUE #148: No way to know element key
There is no way to know an element's key in React tab. It is omitted on the left pane and is not a prop anymore to be shown in the right pane.

image

As a walk-around, you can look for `$key` in `data-reactid` attribute in Chrome's own Elements tab or use `$r._reactInternalInstance`, but these are not the best solutions.

### LINKED ISSUE #226: Show refs in props
Refs are stripped out by React, so they don't get included in the props listing (inline or sidebar). It would be nice to get those back.
Not sure where to get them from though.

raised by #225

Timothee:
Fixes #148.
Fixes #226.

If a `key` or `ref` is set on the React element, it's displayed before
the `props` in the tree view and in the `PropState` sidebar.

If none is set, nothing is changed.

Also removed a piece of code `Node.js` that would never run. (condition
is checked in the previous block that `return`s)

### DIFF
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -24,6 +24,8 @@ function getData(element: Object): DataType {
   var updater = null;
   var name = null;
   var type = null;
+  var key = null;
+  var ref = null;
   var text = null;
   var publicInstance = null;
   var nodeType = 'Native';
@@ -52,6 +54,8 @@ function getData(element: Object): DataType {
 
   if (element._currentElement) {
     type = element._currentElement.type;
+    key = element._currentElement.key;
+    ref = element._currentElement.ref;
     if (typeof type === 'string') {
       name = type;
     } else if (element.getName) {
@@ -95,6 +99,8 @@ function getData(element: Object): DataType {
   return {
     nodeType,
     type,
+    key,
+    ref,
     name,
     props,
     state,
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -54,7 +54,9 @@ function getData(element: Object): DataType {
 
   if (element._currentElement) {
     type = element._currentElement.type;
-    key = element._currentElement.key;
+    if (element._currentElement.key) {
+      key = String(element._currentElement.key);
+    }
     ref = element._currentElement.ref;
     if (typeof type === 'string') {
       name = type;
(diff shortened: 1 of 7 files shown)


Timothee:
CLA signed.

Timothee:
@jaredly @kassens anything I can do to get this PR looked at? Thanks!

Timothee:
@spicyj any chance to get this looked at? Thanks!

Timothee:
😐

jaredly:
@Timothee sorry to take so long on this. It looks great! one request: can you make it handle the "ref is a function" case gracefully?

Timothee:
> React also supports using a string (instead of a callback) as a ref prop on any component, although this approach is mostly legacy at this point

I suppose I've been using legacy code with regards to refs then 😄 

@jaredly if `ref` is a function, should I just not display anything at all? From [the examples](https://facebook.github.io/react/docs/more-about-refs.html), it seems that with a function, there's nothing to display really…

Timothee:
Actually, re-reading #226:

> functions don't have a meaningful representation in the sidebar

but

> we might be able to get a meaningful representation of functions from name and displayName, or if the VM permits, even line number + source link

I'll look into that last one a bit, but otherwise will just skip it if it's a function. (unless you have better ideas)

jaredly:
that sounds fine to me!

Timothee:
@jaredly updated. Now the following JSX:
```js
var Container = React.createClass({
    render: () => {
        return (
            <div>
                <div key='withKey'>with a key</div>
                <div ref='stringRef'>with a string ref</div>
                <div ref={()=>{}}>with an unnamed function ref</div>
                <div ref={function namedFn() {}}>with a named function ref</div>
                <div key='withKeyAndRef' ref={function namedFn() {}}>with a named function ref and a key</div>
            </div>
        );
    }
});
```
is shown like this:

screen shot 2016-04-29 at 1 16 39 pm


The replacement of the unnamed function into `fn()` is handled by `PropVal`.

montogeek:
Please merge this :)

Timothee:
@jaredly bump 👀

gaearon:
This is in. Thank you so much!

montogeek:
THANKS!

Issue #337, opened 2016-02-29
Does "Trace React Updates" respect shouldComponentUpdate short-circuits?


stipsan:
It looks to me as if `shouldComponentUpdate` checks counts as an **update** in the devtools.
Even if a component return false on `sCU` it will flash as updated. Child components of a falsy `sCU` won't flash, suggesting that `sCU` is probably not tracing renders per-se but the `componentWillReceiveProps` part of the lifecycle.

Here's a codepen to see what I'm talking about: http://codepen.io/stipsan/pen/grbaEr

With "Trace React Updates" flipped on I'd expect the first input to flash itself and its parent when you change the value. The second input guard itself with a `sCU` for its local state.
But it flashes, suggesting the `sCU` returned true when it actually returned false.
If the second input is changed, it'll only update its local state and the devtools will show a trace for just the second input, just as you'd expect.

If this is the intended behavior, then it should be much more clear so that lesser experienced devs isn't mislead into thinking pure components are rendering when they aren't.

jaredly:
you're right -- it's confusing.
This has to do with the way that the devtools instruments react to get information on prop/state updates, etc.
The hook [we decorate](https://github.com/facebook/react-devtools/blob/master/backend/attachRenderer.js#L117) is `performUpdateIfNecessary`, and we don't have a clean way of knowing whether `shouldComponentUpdate` returned false...

If you want to add a section to the Readme about this (maybe under FAQ) that would be great!

To actually fix this problem, we'd probably need to make a change to react core, perhaps making [`performUpdateIfNecessary`](https://github.com/facebook/react/blob/92530b4ddcc77be49a50ce4cbffb32ee38a98247/src/renderers/shared/reconciler/ReactReconciler.js#L142) return a bool indicating whether an update was necessary

stipsan:
Sure, I'll add a section to the FAQ about this. 
Thanks for the explanation!

wmertens:
I think it would be very nice if this was an accurate indication of updates, since it's such a visual way of verifying performance.

Will React v15 make this easier/harder?

jaredly:
don't think v15 will change this -- but as I said above, I don't believe it would be too much of a change to support this

wmertens:
Well, it goes pretty deep but indeed it looks like it might be doable. What is the best way to get this in v15?

jaredly:
[...]

PR #340, opened 2016-03-03
Add "trace react updates" to readme

stipsan:
@jaredly something like this?

Related to #337

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,9 @@ In order to fully understand what counts as an "update", you need to understand 
 ![](https://facebook.github.io/react/img/docs/should-component-update.png)
 
 Here "Trace React Updates" will draw a border around every node but C4 and C5.
-Why is other stuff than just renders traced?
-If not careful, the checks done in shouldComponentUpdate can sometimes take longer than the render operation itself.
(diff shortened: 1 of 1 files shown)


jaredly:
:) cool. For the second paragraph I'd probably instead say 

> Why does it trace components that don't actually update? (via shouldComponentUpdate() -> false) 
> This is a limitation of the system used to track updates, and will hopefully change in the future. It doesn't, however, trace the children of components that opt out, as there's no possibility of them updating.

MatthewMiele:
@stipsan I found your commit but trying to find some more information about this feature. Specifically the colours of the box's that show who's updating. Do you know what the colour of the box represents? If so that should be added as well.

stipsan:
OK @jaredly, I'll update it.

@MatthewMiele the color range is from "cold" blue to "hot" red, corresponding to the frequency of update calls.
Cold colours means the updates are infrequent. If you see a lot of red or yellow it may indicate performance optimization is needed :)
I'll add it to the FAQ, no problem!

stipsan:
@jaredly @MatthewMiele updated :)

gaearon:
Thanks!

PR #342, opened 2016-03-08
Fixes issue with TabbedPane not scrolling

kalley:
Can be seen when the Relay tab is present.

Especially apparent when there are a great number of properties.

When viewing components, would push the right pane outside the viewport and it would be unreachable by scrolling.

Before:
image

After:
image

### DIFF
--- a/frontend/TabbedPane.js
+++ b/frontend/TabbedPane.js
@@ -58,6 +58,7 @@ var styles = {
     display: 'flex',
     flexDirection: 'column',
     width: '100%',
+    maxWidth: '100%',
   },
   tabs: {
     display: 'flex',


AndrewIngram:
Can confirm that this has left me unable to use the devtools

jaredly:
thanks for the PR! looking into it now

jaredly:
Hmmm this actually looks like it's already been [fixed on master](https://github.com/facebook/react-devtools/commit/bd3d1e8632034ea1b73de7094a53f5f82ffa9d0b), but it hasn't been released yet to the chrome web store version.
@kassens @spicyj what's the procedure for releasing a patch version?

jaredly:
@kalley @AndrewIngram by which I mean, if I'm running the master version of devtools, and go to for example [this todomvc site](http://www.fashionablenonsense.com/relay-todomvc/#/), I don't see the bug

spicyj:
Just published 0.14.9 to the Chrome store, will get Firefox too.

kalley:
Thanks for letting me know! Works great!

PR #343, opened 2016-03-11
Differentiate between empty arrays and an empty objects in `DataView`

jonathanp:
When inspecting an empty array in the "props and state inspection panel" it currently says `Empty object`. Even though this is technically correct since all arrays in JS are objects this commit changes the label to `Empty array`.


jaredly:
:+1: super

Issue #315, opened 2016-01-17
Feature Request - ability to copy component tree to clipboard


robtweed:
Could you add a mechanism for copying the component tree to the clipboard - either as the expanded tags or (better) a JSON object?  I can't currently find a way to copy the tags from within the Dev Tools window (perhaps I've missed a trick?)

spicyj:
Curious: What would you use this for?

robtweed:
Understandable question, spicyj.  The answer is that I've been looking for a way to graphically visualise a React application's component hierarchy, largely for documentation purposes, but also to allow proper understanding within a team of how the application is constructed and, for example, where a React application's re-rendering events need to be controlled within the component tree.  I've not yet found any reliable tool for creating such a visual representation, but React Developer Tools is one of the few that actually easily and reliably provides the information needed to produce such a visualisation.  The problem is exporting that information from React Dev Tools :-)

iamdustan:
@robtweed would something like https://github.com/algolia/react-element-to-jsx-string#usage solve that for you?

robtweed:
iamdustan - Not sure I see how it would help, but perhaps I'm missing something (which I'm good at! :-)

iamdustan:
If the goal is to `to graphically visualise a React application's component hierarchy`, you can use `react-element-to-jsx` string to get the resulting DOM hierarchy back.

```
console.log(reactElementToJSXString(<My App />);

// would output the following string =>
<div className="root">
  <header className="header">...</header>
  <main className="main">
    ...
  </main>
</div>
```

robtweed:
All I got from using it in my top-level component was
<MainPage
  componentPath={['Top']}
  control={{emit: function noRefCheck() {}, events: {}, handlerFunction: {}, keepAlive: function noRefCheck() {}, log: true, off: function noRefCheck() {}, on: function noRefCheck() {}, sendMessage: function noRefCheck() {}, submitForm: function noRefCheck() {}, updateComponentPath: function noRefCheck() {}}}
/>

No doubt this is due to how and where I added the reactElementToJSXString() command.  My React app is split into individual files - one per component - and browserify is used to compile them together.

Meanwhile React Dev Tools displays the component tree correctly without the need for any invasive code in the application components, but of course it doesn't let me export that info so I can visualise it independently.

gaearon:
I don't think there's a reliably way to do that so I doubt we'll implement it.

PR #344, opened 2016-03-14
[RFC] Copy component to clipboard

aadsm:
This is an action you perform on a node by right clicking on it and selecting "Copy X Usage to Clipboard".

There are two versions of this operation and I would like to use this PR to discuss both and decide which one or even if any is useful.

## Copy Props Children
This is the original one I created and the motivation for me to implement this feature.
When stringifying the component tree I use the children defined in the component props. This matches the children that you see when using a component in code:

```
<Grid>
  <GridItem ... />
</Grid>
```

When stringifying this node in the react-devtools the end result will be exactly what you see in code.
This is useful for me because I usually need to develop or test a component in isolation and I use a fiddle tool for that.
The advantage for me is that by copying in the react-devtools I get the exact component props that are being used in the context I'm seeing in the browser.
I just copy and paste in the fiddle and get a replica of what I see in the running app.

Caveats:
* It works great for components that depend only on props but not so much for store/context dependent ones.
Additionally, random objects, DOM elements and functions are not stringified because there is no known way to create then. I still try to generate function/object friendly names based on the data available.

* The fiddle is created by referring the components by name but they need to be available somehow.. I'm fortunate enough to use a fiddle tool that automatically infers the requires, so I haven't thought much about it.

props-components

## Copy Rendered Children
This is what has been asked in the past (facebook/react-devtools#315) and it just stringifies what you see in the react-devtools panel.
This is the return of the render() method of the component selected.
I personally do not have a use case for this.

rendered-components

# Discuss
If you want to try it you can toggle the features using the `USE_RENDERED_CHILDREN` constant in `ReactStringifier.js` (which should actually be a constructor argument).

Any interest in having both of them, one of them?

@zpao: I made copyToClipboard available as a property so that each native part can pass its own implementation, however, it seems that only firefox can benefit from it. I still use the DOM version as a fallback.

### DIFF
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -20,6 +20,7 @@ var assign = require('object-assign');
 var Bridge = require('../agent/Bridge');
 var NativeStyler = require('../plugins/ReactNativeStyle/ReactNativeStyle.js');
 var RelayPlugin = require('../plugins/Relay/RelayPlugin');
+var ReactStringifier = require('../plugins/ReactStringifier/ReactStringifier.js')
 
 var consts = require('../agent/consts');
 
@@ -143,6 +144,17 @@ class Panel extends React.Component {
     }, 100);
   }
 
+  copyNodeUsageToClipboard(node: Object) {
+    new ReactStringifier(this._bridge).stringify(node).then(stringification => {
+      var fiddle =
+        'ReactDOM.render(\n' +
+        stringification + ',\n' +
+        'document.body\n' +
+        ');';
+      ReactStringifier.copyToClipboard(fiddle);
+    }).catch(e => console.error(e));
+  }
+
   teardown() {
     this.plugins.forEach(p => p.teardown());
     this.plugins = [];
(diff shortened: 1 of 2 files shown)


jaredly:
[pretty-format](https://www.npmjs.com/package/pretty-format#reacttestcomponent-plugin) is the library used by Jest in the Snapshot testing -- would it work for your stringify needs?

gaearon:
Hey, thanks for the PR, and sorry for the long turnaround on this. I think these are cool as a proof of concept, but the limitations are severe enough that it’s hard for me to imagine this being really useful when developing apps for the majority of our users. I will close this out because it doesn’t seem like a commonly requested feature.

aadsm:
@gaearon What do you recommend to integrate this into react-devtools without me having to maintain a fork of the project?
I was thinking about a way to provide plugins (in form of other chrome extensions) to react-devtools using this: https://developer.chrome.com/extensions/messaging#external. It might be tricky from a security point of view though but manageable I think. Any thoughts on this?

gaearon:
I don't really have much context on this, @jaredly maybe can answer. Since nobody actively owns DevTools development and it is done sporadically I would be wary of adding extensions mechanisms and increasing complexity and surface area.

PR #345, opened 2016-03-14
Fix #248, when an <option> contains raw strings

### LINKED ISSUE #248: Multiple <option> children causes TypeError: Cannot read property 'props' of undefined
When I open react-devtools, I see this:

screenshot 2015-10-07 12 14 45

After clicking the button, it fails with this error:

screenshot 2015-10-07 12 16 47

I tried closing the tab and opening everything again, but the problem persists.

My page uses React `0.14-rc1`. I tried react-devtools on another page that uses react `0.13` and it works fine (in spite of the warning about version).

jaredly:
[awkward that this was a bug for so long...]

The bug is easily replicable (see https://github.com/facebook/react-devtools/issues/248), and the problem is
that `element` is just a string in that case, such that there is no
`element._currentElement`.

Test Plan:
Go to a page with `<option>some text {1}</option>`, see that devtools doesn't
break.

### DIFF
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -30,7 +30,7 @@ function getData(element: Object): DataType {
   // If the parent is a native node without rendered children, but with
   // multiple string children, then the `element` that gets passed in here is
   // a plain value -- a string or number.
-  if (typeof element !== "object") {
+  if (typeof element !== 'object') {
     nodeType = 'Text';
     text = element + '';
   } else if (element._currentElement === null || element._currentElement === false) {
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -27,7 +27,11 @@ function getData(element: Object): DataType {
   var text = null;
   var publicInstance = null;
   var nodeType = 'Native';
-  if (element._currentElement === null || element._currentElement === false) {
+  // Sometimes element is just a string or number
+  if (typeof element !== "object") {
+    nodeType = 'Text';
+    text = element + '';
+  } else if (element._currentElement === null || element._currentElement === false) {
     nodeType = 'Empty';
   } else if (element._renderedComponent) {
     nodeType = 'NativeWrapper';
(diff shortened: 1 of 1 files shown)


spicyj:
What do the devtools show after this?

jaredly:
Good call. fixed so that devtools actually shows something useful. If a raw value is entered, it is coerced to a string (as the web does)
image

backend/getData.js
```diff
@@ -27,7 +27,11 @@ function getData(element: Object): DataType {
   var text = null;
   var publicInstance = null;
   var nodeType = 'Native';
-  if (element._currentElement === null || element._currentElement === false) {
+  // Sometimes element is just a string or number
+  if (typeof element !== "object") {
```
spicyj: I thought `element._currentElement` was a string/number. Am I misunderstanding here?

jaredly: It is... most of the time. Just not when you've got a string element of an `<option/>`, which is what has been causing #248. In that case, and only in that case, `element._renderedChildren` is a list of raw strings/numbers, b/c it's not allowed to put a `<span>` inside of an `<option>`. In every other case, if there are multiple children, then `element._renderedChildren` is a list of more elements.

spicyj: `_renderedChildren` should be null: http://react.jsbin.com/dikerovequ/edit?js,output.

jaredly: oh you're right I misspoke. If `_renderedChildren` is null, then we iterate through the `props.children`, which contains plain values. I've updated the comment to be more accurate

jaredly:
@spicyj bump?

AaronHarris:
This is also a blocking issue for me. Is there any reason @spicyj this cannot be merged?

nicolanicola:
can this be merged please?  Its blocking me

spicyj:
All the duck typing here really freaks me out but I guess this is okay.

PR #356, opened 2016-03-31
Make setting props work again

### LINKED ISSUE #314: Feature request: ability to change props
It would be really useful if we could modify props!

spicyj:
Fixes #314. Works on React master but I haven't had the chance to test older versions.


kassens:
Sounds good. I think it was broken a long time now and better to work with the current version then old ones.

jaredly:
awesome thanks for getting into this! I was stumped

msankhala:
On chrome web store it is still having 2.5.2 version which do not allow to edit props.

PR #359, opened 2016-04-06
Fix rendering 0

bruderstein:
When a component renders a numerical `0`, the `_currentElement` property
is set to numerical `0`. The check for a "raw" element was inside a
truthy check for `_currentElement`, which in the case of `0` is
obviously false. In the devtools, a rendered `{0}` shows up as `</>`.


spicyj:
Thanks!

PR #360, opened 2016-04-07
Handle undefined component type

josestbernard:


jaredly:
@josestbernard under what circumstance does this occur?

gaearon:
Thanks for the PR! Seems like the better way to fix it would be this: https://github.com/facebook/react-devtools/pull/373#issuecomment-220127992.

PR #361, opened 2016-04-08
Check nodeType before attempting to highlight.

STRML:
Since React v15 uses comment nodes to split text (rather than `<span>`s),
this fixes a crashing bug.


STRML:
Updated to fix Flow typecheck error.

spicyj:
Seems reasonable enough, thanks.

Arnavion:
The same fix is needed for [`Agent.scrollToNode`](https://github.com/facebook/react-devtools/blob/2beafbf/agent/Agent.js#L205) since `node.scrollIntoView` is undefined for comment nodes.

PR #363, opened 2016-04-14
flow 0.23

iamdustan:
Seeing @kassens PRs to this repo was always my notification that a new flow release was available. When first upgrading from flow@0.21 -> flow@0.23 there were ~165 errors (exact number I forget, but it was in that range).

Most of it was that flow now requires the `State` shape to be declared on components that use it and other static properties like `node => this.node = node`. After getting through all those rather mundane changes there were a few other changes that required a bit more work.

Specifically the following warning I couldn’t reconcile with only flow definition changes, so I explicitly converted it to a `string` in the `BlurInput.setState` call.

```sh
plugins/ReactNativeStyle/StyleEdit.js:63
 63:           <BlurInput
               ^ React element `BlurInput`
 65:             onChange={newAttr => this.setState({newAttr})}
                                                     ^^^^^^^ number. This type is incompatible with
 26:   newAttr: string,
                ^^^^^^ string
```

Additionally, the `Panel` was being used by the different shells with a `{...config}` object that didn’t have many of the required props. This was solved by marking more fields as optional and putting in the appropriate checks for it.

Lastly `ReactElement` -> `React$Element` in the flowdef. https://github.com/facebook/flow/blob/master/lib/react.js#L106-L116 Simply renaming it seemed to suffice, but there may be details I’m missing.

### DIFF
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -19,18 +19,23 @@ var assign = require('object-assign');
 var consts = require('../../agent/consts');
 var previewComplex = require('./previewComplex');
 
-class DataView extends React.Component {
-  props: {
-    data: Object,
-    path: Array<string>,
-    inspect: (path: Array<string>, cb: () => void) => void,
-    showMenu: (e: DOMEvent, val: any, path: Array<string>, name: string) => void,
-    startOpen: boolean,
-    noSort?: boolean,
-    readOnly: boolean,
-  };
+type Inspect = (path: Array<string>, cb: () => void) => void;
+type ShowMenu = boolean | (e: DOMEvent, val: any, path: Array<string>, name: string) => void;
 
-  render(): ReactElement {
+type DataViewProps = {
+  data: Object,
+  path: Array<string>,
+  inspect: Inspect,
+  showMenu: ShowMenu,
+  startOpen?: boolean,
+  noSort?: boolean,
+  readOnly?: boolean,
+};
+
+class DataView extends React.Component {
+  props: DataViewProps;
+
+  render() {
     var data = this.props.data;
     if (!data) {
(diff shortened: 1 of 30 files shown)


frontend/DataView/DataView.js
```diff
@@ -19,18 +19,20 @@ var assign = require('object-assign');
 var consts = require('../../agent/consts');
 var previewComplex = require('./previewComplex');
 
+type DataViewProps = {
+  data: Object,
+  path: Array<string>,
+  inspect: (path: Array<string>, cb: () => void) => void,
+  showMenu: boolean | (e: DOMEvent, val: any, path: Array<string>, name: string) => void,
```
jaredly: given that showMenu is used twice, maybe make a separate declaration for it?

iamdustan: done

jaredly:
hmm can you fix travis to install the right version of flow?

iamdustan:
any idea why it’s not? that’s kinda throwing me. I updated it in the package.json and I don’t see any other reason for 0.21 to be coming in instead of 0.23...

iamdustan:
ah. shrinkwrap. Updating now.

iamdustan:
wow shrinkwrap is a PITA. flow is now passing on travis, but something changed in eslint while trying to regenerate the shrinkwrap and now that fails. :(

iamdustan:
:boom: shrinkwrapped properly. CI is back to green!

kassens:
Sorry for the delay here, we should probably just get rid of the shrinkwrap.

iamdustan:
I suppose I can start the update to flow 0.25 now, too :)

kassens:
@iamdustan btw, I created #374 to delete the shrinkwrap file. Should make this easier :)

PR #364, opened 2016-04-20
Setup Relay backend in Electron shell

matthewwithanm:
This avoids an "unknown call" warning (yellow box).

I mentioned this [here][1] but if this isn't the right way to fix it, lmk!

cc @frantic

[1]: https://github.com/facebook/react-devtools/pull/263#issuecomment-182160468


iamdustan:
Curious if this is basically a noop if the target env isn’t using Relay?

frantic:
cc @kassens

jaredly:
@iamdustan yup, it doesn't do anything if Relay isn't present

kassens:
lgtm

PR #374, opened 2016-05-19
Remove npm-shrinkwrap

kassens:
The shrinkwrap file is a pain to work with:

- it makes it confusing to update modules since npm-shrinkwrap.json silently overrides package.json
- npm-shrinkwrap.json diffs are huge and not really human readable, so there's no benefit in "seeing what changed"

It seems close enough to just specify strict version ranges instead. The
likelihood that something breaks due to a recursive dependency update seems
small compared to the pain the shrinkwrap causes.


spicyj:
Okay sure.

iamdustan:
Of course that would cause CI to fail 👹

gaearon:
I opened #377 as a successor to this that fixes Travis.

kassens:
Thanks @gaearon!

PR #376, opened 2016-05-19
Modify for-in loop to use Object.getOwnPropertyNames, surfacing enume…

tquetano-r7:
…rable properties and avoiding prototype properties

### DIFF
--- a/agent/dehydrate.js
+++ b/agent/dehydrate.js
@@ -83,9 +83,15 @@ function dehydrate(data: Object, cleaned: Array<Array<string>>, path?: Array<str
     };
   }
   var res = {};
-  for (var name in data) {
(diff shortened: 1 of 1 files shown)


tquetano-r7:
This PR is a clean redo of PR [293](https://github.com/facebook/react-devtools/pull/293), which had a ton of changes and was months old so it caused regression issues.

omeid:
Any updates on this? I would be happy to help.

gaearon:
Can you test this and post results of how you tested, and what the output looks like before and after? It’s a bit hard to jump on this without the context.

gaearon:
I re-read the thread. I think this change looks good and is similar to what e54919495d3cbfa6cc45573cb9caad1eab4db663 is doing.
LGTM

PR #377, opened 2016-05-19
Update ESLint

gaearon:
This takes #374 that currently fails linting due to some transitive dependency issue, and updates ESLint to 2.x so that it lints and does it cleanly. I had to change the renamed rules and remove the `strict` rule because I couldn’t get to disable ESLint `module` mode without also disabling support for `import type`. Finally, `consistent-return` has gotten more aggressive so I fixed the places where the lint would fail.

Reviewers: @spicyj


backend/attachRenderer.js
```diff
@@ -50,7 +50,9 @@ function attachRenderer(hook: Hook, rid: string, renderer: ReactRenderer): Helpe
     extras.getNativeFromReactElement = function(component) {
       try {
         return renderer.Mount.getNode(component._rootNodeID);
-      } catch (e) {}
+      } catch (e) {
+        return undefined;
+      }
```
iamdustan: Any reason simply `return;` isn't used here?

gaearon: Yea, `consistent-return` ESLint rule doesn’t like this. The idea is you either `return;` everywhere in a function or `return <something>;` in it.

spicyj:
Seems reasonable enough, thanks.

PR #380, opened 2016-05-26
Visual style tweaks

spicyj:
Before:

After:

The click targets are a little better now too.


iamdustan:
Looks legit in Firefox, too.

in Firefox

iamdustan:
lgtm :)

jaredly:
looks great!

PR #392, opened 2016-06-27
Fix number type input for React Native style editor

jhen0409:
Currently it have `RCTConvert` error on React Native:


jaredly:
cool, can't hurt. Would be good to detect what attributes are supposed to be numbers, but this is good for now

PR #393, opened 2016-06-27
Fix el.setNativeProps in React Native setupBackend

jhen0409:
The prop should be `{ style: newStyle }`.

This is screenshot for set style to native component:


spicyj:
Thanks!

PR #400, opened 2016-07-12
Fix NativeWrapper check in nodeMatchesText

jhen0409:
The DevTools will get undefined for `store.getParent(key)` when reproduce following steps:

1. Keep search text `Use redux` on DevTools
2. Add new todo `Use redux` on app
3. It will broken when I changed search text on DevTools


jaredly:
cool!

PR #406, opened 2016-08-05
Handle & render the _source property

jaredly:
To take advantage of this, use the [transform-react-jsx-source](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx-source) babel plugin.

The `_source` property was added to react [last year](facebook/react@009b7c8) and can be used by development tools to track the source file & line of react elements. React Native is also [taking advantage of it](facebook/react-native#6351).

Test Plan:
- install the [transform-react-jsx-source](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx-source) babel plugin
- open the react devtools to an element that was created in your app (can't
  have been created by a library you're using, because those are probably
  pre-transpiled & ignored by babel)

image
^ `src/run.js:12`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ Tools or use a pre-prelease version.
 
 ## Usage
 
+### Supporting tools
+
+- The babel plugin [trasnform-react-jsx-source](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx-source) is required if you want react devtools to tell you the source file & line number of created react elements. It's display in the bottom of the right panel if the information is present.
+
 ### Tree View
 
 - Arrow keys or hjkl for navigation
--- a/frontend/PropState.js
+++ b/frontend/PropState.js
@@ -20,17 +20,6 @@ var React = require('react');
 var decorate = require('./decorate');
 var invariant = require('./invariant');
 
-/**
- * Shorten a file path to something reasonable.
- * - if it has `/src/` in it, assume that's the project root
- * - if it starts w/ `/Users/name/`, replace that with `~/`
- */
-var shortFilePath = text => {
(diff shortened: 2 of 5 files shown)


jaredly:
ping @spicyj or @frantic?

frontend/PropState.js
```diff
@@ -20,6 +20,18 @@ var React = require('react');
 var decorate = require('./decorate');
 var invariant = require('./invariant');
 
+/**
+ * Shorten a file path to something reasonable.
+ * - if it has `/src/` in it, assume that's the project root
+ * - if it starts w/ `/Users/name/`, replace that with `~/`
+ */
+var shortFilePath = text => {
```
jaredly: I like having an absolute path -- and it's critical for React Native's "open the file in your editor" functionality

jaredly: Of course, we could change it to include both an absolute path and a project-relative path? or sth. or the # that is how long the project prefix is, so we could do `source.filePath.slice(source.prefixLength)` or sth

spicyj: I think the project-relative path is always what you want. The packager would know where the root is.

spicyj:
lgtm

README.md
```diff
@@ -25,6 +25,10 @@ Tools or use a pre-prelease version.
 
 ## Usage
 
+### Supporting tools
+
+- The babel plugin [trasnform-react-jsx-source](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-react-jsx-source) is required if you want react devtools to tell you the source file & line number of created react elements. It's display in the bottom of the right panel if the information is present.
```
chibicode: It's display → It's displayed

PR #409, opened 2016-08-18
Avoid unnecessary measurements

gaearon:
If "Trace React Updates" is disabled, don't spend time measuring nodes.
This gives significant perf improvements in DEV on pages with many nodes.

Before:

screen shot 2016-08-18 at 21 19 21

After:

screen shot 2016-08-18 at 21 29 22

When I toggle "Trace React Updates" the updates show up so it seems like feature still works.


spicyj:
lgtm

PR #410, opened 2016-08-18
Compile React for production

gaearon:
Seems like we weren't doing it before. Also need to update internal release instructions because they currently don’t include `webpack` step for FF.


spicyj:
lgtm

PR #411, opened 2016-08-19
Use React 15

gaearon:
I couldn’t find any issues so far.
I tried:

* Facebook.com
* Power Editor
* React website


jaredly:
yeah, it doesn't look like we're using any deprecated APIs

PR #412, opened 2016-08-19
Flush search pane updates synchronously

### LINKED ISSUE #220: Typing quickly into filter box drops characters
Reported by @MichelleTodd.

gaearon:
Ideally we would debounce this but it seems less important.
Verified it works on Power Editor.

Fixes #220, related to #311.

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -163,21 +163,26 @@ class Store extends EventEmitter {
   }
 
   emit(event: string): boolean {
+    if (this._eventQueue.indexOf(event) !== -1) {
+      // to appease flow
+      return true;
+    }
+    this._eventQueue.push(event);
     if (!this._eventTimer) {
-      this._eventTimer = setTimeout(() => {
-        this._eventQueue.forEach(evt => {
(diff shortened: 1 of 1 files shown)


frontend/Store.js
```diff
@@ -163,23 +163,28 @@ class Store extends EventEmitter {
   }
 
   emit(event: string): boolean {
+    if (this._eventQueue.indexOf(event) !== -1) {
+      // to appease flow
+      return true;
+    }
+    this._eventQueue.push(event);
     if (!this._eventTimer) {
-      this._eventTimer = setTimeout(() => {
```
gaearon: I just extracted this into `flush`.

gaearon:
I don’t think the description in https://github.com/facebook/react-devtools/issues/220#issuecomment-165283894 is accurate (maybe it’s outdated?) It didn’t seem to me that there’s anything related to `postMessage` going on. This seems like a classic problem of controlled input that relies on previous value from the store when the store is updated asynchronously. Solved by updating it synchronously.

spicyj:
lgtm

PR #413, opened 2016-08-19
Fix scroll to text node

gaearon:
This fixes “Scroll to Node” for text nodes.
Test plan: `<div>{'a'}{'b'}</div>`, right-click either and click "Scroll to Node".
Before it used to raise an error, as noted in https://github.com/facebook/react-devtools/pull/361#issuecomment-212703004.

### DIFF
--- a/agent/Agent.js
+++ b/agent/Agent.js
@@ -206,10 +206,18 @@ class Agent extends EventEmitter {
       console.warn('unable to get the node for scrolling');
       return;
     }
-    if (node.scrollIntoViewIfNeeded) {
-      node.scrollIntoViewIfNeeded();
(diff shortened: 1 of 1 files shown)


agent/Agent.js
```diff
@@ -206,10 +206,18 @@ class Agent extends EventEmitter {
       console.warn('unable to get the node for scrolling');
       return;
     }
-    if (node.scrollIntoViewIfNeeded) {
-      node.scrollIntoViewIfNeeded();
+    var element = node.nodeType === Node.ELEMENT_NODE ?
+      node :
+      node.parentElement;
+    if (!element) {
```
jaredly: could you add `|| !element.scrollIntoView`?

Issue #198, opened 2015-09-09
Uncaught TypeError: Invalid value used as weak map key @backend.js 525


terebentina:
This happens on chrome latest with devtools 0.14.1 when I interact with my app. 
I did a console.log(element) on line 524 in backend.js and when this error appears, one of the elements there is null.

```
Uncaught TypeError: Invalid value used as weak map keygetId @ backend.js:525
(anonymous function) @ backend.js:569
onUpdated @ backend.js:568
(anonymous function) @ backend.js:1015
(anonymous function) @ VM55:42
emit @ VM55:41
performUpdateIfNecessary @ backend.js:2290
obj.(anonymous function) @ backend.js:2363
runBatchedUpdates @ ReactUpdates.js?ce09:151
Mixin.perform @ Transaction.js?6dff:134
Mixin.perform @ Transaction.js?6dff:134
assign.perform @ ReactUpdates.js?ce09:95
flushBatchedUpdates @ ReactUpdates.js?ce09:175
wrapper @ ReactPerf.js?ef93:70
Mixin.closeAll @ Transaction.js?6dff:207
Mixin.perform @ Transaction.js?6dff:148
ReactDefaultBatchingStrategy.batchedUpdates @ ReactDefaultBatchingStrategy.js?ef70:66
batchedUpdates @ ReactUpdates.js?ce09:110
ReactEventListener.dispatchEvent @ ReactEventListener.js?2365:174
```

jaredly:
What's your React version?

terebentina:
0.13.3

terebentina:
I just upgraded to react 0.14.0 and this still happens.

jaredly:
Strange. So [this line](https://github.com/facebook/react-devtools/blob/e8b760d723f0de30163159a0fe5deb2626d39801/agent/Agent.js#L318) is where that error would be happening, and it's caused by `element` not being an object -- it's probably `undefined` or `null` or something, and that's causing the error. I can't imagine *why* it would be that way though.
Are you able to reliably reproduce this? Any chance you could spelunk in an tell us what the value of `element` is there, or come up with a repro example?

terebentina:
Yes, reliably, every time in my app. See above in the issue description: `I did a console.log(element) on line 524 in backend.js and when this error appears, one of the elements there is null.`.
I don't know if it's related or not but I do set some elements to null so that they're not displayed anymore in certain cases. 
I will try to come up with an example later today.

markoshust:
[...]

PR #414, opened 2016-08-19
Fix "Invalid weak map key" error #198

jaredly:
This is a longstanding bug that I finally tracked down.

When a child is `null`, its `typeof` === `"object"` :facepalm:
And `null` is an invalid WeakMap key.

See repro from [this comment](https://github.com/facebook/react-devtools/issues/198#issuecomment-237571581)

### DIFF
--- a/agent/Agent.js
+++ b/agent/Agent.js
@@ -325,7 +325,7 @@ class Agent extends EventEmitter {
   }
 
   getId(element: OpaqueNodeHandle): ElementID {
-    if (typeof element !== 'object') {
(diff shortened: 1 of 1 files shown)


gaearon:
I thought you couldn’t figure out how it’s possible *in React* and didn’t fix because of that 😄 .

FWIW I don’t know a situation where `null` ends up as a child node because `ChildReconciler` should skip it.. I think. And I couldn’t repro with that example. 😢

jaredly:
I've only encountered it in the presence of hot reloading...

PR #415, opened 2016-08-19
Slice flush into small parts

gaearon:
This is a proof of concept of slicing flush into small parts.
React DevTools currently heavily slow down big apps like (PE) with all the synchronous work.

Example before:

screen shot 2016-08-19 at 18 30 18

Example after:

screen shot 2016-08-19 at 18 28 49

### DIFF
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -15,10 +15,29 @@ var hydrate = require('./hydrate');
 var dehydrate = require('./dehydrate');
 var performanceNow = require('fbjs/lib/performanceNow');
 
+// Based on https://gist.github.com/paullewis/55efe5d6f05434a96c36
+var cancelIdleCallback = window.cancelIdleCallback || clearTimeout;
+var requestIdleCallback = window.requestIdleCallback || function(cb) {
+  var start = performanceNow();
+  return setTimeout(() => {
+    cb({
+      didTimeout: false,
+      timeRemaining() {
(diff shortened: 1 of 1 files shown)


jaredly:
this is awesome. Does it get super behind though?

gaearon:
In my testing it gets behind by a few seconds if you're doing heavy operations (e.g. unmount/mount a FixedDataTable multiple times without giving it time to breathe).

For React Motion-like demos with tons of fast updates, it gets behind by about half a second so it’s “trailing” updates at a slower speed.

It’s a bit hard for me to say if this is confusing or not unless we let people use this, and wait for complaints.

agent/Bridge.js
```diff
@@ -197,34 +214,45 @@ class Bridge {
     this._inspectables.set(id, {...prev, ...data});
   }
 
-  sendOne(evt: string, data: any) {
```
gaearon: This one seems dead, I couldn’t find calls to it anywhere.

jaredly:
I'd love for there to be an indicator that we're lagging, to reduce confusion. maybe add a flag to "send-many" indicating how many evens we still have to send. Then in the UI we could show a loading indicator if that number gets above 100 or something.

That doesn't have to be part of this diff though

gaearon:
I did some tweaks to better handle real world scenarios, now it’s consistent pretty fast even with large number of events so I think maybe we don’t need an indicator.

PR #416, opened 2016-08-21
UI improvements, behaviour closer to native dev tools

pastelsky:
I've tried to introduce a few minor UX / UI improvements  to issues that I found in devtools over the weekend. These include : 

1) Rearranged controls and tweaked UI for a cleaner look.
image

2) Breadcrumbs are now highlighted on hover.
image

2) Replaced `cursor:pointer` with `cursor:default`. Pointers are usually a thing of webpages.

3) Syntax highlighting and colors are tweaked a bit to chrome's design. 

4) Tooltips are now closer to how chrome does them — more readable.
image

5) Code wrapping works in tree-view - better usability when devtools is docked towards the right.
image

6) Hitting `enter` after a editing a field in DataView now closes the field in addition to applying changes (standard behaviour).

If these changes seem to go a lot in one PR, I could break them down into multiple.

PR #420, opened 2016-08-26
Added hover state to breadcrumbs

pastelsky:
As discussed in #416, I'm breaking down my PR into multiple ones to make discussions easier.


gaearon:
Thanks for the PR! To be honest I don’t find this to be an improvement. I think component names look odd in a sans serif font, and lack of pointer makes it non-obvious tags are clickable.

What problem are you solving? Is there a different way to address it?

gaearon:
Never mind, I used it for five minutes and I do agree it's better. 😛 
Thanks!

PR #423, opened 2016-08-30
(WIP) Try to make FF faster when paused

gaearon:
I’m dropping this because I couldn’t finish this and I don’t even know if it would result in an improvement.

The idea was to do pass `pause` and `resume` messages to the bridge (like we do for Chrome) so that it could do work less intensely when DevTools are paused. This worked great in Chrome but I don’t know if we can really reap the same benefits in FF.

Debugging FF extensions is a nightmare because:
1. Debugger often hangs
2. Breakpoints don’t consistently break
3. It’s super slow

If you want to pick this up, you’ll want to:
1. Make sure we actually track visibility change correctly—I’m not sure it works in all cases, especially since we don’t get an event on first open with this approach
2. Make sure we actually pass event to the bridge. I don’t understand the port system and I have no idea how to correctly send the message to it from the panel.
3. Make sure we use `this._isPaused` in the bridge wisely. The current handling works great for Chrome but I’m not sure if same heuristics work for FF, especially considering we use that hacky `requestIdleCallback` “polyfill”.

### DIFF
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -22,9 +22,11 @@ var cancelIdleCallback = window.cancelIdleCallback || clearTimeout;
 var requestIdleCallback = window.requestIdleCallback || function(cb, options) {
   // Magic numbers determined by tweaking in Firefox.
   // There is no special meaning to them.
-  var delayMS = 3000 * lastRunTimeMS;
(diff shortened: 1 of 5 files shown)


gaearon:
Meh.

Issue #424, opened 2016-09-01
react devtool doesn't support react chrome extension


Nomia:
Hi,

I've built a chrome extension based on react. Recently , it's fairly slow when render list. I wanna have deep look at it. But, the react tab won't show up in the chrome devtool. My react version is 0.14.8. I've exposed both React and ReactDOM as global variables. 

The react tab will show when transfer my chrome extension app to a server.
Hope that you could help me

Thank you!

ngyikp:
Duplicate of https://github.com/facebook/react-devtools/issues/91

Nomia:
Sorry,

After reading that thread(#91) 3 times. I still don't understand how to fix this issue.

What should I do?
Should I hack into the react-devtool extension and inject some code, then upload a new extension?

gaearon:
Try the [standalone version](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools) of DevTools.

It should work within a Chrome extension: https://github.com/facebook/react-devtools/issues/434#issuecomment-280116615

Issue #350, opened 2016-03-19
Usage with a web app and iOS Simulator


elado:
Is there a way to use react-devtools with a web page on iOS Simulator Mobile Safari (currently running from localhost)? Maybe some remote debugger that would work in Chrome with the extension?

elado:
#440

Issue #320, opened 2016-01-27
Support iframes that doesn't have the top===0


hackhat:
I've found this solution ( https://github.com/facebook/react-devtools/issues/76 ) to debug iframes. Let's say my iframe has `postion: absolute, top: 10px` then when I hover a element in the react-devtools the debug boxes that appear on the webpage are offset by 10px up. This happens because of the top: 10px;

![image](https://cloud.githubusercontent.com/assets/11173732/12611910/2e163bc4-c4e7-11e5-8090-521934a897ce.png)

hackhat:
The fix is to use margin instead of top. Solution: `top: 0; margin-top: 10px`

jaredly:
yeah. I'm not sure there's a super good way to do that. I looked into it a bit

gaearon:
You can now use the [standalone DevTools](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools) to debug iframes.

zlbben:
hi，standalone DevTools had the same problem

Issue #163, opened 2015-08-26
No react tools tab in chrome mobile developer tools


lfrodrigues:
We're developing an app using cordova and react. 

We use mobile developer tools for debugging but there we don't get the react tab.

cmacnaughton-al:
Same issue, react tab is gone.
Chrome Version 44.0.2403.157 (64-bit)

spicyj:
This is a problem with the chrome web store version of the extension (not just the beta), correct?

lfrodrigues:
yes it is

spicyj:
This is somewhat related to #91, I think, in that the devtools hook doesn't get added to the page. Not sure how easy it is to fix.

chauthai:
+1 for cordova support

elad-lachmi:
another +1 for cordova support

TheBoroer:
Another +1 for cordova support! :+1:

gaearon:
Unfortunately +1’s won’t help here because we are not working on Cordova apps. If you are working on one, and would like to see this fixed, please consider fixing this and submitting a PR. You can find instructions to get started here:

https://github.com/facebook/react-devtools#for-hacking

gaearon:
This should be solved with the new standalone version.
https://github.com/facebook/react-devtools/tree/master/packages/react-devtools

PR #440, opened 2016-10-04
Devtools for everyone!!!

### LINKED ISSUE #91: Not showing up when inspecting a Chrome app

jaredly:
Just add `<script src="http://localhost:8097"></script>` to the top of any html file, and magically debug the react running there. (swap out your network-local IP for `localhost` for pages not on your machine)

This works for:
- iframes
- chrome apps
- safari
- mobile safari/chrome/whatever
- embedded webviews
- chrome extensions (need to disable CSP though to eval... could probably fix this at some point)
- like anything

Test Plan:
- `npm run build` in `/shells/electron`
- then `npm start`

Add the script tag to the top of an html file.
Load the html file anywhere.
Debug to your heart's content.

Fixes #91 #424 #76 #320 #350 #163 #98 and probably others

image

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,6 @@
 [ignore]
 .*/react/node_modules/.*
+.*/electron/node_modules/.*
 .*node_modules/babel.*
 .*node_modules/classnames.*
 .*node_modules/json-loader.*
--- a/backend/websocketConnect.js
+++ b/backend/websocketConnect.js
@@ -1,0 +1,94 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+'use strict';
+
+function websocketConnect(uri: string, WebSocket?: (val: string) => Object) {
+  WebSocket = WebSocket || window.WebSocket;
+  var messageListeners = [];
+  var closeListeners = [];
+  var ws = new WebSocket(uri);
+  // this is accessed by the eval'd backend code
+  var FOR_BACKEND = { // eslint-disable-line no-unused-vars
+    wall: {
+      listen(fn) {
+        messageListeners.push(fn);
+      },
+      onClose(fn) {
+        closeListeners.push(fn);
+      },
+      send(data) {
+        ws.send(JSON.stringify(data));
+      },
+    },
+  };
+  ws.onclose = () => {
+    console.warn('devtools socket closed');
+    closeListeners.forEach(fn => fn());
+  };
+  ws.onerror = error => {
+    console.warn('devtools socket errored', error);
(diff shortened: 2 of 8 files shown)


jaredly:
ping @frantic or @spicyj?

frantic:
LGTM!

oelmekki:
> Fixes #91 #424 #76 #320 #350 #163 #98 and probably others

This is epic. 

Thanks a lot Jared, you just made my work on chrome extensions way 
easier :)

matthewwithanm:
woooah

kintendo:
@oelmekki could you explain how were able to get the dev tools to inspect your chrome extension?

oelmekki:
@kintendo I have not, this commit is not yet released :)

It will be in `0.15.5`, or whatever is released after `0.15.4`.

kintendo:
@oelmekki ah! thanks for that. Here I was already trying to get it to work, haha

maks:
@jaredly I'm probably missing something because when I tried this for a chrome (packaged) app I get an error in devtools console:

> Refused to load the script 'http://localhost:8097/' because it violates the following Content Security Policy directive: "default-src 'self' blob: filesystem: chrome-extension-resource:". Note that 'script-src' was not explicitly set, so 'default-src' is used as a fallback.

I used the current master branch as now from this repo and followed instructions in shells/chrome/Readme. The extension loaded ok but I got the above when I included the localhost script in my main html file. So for now it doesn't fix #91 for me.

jaredly:
@maks look up "disable CSP" for your chrome packaged app. then it should work for you

maks:
@jaredly Thansk for the very quick response!
But I think this might be a show stopper for me, as I don't think chrome packaged apps are able to disable CSP.
When I try per docs from (https://developer.chrome.com/extensions/contentSecurityPolicy#relaxing) I get the following next to my Chrome App in chrome://extensions:
There were warnings when trying to install this extension:
'content_security_policy' is only allowed for extensions and legacy packaged apps, but this is a packaged app.

Others have found the same:  http://stackoverflow.com/a/21131034/85472
as "v2" packaged apps are supposed to use "sandboxes" but I think they won't help in this case.

I guess I need to file a bug with Chromium...

jaredly:
hmmmm darn.

maks:
Yes :-(
In case it helps to come up with something in the future this is the [actual docs for CSP for packaged apps](https://developer.chrome.com/apps/contentSecurityPolicy) instead of for extensions which have different restrictions.
As far as I can see though, the things they offer as alternatives on that page won't help here, so I guess for now, no nice devtools for me when using react for chrome packaged app development.

PR #442, opened 2016-10-19
Guard window.addEventListener call

matthewwithanm:
When upgrading Nuclide's build of the devtools and testing in RN, this was causing an error (since `window.addEventListener` isn't defined for RN).

lmk if I should check `capabilities.scroll` instead.


spicyj:
Blames to ba272dff8c9ff5bd27461ad914381b4765a69ffc. Lgtm.

PR #451, opened 2016-11-06
Fix #444: Electron shell exception due to missing 'ip' module

### LINKED ISSUE #444: Exception running in electron shell: no such module 'ip'
When I tried running the devtools using the electron shell, I got an error about the module 'ip' not being found.

[This is](https://github.com/facebook/react-devtools/blob/ec328ee3da8bbcb1afa8fcdd84197cbed33a542e/shells/electron/index.html#L72) the problematic line and it looks like the line was added in d3b52777e143eee737a65995ecbbd32fa75e7d14. Perhaps `ip` needs to be added to `package.json`. /cc @jaredly

rigdern:


gaearon:
Sorry, missed this PR. Fixed it separately in 6dbffabfc1451cc97b3f345ed6c129667d4355c9.
Appreciate the PR though!

PR #452, opened 2016-11-08
Fix invariant(state.toJS) on dev mode

jhen0409:
Just add [`format` arg](https://github.com/facebook/react-devtools/blob/cf0578966baf954038b7a8217165603133ee9af1/frontend/invariant.js#L28-L30) to avoid it crash on dev mode.

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -470,7 +470,7 @@ class Store extends EventEmitter {
   changeBananaSlug(state: ControlState) {
     this.bananaslugState = state;
     this.emit('bananaslugchange');
-    invariant(state.toJS);
+    invariant(state.toJS, 'the state must be immutable record');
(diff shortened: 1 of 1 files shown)


frontend/Store.js
```diff
@@ -470,7 +470,7 @@ class Store extends EventEmitter {
   changeBananaSlug(state: ControlState) {
     this.bananaslugState = state;
```
sohanpurohit: Leave a comment

frontend/Store.js
```diff
@@ -470,7 +470,7 @@ class Store extends EventEmitter {
...
     this.emit('bananaslugchange');
-    invariant(state.toJS);
+    invariant(state.toJS, 'the state must be immutable record');
     this._bridge.send('bananaslugchange', state.toJS());
   }
```
sohanpurohit: @

PR #454, opened 2016-11-13
Incorporate react tags into UI

### LINKED ISSUE #439: [React Native] Include react tags in inspector UI
When seeing a native view in the debugger, I do not know of a good way to map between that native view and the JavaScript view that generated it.

A couple of changes to the devtools could be useful in this scenario:
- When inspecting a view, include its tag in the UI. This would give me a clue as to which native view I'm looking at.
- Allow searching by react tag. That way, if I encounter a native view in the debugger, I can quickly see which JavaScript view it corresponds to.

rigdern:
Fixes #439. For a node that has a react tag, its name is now rendered as `<RCTView(17)>` where "RCTView" is the component type and "17" is the react tag.

React tags are only rendered for native primitives. Changes:
  - React tag is rendered in the right pane when inspecting a node.
  - React tag is rendered next to the component type inside of the TreeView.
  - Search now supports searching by react tag.

### Outstanding Issue: ReactDOM vs React Native

Currently, this feature is always enabled. I want this feature for React Native. Developers who are either debugging React Native internals or writing their own native modules may find this feature to be useful. It helps them to correlate components they are debugging in the native debugger with JS components.

However, I'm not sure if this feature is as useful for ReactDOM. Fewer people will probably be debugging ReactDOM internals than React Native internals + native modules. Also, it may already be easy to correlate nodes in ReactDOM internals with nodes from the app. I wouldn't know, I haven't had to do this before.

If we only want to enable this for React Native, I would appreciate some advice on the best way to do that.

### DIFF
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -30,6 +30,7 @@ function getData(element: Object): DataType {
   var text = null;
   var publicInstance = null;
   var nodeType = 'Native';
+  var reactTag = null;;
   // If the parent is a native node without rendered children, but with
   // multiple string children, then the `element` that gets passed in here is
   // a plain value -- a string or number.
@@ -111,6 +112,10 @@ function getData(element: Object): DataType {
     }
   }
 
+  if (nodeType === 'Native') {
+    reactTag = element._rootNodeID;
+  }
+
   return {
     nodeType,
     type,
@@ -125,6 +130,7 @@ function getData(element: Object): DataType {
     text,
     updater,
     publicInstance,
+    reactTag,
   };
 }
 
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -79,6 +79,10 @@ class Node extends React.Component {
       children = this.props.wrappedChildren;
     }
 
+    var reactTag = this.props.node.get('reactTag');
+    var name = reactTag == null
+      ? `${this.props.node.get('name')}`
+      : `${this.props.node.get('name')}(${reactTag})`;
     var collapsed = node.get('collapsed');
 
     var leftPad = {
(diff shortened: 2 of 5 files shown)


brentvatne:
cc @jaredly @gaearon

gaearon:
Can you show screenshots before and after?

gaearon:
Thanks for the PR. I think it's unlikely we'll be exposing implementation details like this to most users through the default UI, so please feel free to maintain a fork for this purpose.

gaearon:
Oops, I missed the part where you said it's useful to folks writing their own native modules (and not just RN contributors). Let me ask some folks about this and come back to you.

rigdern:
@gaearon Below are before and after screenshots for a React Native app.

### Before

image

### After

image

Notice that the each native primitive in the tree (grey) has a parenthesized number next to it representing the react tag of that primitive. Similarly, the selected node has a parenthesized react tag next to it (pink) in the inspection panel on the right because it's a native primitive.

### Discussion

As I mentioned in my PR description, we might want to disable this feature for ReactDOM and only keep it for React Native. I'm interested in hearing your thoughts on that.

Here's a scenario where this debugging feature would be useful. Imagine somebody has a React Native app and they just pulled down the latest code. They're testing some page of the app and touches seem to be ignored when tapping on some element. Maybe another element is eating the touch events. Which one? Using the native or JS debugger, they are able to determine that the element that is the touch target has a react tag of `26`. Now how do they correlate this with something meaningful within their app? Using the feature added by this PR, they could type `26` into the search field and see the interesting component, its props, its state, its ancestors, etc. This is plenty of information to diagnose the bug. Without this PR, I'm not sure of an easy way to go about diagnosing the bug just from knowing that the react tag of the perpetrator is `26`.

gaearon:
I think I'd like to wait around with this one for now. We are in the middle of rewriting React, and the way RN is wired up is changing. So I'm not sure that field will even exist. I'm happy to come back to this in a month or two after everything else is ready.

gaearon:
Please see my comment in https://github.com/facebook/react-devtools/issues/439#issuecomment-280077429

PR #455, opened 2016-11-16
Fix `forward slash` key listen for typing on another input

jhen0409:
Currently it will broken typing `/` on another input, like component props / state:

### DIFF
--- a/frontend/SearchPane.js
+++ b/frontend/SearchPane.js
@@ -53,8 +53,10 @@ class SearchPane extends React.Component {
   }
 
   onDocumentKeyDown(e) {
-    if (e.keyCode === 191 && // forward slash
-      this.input && e.target && e.target.nodeName !== 'INPUT'
+    if (
+      e.keyCode === 191 && // forward slash
+      e.target.nodeName !== 'INPUT' &&
+      this.input
     ) {
       this.input.focus();
       e.preventDefault();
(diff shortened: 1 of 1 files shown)


frontend/SearchPane.js
```diff
@@ -53,14 +53,6 @@ class SearchPane extends React.Component {
   }
 
   onDocumentKeyDown(e) {
-    if (e.keyCode === 191) { // forward slash
-      var doc = ReactDOM.findDOMNode(this).ownerDocument;
-      if (!this.input || doc.activeElement === this.input) {
-        return;
-      }
```
Timothee: I don't know that codebase very well, but maybe you could check on if `doc.activeElement` is an input field and `return` if that's the case? That would leave the functionality intact for the most part.

That's assuming that the prop/state field is actually an input field, but, if not, some other similar test can probably be performed.

jhen0409: @Timothee Thanks! Just change to check `doc.activeElement.nodeName === 'INPUT'` works good.

jhen0409:
Although I solved it by build extension myself, but I still hope it land here, after all, it's an obvious bug, and auto update is more convenient, so someone could review this? 😅

spicyj:
Maybe we could check if `e.target` is body?

jhen0409:
@spicyj I got an exception is when I click the checkbox above, the `e.target` is div, so I think check it isn't `input` may be better. And thanks! We can check `e.target` directly, so I removed the code for `activeElement`.

gaearon:
Thanks!

PR #457, opened 2016-11-21
Adds missing library and documentation for electron stand-alone shell.

### LINKED ISSUE #444: Exception running in electron shell: no such module 'ip'

jesusmaldonado:
The ip library was not installed and in addition the build scripts would call node_modules at the root level.

Changes
-Added ip package to package.json
-Extended, formatted Readme.md for clarity on startup instructions.

@jaredly @kassens @spicyj

### DIFF
--- a/shells/electron/Readme.md
+++ b/shells/electron/Readme.md
@@ -4,12 +4,14 @@ This is useful for working on a React Native app *without* debugging in
 chrome. The js runs in jscore, and the devtools talk to it through a
 websocket.
 
-You'll need webpack, and [electron](http://electron.atom.io/#get-started) (`npm install electron-prebuilt -g`).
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks for the PR! I separately fixed those issues a little differently (e.g. neither electron nor webpack are required globally now), and fixed the import. So I’ll close this PR but thanks for taking the time to send it.

Issue #20, opened 2014-01-09
Components displayed as <Unknown ... /> when declaring js variables at the top.


tiengtinh:
I'm using grunt task to first build *.jsx files from *.coffee, and then build *.jsx to *.js files.
Therefore, my *.js files would be something like

```
var PhotoListWrapperView, PhotoListView,  PhotoView;
...
PhotoListWrapperView = React.createClass({
...
});
```

And with this js style, when I inspected into the React tab in Chrome, things displayed like:

```
<Top Level>
    <Unknown data="...">...</Unknown>
</Top Level>
```

When I tried to manually change the js variable declaration style to `var PhotoListWrapperView = React.createClass({...` , React inspection displayed correctly.

Look like declaring like `var PhotoListWrapperView` is a must for the tool to work. This may lead to some inconveniences. My case is an example.

vkbansal:
Recently started using devtools and I'm facing similar problems
![capture](https://cloud.githubusercontent.com/assets/5930351/5971447/9c448dd8-a86b-11e4-9050-0e5b84ea683b.JPG)

Schtauffen:
I use coffeescript (without jsx), and it never knows what to name custom elements. 
Is there a variable I can set explicitly set to tip off the react devtools?

sebmarkbage:
You can set `displayName`

PR #459, opened 2016-12-01
Support Iterators/Iterables, Immutable, other inspector improvements

cmmcleod:
### Features / Improvements:
- Improve inspection/rendering of nested Arrays
- Improve inspection of TypedArrays (rendered similarly to Arrays)
- Improve display of ArrayBuffers and DataViews (Shows size, no deep inspection)
- Iterables are rendered as `<constructor name>(…)`
- Iterables are read only
- Iterables entries are rendered as lists/arrays
- Inspecting Iterables/Immutable values is greatly improved (but keyed collections need improving. See below)
- ES6 Map/Set objects no longer throw an error

Addresses much of Issue #52 and was inspired by PR #149, particularly @gaearon's [comment](https://github.com/facebook/react-devtools/pull/149#issuecomment-220366852)

#### Code Change Summary:
- Moved duplicated `getIn(base, path)` to separate file `/agent/getIn.js`
- Refactored `/agent/dehydrated.js` to improve readability (in addition to enhancements)
- Added `readOnly` and `uninspectable` propteries to `meta` for dehydrated values where appropriate
- `/fronted/DataView/DataView.js` `DataItem` Component check for `uninspectable` when rendering
- Added `typed_array, array_buffer, data_view, iterator` cases to `Propval.js` and `previewCoplex.js` in addition to some refactoring for readbility
- Added several props to empty elements to `test/example` to demonstrate changes.

#### Caveats / For future Improvement
Typed Arrays are set read only in the inspector with these changes - Editing them (in the current released version) would cause them to be converted to objects, but it should be possible to handle this.

Iterators/Iterators are read only for fairly obvious reasons, however it might be reasonable to implement a live-edit feature for built-in iterables (i.e. `Map`, `Set`, `TypedArray`).

#### Improvement of Keyed iterables/iterators
With these changes the inspector / `DataView` renders *entries* of *iterators* as Arrays of `[key, value]` arrays i.e. `0: Array[2], 1: Array[2], ...` rather than `keyA: <value>, keyB: <value>, ...`.
Determining if an iterable is a keyed iterable is tricky because an iterator entry can be of any value. Simply checking that entries are arrays of length `2` is insufficient because a Set of arrays (see `tuplesSet` in `example/test/target.js`) would 'break' this approach. In addition ES6 `Map` and `Immutable.Map` allow complex keys and use [Same-value-zero equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) to determine uniqueness. This means the keys/attribute strings in the inspectors `path` cannot easily be converted to strings through either `toString` or `JSON.stringify` without breaking correctness in some cases.

#### Requesting Feedback & Help for:
- Code style, improvements, etc
- UI/UX/Styling/Color tweaks for `Iterables`, `Typed Arrays`, etc?
- Ideas for handling keyed iterable case
- Community assistance in testing with other collection libraries e.g. [seamless-immutable](https://github.com/rtfeldman/seamless-immutable), etc

**Finally**
Substantial unit test coverage is on it's way, I just wanted to get some eyeballs on this first!


cmmcleod:
screen shot 2016-11-30 at 10 48 55 pm
screen shot 2016-11-30 at 10 49 14 pm
screen shot 2016-11-30 at 10 49 23 pm
screen shot 2016-11-30 at 10 52 28 pm
screen shot 2016-11-30 at 10 52 47 pm

gaearon:
Flow is failing, could you fix that?

cmmcleod:
Yes, forgot to run typecheck. Most of this is fixed in the latest commits, however I'm getting a peculiar failure now:

```
$ npm run typecheck

> @ typecheck ..../react-devtools
> flow check

node_modules/invariant/invariant.js.flow:3
  3: declare module.exports: (
                   ^ Unexpected token .


Found 1 error
```

Running node `v6.3.0`. Does `invariant` need to be added to the `.flowconfig` ignore list?

```
$ npm ls invariant

└─┬ babel-core@6.3.21
  └─┬ babel-traverse@6.19.0
    └── invariant@2.2.2
```

cmmcleod:
Running with node `v4.6.2` and npm `2.15.11` (same as TravisCI) works fine

agent/getIn.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+function getIn(base, path) {
+  return path.reduce((obj, attr) => {
+    if (obj) {
+      if (obj.hasOwnProperty(attr)) {
+        return obj[attr];
+      }
+      if (typeof obj[Symbol.iterator] === 'function') {
+        // Convert iterable to array and return array[index]
+        return [...obj][attr];
```
gaperton: 1. Won't there be n^2 complexity? getIn seems to be called in a loop, and we traversing the container every time it's called.
2. Will line 25 work with key-value maps, such as decorated records creating properties accessors for its members (so check in line #20 will fail)?

cmmcleod: 1. `getIn` was duplicated code moved fro Bridge.js#441 and Agent.js#401 and updated to support iterators. The complexity of the previous implementation hasn't changed (with the exception of spreading the iterable on line 25).

2. This does work with key-value maps (if they are iterable) with one caveat, in order to support maps where the key is not a primitive the tool must treat the `[key, value]` entry as an iterable as well resulting in something like this:

```
myMap: Map(..)
  0: [
    0: <complex key A>
    1:  "some value"
  ]
  1: [
    0: "string key B"
    1:  "some other value"
  ]
  ...
```

see **Improvement of Keyed iterables/iterators** in this [comment](https://github.com/facebook/react-devtools/pull/459#issue-192774346)

I hope I'm making sense here.

gaearon:
>Determining if an iterable is a keyed iterable is tricky because an iterator entry can be of any value

Can you use `iterable.entries === iterator`? I think this is what we used in React for a while (we are removing support for it though).

gaearon:
Can you rebase please? Sorry for the trouble. I'd like to give this a review.

cmmcleod:
Thanks for the `iterable.entries` suggestion. I just noticed this in React the other day infact, too bad it's being removed.

Will update and add tests (this may take a day or two for me to get around to it)

gaearon:
Thanks. Excited about this PR!

gaearon:
Ping, do you think you'll have time to finish this?

cmmcleod:
Yeah, pretty confident I'll have this updated by the end of this weekend!

gaearon:
Awesome.

cmmcleod:
Fixed the conflicts. 

Looking into using `iterable.entries` for Map types

cmmcleod:
Not sure how `iterable.entries` easily helps to distinguish a `Set` from a `Map`. For example in

target.js:369

```
  tuplesSet: new Set([
    [1, 2],
    [3, 4],
    [5, 6],
    [7, 8],
    [9, 10],
  ]),
```

would we expect the debugger to render something like this?

screen shot 2017-02-27 at 12 01 42

cmmcleod:
Potential diff for support entries (based on this PR)

https://github.com/cmmcleod/react-devtools/compare/master...cmmcleod:feature/support-entries

gaearon:
In particular, the check I suggested is [this one](https://github.com/facebook/react/blob/03464f06d2efd1f736e04f3b5648eabe07c69904/src/shared/utils/traverseAllChildren.js#L112-L114) (but with opposite boolean check).

See how it lets you tell `Set` from `Map`:

screen shot 2017-03-02 at 12 53 36 am

gaearon:
Did my comment make sense? Does it help?

cmmcleod:
Updated to support map entries which seems to be working quite nicely.

Not entirely satisfied with the solution for maps where the keys themselves are objects:

[Solution](https://github.com/facebook/react-devtools/pull/459/files#diff-e5b2c0fe1780e9c5969e3cc3295a41f3R415) - If map entry key is an object use key-value `[Object #]`
[Example](https://github.com/facebook/react-devtools/pull/459/files#diff-926c5a0eb9dcb030a49e05904f7076c2R327)

gaearon:
Can you show screenshots of what doesn't look nicely, and could be done better? You have more context on this than I do so please try to dumb it down for me 😄

cmmcleod:
```javascript
objMap: new Map([
    [{ a: 'a'}, true],
    [{ a: 'a'}, false],
    [{ b: 'b'}, [1, 2, 3]],
    [{ c: 'c'}, { k1: 'v1', k2: 'v2' }],
  ])
```

screen shot 2017-03-02 at 10 13 37

This looks ok I suppose, but not necessarily the most helpful for dev/debugging since you cannot directly identify the key. However if you are using Objects as keys then presumably you would typically be using other/additional debugging techniques..

gaearon:
How does Chrome display this?

gaearon:
Hmm, maybe we should just do this instead.

screen shot 2017-03-02 at 6 23 26 pm

What do you think? How hard would this be?

cmmcleod:
Pretty straightforward actually:

screen shot 2017-03-02 at 10 36 06

gaearon:
Nice! If you could just add the `[[Entries]]` thing to make it clear what's going on, that would be rad.

cmmcleod:
Cool, have this working now, but I'm just investigating/fixing a bug for nested iterables. Will update once I've fix it.

cmmcleod:
Hmm.. So nesting keyed entries under `[[Entries]]` has opened up a can of worms because it breaks the way that `getIn` is expected to work. 

I.e. getIn is expected to traverse the 'natural' path, but since we are adding `[[Entries]]` upon inspection this kind of throws a spanner in the works which means getIn becomes rather hacky and brittle.

I think either the initial implementation (without `entries` support) or the implementation mentioned in this [comment](https://github.com/facebook/react-devtools/pull/459#issuecomment-283734803) are saner options, but I'm open to suggestions.

gaearon:
OK, let's display it in tuples, like you did originally.

agent/Bridge.js
```diff
@@ -396,15 +397,26 @@ class Bridge {
...
+
     if (inspectable) {
       var val = getIn(inspectable, path);
       var protod = false;
       var isFn = typeof val === 'function';
+
+      if (val && typeof val[Symbol.iterator] === 'function') {
+        var iterVal = Object.create({});  // flow throws "object literal incompatible with object type"
+        var count = 0;
+        for (const entry of val) {
```
jaredly: this will get fairly unhappy for long lists. Can we bail after ~100 and indicate that there's more we're not showing?

agent/dehydrate.js
```diff
@@ -11,6 +11,60 @@
...
+    if (data instanceof DataView) {
+      return 'data_view';
+    }
+    return 'typed_array';
+  }
+  if (data instanceof ArrayBuffer) {
+    return 'array_buffer';
+  }
+
+  var type = typeof data;
```
jaredly: I think I have a slight preference for doing this first, as all of the above checks are only if `typeof -> 'object'`

agent/dehydrate.js
```diff
@@ -30,63 +84,72 @@
...
+          length: data.byteLength,
+          uninspectable: true,
+        },
+      };
+
+    case 'array':
+      if (level > 2) {
+        return createDehydrated(type, data, cleaned, path);
+      }
+      return data.map((item, i) => dehydrate(item, cleaned, path.concat([i]), level + 1));
```
jaredly: this would also be nice to limit to e.g. 100, but I'm fine w/ just putting a TODO

agent/Bridge.js
jaredly: this can probably be done as a separate diff, I just would rather an extension update not make people's apps suddenly really slow under debugging

agent/dehydrate.js
cmmcleod: Yeah, that's a good point, easily refactored. Will do

agent/dehydrate.js
cmmcleod: Adding a fixed limit seems like a reasonable (but temporary) compromise. 

I think the more ideal solution would be to add an optional `breadth` argument (defaulting to some small number (20-30) and then adding a `show more...` link/button in the interface which would increase the `breadth` and inspect+render the next 'page' of items (regardless of whether the inspected object is a iterable, array, map, etc).

I'm happy to update the PR with a hardcoded limit and add a TODO

gaearon:
@jaredly Thank you very much for reviewing this.

@cmmcleod Let’s land this after:

* Hardcoding a limit on 100 items with a temporary `(Only first 100 items were shown)` in UI or something like this.
* Refactoring the `typeof` thing.

cmmcleod:
Sounds good. Will get this done and updated tonight

cmmcleod:
Took me a lot longer to get around to this than I would have liked!

-  Refactored `typeof` in `dehydrate.js`
- Added hardcoded limit and TODO.

Unfortunately I didn't find any simple mechanism to add `(Only first 100 items were shown)`, but I've been working on a better solution which I'd like to submit in a separate PR (click to load more type of solution).

gaearon:
Could you fix lint please? Something minor:

```
/home/travis/build/facebook/react-devtools/test/example/target.js
  454:12  error  'i' is already declared in the upper scope  no-shadow
  508:14  error  'i' is already declared in the upper scope  no-shadow
```

gaearon:
Just a heads up: I'm in the middle of some other work but I'll get back to this as soon as I can. Likely within one or two weeks.

gaearon:
One potentially confusing thing I noticed is it still lets you *edit* stuff inside the entries (but this won’t work since they’re not real arrays).

gaearon:
Not going to hold this any longer. 😄 
This is an incredible PR, thank you so much!

I’ll create a new issue for the followup work.

gaearon:
Please see https://github.com/facebook/react-devtools/issues/608 for followup items.
Let me know if you’d like to work on these! I’d be happy to give you contributor permissions if that would help you move faster.

cmmcleod:
Thanks for the merge! I'm definitely up for looking into some of the remaining issues and will follow up on #608.

Issue #316, opened 2016-01-18
$r overwrites application's global $r variable


patbl:
> Selected component instance is available as $r from the console

(https://facebook.github.io/react/blog/2015/09/02/new-react-developer-tools.html)

The app I work on uses `$r` extensively ("r" is short for "route"). Could you assign that variable only if it's undefined?

kassens:
Yes, we should do that.

yinghaochan:
pretty please? @kassens

gaearon:
Fixed in https://github.com/facebook/react-devtools/pull/463.

PR #463, opened 2016-12-08
Don't overwrite global $r if it is already set

wrakky:
Thought I would try to get the ball rolling on #316.

Seems like there are two parts to this:
- Detect global $r and assign a different variable if it exists
- Pass this variable name to the view so it can display it in the sidebar

The `capabilities` object is passed to the frontend store as part of the initial connection handshake so this seemed like an easy place to pass through the name of the `$r` variable to the view but I'm not sure you can really say it's a capability so maybe it wants it's own way of being passed instead?

### DIFF
--- a/agent/Agent.js
+++ b/agent/Agent.js
@@ -88,7 +88,7 @@ class Agent extends EventEmitter {
   elementData: Map<ElementID, DataType>;
   roots: Set<ElementID>;
   reactInternals: {[key: RendererID]: Helpers};
-  capabilities: {[key: string]: boolean|string};
+  capabilities: {[key: string]: boolean};
   renderers: Map<ElementID, RendererID>;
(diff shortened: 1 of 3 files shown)


gaearon:
This only works if `$r` is assigned by the application immediately. I don't think this necessarily solves the use case in https://github.com/facebook/react-devtools/issues/316.

I would prefer that we keep it simple and just avoid reassigning it if we didn't assign it ourselves. We can have a separate hidden flag to check for it. Would you like to change the PR to do it instead?

It is fine to still display the message as if `$r` is available. If users set this global, they probably know what it means.

gaearon:
Looks good to me. Let's merge if CI passes.

gaearon:
Meh, I trust it works.

Issue #390, opened 2016-06-25
Display "..." on getters to defer their evaluation


ddsol:
In `Bridge.js` there's an access to [`val.__proto__[name]`](https://github.com/facebook/react-devtools/blob/dfe8977860d1ee4298fe18c34dc86012c3d8288e/agent/Bridge.js#L381), which seems like it should be okay. However, it is possible to define a getter that expects a complete instance, rather than a empty prototype.

For instance:

```
var proto = {
  get upperName() { return this.name.toUpperCase()}
};

var instance = Object.create(proto);
instance.name = 'Foo';
console.log(instance.upperName); //=> FOO

console.log(instance.__proto__.upperName); //Throws: Cannot read property 'toUpperCase' of undefined
```

Bridge.js could
- Fetch the property get result from the original instance and put the value on the `result` upon an exception.
- `defineProperty` any getters on the `__proto__` so they continue to refer to the same instance. Any readout could then throw (i.e. `result.__proto__.upperName` still throws), but referring tto `result[name]` would not throw.
- Fetch the property get result from the original instance and put the value on the `__proto__` instead upon an exception.
- Just show an error as the property value.

It seems to me the value should show somewhere, rather than an error displayed. In any case, as it stands now, it breaks the extension. The error can only be found in the console.

Also, property getters could throw for reasons other than an instance/proto issue. It's probably a good idea to catch errors thrown and display the error in the UI, rather than aborting the code flow of the extension. It could be as simple as storing the error as a string on the property or just calling `dehydrate(err)` on the caught error.

orangy:
I'm having the same issue with kotlin.js. `List` defines `size` property with custom getter. It expects full instance, not just prototype. Now it crashes when trying to inspect it when part of the props.

jaredly:
We could probably use [`getOwnPropertyDescriptor`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/getOwnPropertyDescriptor) to detect things that have getters & dehydrate them specially (chrome devtools does this, rendering the value as `...` until you click it & force evaluation)

ddsol:
Eval on request is technically superior since getters can have side effects.

ianstormtaylor:
Running into this in [`slate`](https://www.npmjs.com/package/slate) as well with getters defined.

gaperton:
PR #469 fixes this issue as well.

gaearon:
Happy to accept a standalone fix for this!

[...]

Issue #388, opened 2016-06-22
Show more info for Date objects.


bigblind:
I'm trying to debug a component that stores `Date` objects in its state, but with the current react dev tools, there's no information about them, other than that it's a Date.
<img width="288" alt="screenshot 2016-06-22 11 14 27" src="https://cloud.githubusercontent.com/assets/737493/16261225/1bda0786-386b-11e6-8cd4-d0bcc7ad4f90.png">

gaperton:
PR #469 contains the fix for the Date type. It's shown like this:

![fixed dates](https://cloud.githubusercontent.com/assets/469462/21735438/e6dd897c-d437-11e6-9b76-82d13e80b711.jpg)

Is it good enough?

laurentvd:
I think this is exactly what we need!

gaearon:
Since we didn't go with #469, happy to accept a PR fixing just this specific case.
@laurentvd @bigblind Would you like to contribute?

tarang9211:
@gaearon Is this still open? If no one's taking it, I would like to contribute.

laurentvd:
I'd love to contribute, but I'm a little short on time at the moment. If @tarang9211 would like to take this, that would be great!

zalmoxisus:
@gaearon we are using [`jsan`](https://github.com/kolodny/jsan) for Redux DevTools Extension. It supports unserializable by `JSON.stringify` data types like `date`, `regex`, `symbol`, `undefined`, `error`, and `function`. Also we'll add support for [global Symbols](https://github.com/kolodny/jsan/pull/11) and [ES6 Map and Set](https://github.com/kolodny/jsan/pull/12) in the next release.

I guess, you'll just have to replace `JSON.stringify(data)`/`JSON.parse(data)` with `jsan.stringify(data, null, null, true)`/`jsan.parse(data)`.

@kolodny did a great work on making it rather performant. Of course, it would be less performant than `JSON.stringify`, but if we'll traverse data to check for `Date` objects it wouldn't be as performant either.

[...]

Issue #83, opened 2015-04-08
Support Inspection of Immutable JS objects 


abritinthebay:
It's your own library guys - checking that it's Immutable and displaying the toJS output can't be that hard can it? ;)

But seriously - right now it doesn't even show the data when you expand it - just "fr" and an object with properties no-one care about.

Given that Immutable objects are becoming a _recommended_ pattern (and FB has a library for it) this seems like a no brainer.

I'll see if I can fork + patch but would love to know if you're working on this...

Daniel15:
+1 I'd love to see this too

phuson:
+1

wilkerlucio:
+1 I would love this feature as well

zpao:
Going to merge this into #52.

gaearon:
Immutable support is implemented in the latest version.

lukeknep:
@gaearon Thanks for digging into this. Is the latest version released? I just updated to 2.1.7 (released today) and am still seeing Immutable Maps and Records visualized like arrays of size-2 arrays (rather than key / value pairs).

gaearon:
I don’t think it’s too hard. Just needs someone dedicated to fixing it.

vdh:
@gaearon So will Immutable.js Maps get a better implementation? Maps with string keys are really hard to read like this.

PR #469, opened 2017-01-05
Generic support for third-party data objects inspection (ImmutableJS, BackboneJS, etc)

gaperton:
General problem is that sometimes developers are using data objects which are more sophisticated than raw JS objects and arrays. Typically, these objects are represented as some facade class at the top level holding an actual state inside. Navigating through such an objects in props as well as in state is clumsy and the property inspector is almost unusable.

There is a number of issues on this topic (https://github.com/facebook/react-devtools/issues/83, https://github.com/facebook/react-devtools/issues/52). And there was an attempt to solve it in the PR specifically for ImmutableJS: https://github.com/facebook/react-devtools/pull/149

In this PR I attempt to solve the problem in the way that it would be easy for data framework developers to customize the appearance of data objects in the inspector. I'm doing it to make [NestedReact](https://github.com/Volicon/NestedReact) models and collections look good in the inspector, but it's trivial to do it for any other data library including ImmutableJS (which I can do as well to resolve that pool of duplicate issues as soon as this PR will be accepted).

Idea is to attach state inspector functions to the object's prototype, which will take an object instance and return plain JS object or array (whatever is appropriate for the particular case). Whenever an object has state inspector it will be used to transform its state. Using BackboneJS as an example, something like this needs to be done to make models and collections look good:

```javascript
if (__REACT_DEVTOOLS_GLOBAL_HOOK__) {
  const { addInnerStateInspector } = __REACT_DEVTOOLS_GLOBAL_HOOK__;
  if (typeof addInnerStateInspector === 'function') {
    addInnerStateInspector( Backbone.Model, x => x.attributes );
    addInnerStateInspector( Backbone.Collection, x => x.models );
  }
}
```

Another example - `Date` instances currently appears as an empty object (https://github.com/facebook/react-devtools/issues/388). Which is [fixed by this PR](https://github.com/facebook/react-devtools/pull/469/files#diff-2b80787c27b4170cbf31cebaa24ced44R13) in the following way:

```javascript
addInnerStateInspector( Date, ( x : Date ) => ({
   local : `${ x.toDateString() } ${ x.toTimeString() }`,
   iso : x.toISOString(),
   timestamp : x.getTime(),
});
```

This API can be used [internally to improve the appearance of built-in JS classes](https://github.com/facebook/react-devtools/pull/469/files#diff-2b80787c27b4170cbf31cebaa24ced44R12), by data framework developers to make their custom classes appear well, and by regular developers as well.

In the case of the more sophisticated data structure, the corresponding raw JS representation can be calculated inside of the inspector functions and memorized in the data object's instance (which will work perfectly in case of ImmutableJS).

P.S.: One more issue this PR resolves (https://github.com/facebook/react-devtools/issues/390) is an exception being thrown when you're trying to inspect an object with `get prop(){...}` calculated properties defined on the prototype.

### DIFF
--- a/agent/Agent.js
+++ b/agent/Agent.js
@@ -410,14 +410,4 @@ function getIn(base, path) {
   }, base);
 }
 
-function addInnerStateInspector(Ctor, getInnerState) {
-  Object.defineProperty( Ctor.prototype, '__inner_state__', {
-    get : function() {
-      return getInnerState( this );
-    },
-  });
-}
-
-window.__REACT_DEVTOOLS_GLOBAL_HOOK__.addInnerStateInspector = addInnerStateInspector;
-
 module.exports = Agent;
--- a/agent/Agent.js
+++ b/agent/Agent.js
@@ -410,4 +410,14 @@ function getIn(base, path) {
   }, base);
 }
 
+function addInnerStateInspector(Ctor, getInnerState) {
+  Object.defineProperty( Ctor.prototype, '__inner_state__', {
+    get : function() {
+      return getInnerState( this );
+    },
+  });
+}
+
+window.__REACT_DEVTOOLS_GLOBAL_HOOK__.addInnerStateInspector = addInnerStateInspector;
+
 module.exports = Agent;
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -404,23 +404,27 @@ class Bridge {
     if (inspectable) {
       var val = getIn(inspectable, path);
 
-
-      var protod = false;
+      var protod = false, isWrapper = false;
       var isFn = typeof val === 'function';
-      var data = val._state || val;
 
-      Object.getOwnPropertyNames( data ).forEach(name => {
-        if (name === '__proto__') {
+      if( val && val._state ){
+          val = val._state;
+          isWrapper = true;
+      }
+
+      Object.getOwnPropertyNames( val ).forEach(name => {
+        if (name === '__proto__' && !isWrapper ) {
           protod = true;
         }
+
         if (isFn && (name === 'arguments' || name === 'callee' || name === 'caller')) {
           return;
         }
-        result[name] = dehydrate( data[name], cleaned, [name]);
+        result[name] = dehydrate( val[name], cleaned, [name]);
       });
 
       /* eslint-disable no-proto */
-      if (!protod && val.__proto__ && val.constructor.name !== 'Object') {
+      if (!protod && val.__proto__ && val.constructor.name !== 'Object' && !isWrapper ) {
         var newProto = {};
         var pIsFn = typeof val.__proto__ === 'function';
         Object.getOwnPropertyNames(val.__proto__).forEach(name => {
@@ -447,7 +451,7 @@ class Bridge {
 
 function getIn(base, path) {
   return path.reduce((obj, attr) => {
-    return obj ? ( ( obj._state || obj )[ attr ] ): null;
+    return obj ? ( ( obj._state || obj : obj )[ attr ] ): null;
   }, base);
 }
 
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -426,7 +426,7 @@ class Bridge {
         var newProto = {};
         var pIsFn = typeof val.__proto__ === 'function';
         Object.getOwnPropertyNames(val.__proto__).forEach(name => {
-          if( pIsFn && (name === 'arguments' || name === 'callee' || name === 'caller') ){
+          if (pIsFn && (name === 'arguments' || name === 'callee' || name === 'caller') ) {
             return;
           }
 
@@ -452,22 +452,24 @@ function getIn(base, path) {
   let isPrototypeChain = false;
 
   return path.reduce((obj, attr) => {
-    if( !obj ) return null;
-
-    // Mark the beginning of the prototype chain...
-    if( attr === "__proto__" ){
-      isPrototypeChain = true;
-      return obj.__proto__;
+    if (!obj) {
+      return null;
     }
 
-    if( isPrototypeChain ){
+    // Mark the beginning of the prototype chain...
+    if (attr === '__proto__') {
+      isPrototypeChain = true;
+      return obj[attr];
+    }
+
+    if (isPrototypeChain) {
       // Avoid calling calculated properties on prototype.
       const property = Object.getOwnPropertyDescriptor( obj, attr );
       return property.get || property.value;
     }
 
     // Traverse inner state of the third-party data frameworks objects...
-    return ( obj._innerState || obj )[ attr ];
+    return ( obj._innerState || obj )[attr];
   }, base);
 }
 
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -426,7 +426,7 @@ class Bridge {
         var newProto = {};
         var pIsFn = typeof val.__proto__ === 'function';
         Object.getOwnPropertyNames(val.__proto__).forEach(name => {
-          if (pIsFn && (name === 'arguments' || name === 'callee' || name === 'caller') ) {
+          if ( name === '__inner_state__' || ( pIsFn && (name === 'arguments' || name === 'callee' || name === 'caller') ) ) {
             return;
           }
 
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -403,16 +403,20 @@ class Bridge {
     var protoclean = [];
     if (inspectable) {
       var val = getIn(inspectable, path);
+
+
       var protod = false;
       var isFn = typeof val === 'function';
-      Object.getOwnPropertyNames(val).forEach(name => {
+      var data = val._state || val;
+
+      Object.getOwnPropertyNames( data ).forEach(name => {
         if (name === '__proto__') {
           protod = true;
         }
         if (isFn && (name === 'arguments' || name === 'callee' || name === 'caller')) {
           return;
         }
-        result[name] = dehydrate(val[name], cleaned, [name]);
+        result[name] = dehydrate( data[name], cleaned, [name]);
       });
 
       /* eslint-disable no-proto */
@@ -420,10 +424,13 @@ class Bridge {
         var newProto = {};
         var pIsFn = typeof val.__proto__ === 'function';
         Object.getOwnPropertyNames(val.__proto__).forEach(name => {
-          if (pIsFn && (name === 'arguments' || name === 'callee' || name === 'caller')) {
+          if( pIsFn && (name === 'arguments' || name === 'callee' || name === 'caller') ){
             return;
           }
-          newProto[name] = dehydrate(val.__proto__[name], protoclean, [name]);
+
+          var prop = Object.getOwnPropertyDescriptor( val.__proto__, name );
+
+          newProto[name] = dehydrate( prop.get ? prop.get : val.__proto__[name], protoclean, [name]);
         });
         proto = newProto;
       }
@@ -440,7 +447,7 @@ class Bridge {
 
 function getIn(base, path) {
   return path.reduce((obj, attr) => {
-    return obj ? obj[attr] : null;
+    return obj ? ( ( obj._state || obj )[ attr ] ): null;
   }, base);
 }
 
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -404,27 +404,25 @@ class Bridge {
     if (inspectable) {
       var val = getIn(inspectable, path);
 
-      var protod = false, isWrapper = false;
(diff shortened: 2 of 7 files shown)


gaperton:
If this PR is acceptable, I can implement support for ImmutableJS data structures as well.

zavelevsky:
+1 big time. Can save a lot of hassle.

wizzard0:
+1 too, integration is important

elektronik2k5:
I like the idea, but not the magic property `_innerState`. I suggest to replace it with a framework specific function (implemented inside each framework), which would register itself with the devtools (via a global, when devtools is active) and implement this logic on it's own: both the identification of whether a given object has this 'inner state', and what should it return.

Disclaimer: I don't really know how react devtools hooks work, so this is more of a pseudo-code than something usable:
```js
// inside Backbone:
if (__REACT_DEVTOOLS_GLOBAL_HOOK__) {
  const { addInnerStateInspector, } = __REACT_DEVTOOLS_GLOBAL_HOOK__
  if (typeof addInnerStateInspector === 'function') {
    addInnerStateInspector(function inspectBackboneModel(possiblyModel){
      // is this a Backbone Model?
      const isInspectable = possiblyModel instanceof Backbone.Model
      return {
        isInspectable,
        value: isInspectable ? possiblyModel.attributes : undefined, // this is the value to show
      }
    }
    addInnerStateInspector(function inspectBackboneCollection(possiblyCollection){
      // is this a Backbone Collection?
      const isInspectable = possiblyCollection instanceof Backbone.Collection
      return {
        isInspectable,
        value: isInspectable ? possiblyCollection.models : undefined, // this is the value to show
      }
    }
  }
}
```

gaperton:
>  I suggest to replace it with a framework specific function

This code behaves exactly in the same way, works faster, and framework developer doesn't need to know anything specific about dev tools hooks. So, what exactly is the benefit in making it more complex?

```javascript
if (__REACT_DEVTOOLS_GLOBAL_HOOK__) {
    Object.defineProperty( Backbone.Model.prototype, '_innerState', {
        get : function(){
            return this.attributes; // That's internal key-value hash.
        }
    });

    Object.defineProperty( Backbone.Collection.prototype, '_innerState', {
        get : function(){
            return this.models; // That's an array of models
        }
    });
}
```

elektronik2k5:
@gaperton the benefit is prevention of false positives. Any object can have a property called `_innerState` - in which case the devtools would **lie** about that object's contents. Any convention is bound to break at some point - it's only a matter of probability, and while rare, such a case would result in a terrible situation, which is basically impossible to debug. Who would assume that a devtool is buggy? It would be literally the very last place I'd check (if at all) - past the point of despair and well into insanity o_O

My solution **guarantees** interop and prevents accidental false positives. _You want to have that kind of certainty in order for your tools to be reliable._

aniddan:
👍

gaperton:
@elektronik2k5 I don't think that such a dramatic increase of the complexity as in your proposal (not to mention that it will work dramatically slower) worth potential benefits of avoiding the risk of occasionally overlapping name in the class member. As there are _way cheaper_ ways to mitigate such a risk. For instance, rename the property `_innerState` to something more specific, such as `__REACT_DEVTOOLS_EXPOSED_STATE_YOU_KIDDING_ME_CLAIMING_IT_WILL_OVERLAP__`.

> My solution guarantees interop and prevents accidental false positives. You want to have that kind of certainty in order for your tools to be reliable.

I don't see the solution yet. Just the proposal. But you're welcome to proof its superiority by actually implementing it and creating the PR.

gaperton:
> For instance, rename the property _innerState to something more specific

That's a good idea, in fact. `__inner_state__` should be practically safe enough.

wmill:
+1

gaperton:
Guys, please, respond. We need these three issues to be fixed.

gaearon:
@gaperton 

React team is currently busy with [rewriting React itself](https://www.youtube.com/watch?v=aV1271hd9ew) (https://github.com/facebook/react/issues/7925). We recognize it's annoying but we just don't have the capacity to review big changes right now.

If this was implemented as support built in on existing JS primitives (e.g. iterables) instead of adding APIs it would be easier to review. Most people who need Immutable (or other) support wouldn't guess that they need to also install a special hook for it to work. So I don't think it's a very good long term approach.

gaperton:
Dan, people who need Immutable support will just see Immutable stuff
working. They don't need to install any hooks. That API is for framework
developers.

Regarding iterables, there's PR based on them. It's harder to review as the
changes are far more complex, it introduces N^2 complexity instead of
linear (and it is important for large SPA), raises questions like how it
will works in projects requiring transpiration to ES5 (IE support is still
required), and have huge troubles iterating through key-value maps as the
key is not accessible.

And overall approach looks really weird if you would like to customize
appearance of something which is not a collection. Something like Date, for
instance. There are no natural reasons to make it iterable.

gaearon:
>people who need Immutable support will just see Immutable stuff
working. They don't need to install any hooks. That API is for framework
developers.

Maybe I misunderstood you, but is there a corresponding PR for Immutable that has been reviewed and approved by its maintainers? I'm open to merging this if there is a broad agreement on the hooks between library maintainers but I haven't seen this yet.

gaperton:
I will implement support for ImmutableJS if this PR will be accepted. Of
course, I'm not going to start a campaign of convincing all the framework
developer to use the API which is not implemented, nor I'm going to fix any
framework in advance just to see this PR to be rejected later. That's too
much, Dan.

gaperton:
Just to remind you - currently Date shown as empty object, attempt to
access objects with "get" properties throws exceptions, and there is no way
to make custom framework's object to look good at all. Community live with
these problems for more than half a year. They make Dev Tools almost
unusable on anything but small examples.

I see nothing harmful in introducing the experimental dev tools hook API.
What can go wrong with it, Dan? :)

gaearon:
>Just to remind you - currently Date shown as empty object

Could you send a fix for that in particular? It shouldn’t be hard to detect a native Date.

>attempt to access objects with "get" properties throws exceptions

Why? Could we also fix this by doing whatever Chrome DevTools is doing for them? It doesn’t seem like this would require custom hooks either.

>I will implement support for ImmutableJS if this PR will be accepted

The reason I’m asking is it is much harder to get support from libraries than from us. I would like to see Immutable maintainers to at least agree to an RFC proposal (you don't have to send a PR) before proceeding with this. Since you approach this as a generic solution that requires library opt-in it makes sense to first get some feedback and commitment from library authors on the API.

If you think Immutable maintainers won't accept this PR unless we also merge its support, doesn't it raise any issues about the approach itself?

gaperton:
> Could you send a fix for that in particular? It shouldn’t be hard to detect a native Date.

The fix for Date currently relies on `addInnerStateInspector` API. Of course, it can be hardcoded with all related type checks (and the solution size will be the same).

While it can be done, I don't understand why do you think it's beneficial to rewrite the code in the way that it would be hard to do the same thing for any other data type. Including RegExp, byte arrays, files, etc. What exactly is the benefit in *not exposing* an API which make fix for Date looks trivial?

gaperton:
>> attempt to access objects with "get" properties throws exceptions
> Why? 

Good question, Dan. Here's the ticket about it. https://github.com/facebook/react-devtools/issues/390

gaearon:
>Good question, Dan. Here's the ticket about it. #390 

Seems like it has a proposed solution: https://github.com/facebook/react-devtools/issues/390#issuecomment-229434241.

>The fix for Date currently relies on addInnerStateInspector API. Of course, it can be hardcoded with all related type checks (and the solution size will be the same).

Right—the difference is that unlike the solution with a new API, it's easy to merge because we don't need to agree on the API, only on the implementation.

>I don't understand why do you think it's beneficial to rewrite the code in the way that it would be hard to do the same thing for any other data type. Including RegExp, byte arrays, files, etc.

Once you add an API you can't easily break it as people come to rely on it. I'm not convinced this is the best API, or even that a public-facing API for injecting monitors is desirable. I agree though that native JavaScript objects that commonly appear as props should be rendered well.

gaperton:
> The reason I’m asking is it is much harder to get support from libraries than from us. I would like to see Immutable maintainers to at least agree to an RFC proposal (you don't have to send a PR) before proceeding with this. Since you approach this as a generic solution that requires library opt-in it makes sense to first get some feedback and commitment from library authors on the API.
> If you think Immutable maintainers won't accept this PR unless we also merge its support, doesn't it raise any issues about the approach itself?

Dan, I have no idea will they accept it or not. Regarding ImmutableJS, as I said - *I will implement support for it* and create the corresponding PR. If the PR we are discussing will be accepted. 

Being framework developer myself, I'm trying to make developer's life better. And I see absolutely no harm, even potential one, in adding experimental API to the dev hook. Even if ImmutableJS maintainers will decide to reject my PR fixing an appearance in Dev Tools, what wrong will happen? This PR is +75/-8 lines of code or so, what are we talking about? Is an idea to expose the hook bad by itself? Why? Will it stop anyone to add inspections for iterables? No, it's a good idea as well. So what?

gaperton:
> Seems like it has a proposed solution: #390 (comment)
<https://github.com/facebook/react-devtools/issues/390#issuecomment-229434241>
.

Maybe. It also looks like you're commenting in the PR which already
contains such a solution :) Is it bad, or what?

> Right—the difference is that unlike the solution with a new API, it's
easy to merge because we don't need to agree on the API, only on the
implementation.

Yes. Thing is that we're not discussing an API for some reason. Didn't even
started.

> Once you add an API you can't easily break it as people come to rely on
it.

Oh, yeah, it's terribly important, no doubts. Did you look at the API?

пт, 27 янв. 2017 г. в 19:59, Dan Abramov <notifications@github.com>:

gaearon:
I think that it is extremely unlikely a library like Backbone or Immutable would add React-specific hook support. For this reason, this change won't benefit most users anyway.

I also think that the implementation is a bit hacky (e.g. what is `__inner_state__`? why are we hardcoding a special case for this prop name? what if somebody has a legit use case for it?)

Hacky code is fine if it's worth it, but this just seems to adds complexity that we need to understand and maintain in the future, for little gain, as only a tiny portion of our users will know about the hooks (since, as I said earlier, I don't believe either Backbone or Immutable will add such hooks, based on my experience with open source).

A generic solution like https://github.com/facebook/react-devtools/pull/459 seems like a better way to achieve these goals.

I’m sorry I couldn’t help more.

I’m still happy to take PRs fixing specific cases without exposing new APIs (e.g. to fix Dates).

gaperton:
> I’m sorry I couldn’t help more.

I'm sorry, but it's me who couldn't help more. Given the quality of the discussion here (where you practically avoiding discussing the matter of the topic), it's highly unlikely that I will spend any more time contributing to the Dev Tools and any other of your repos.

Take care.

gaearon:
>where you practically avoiding discussing the matter of the topic and ignoring answers

I’m not sure I fully understand what you are referring to. It seems like this PR would only be useful if you could persuade maintainers of Immutable or Backbone to add such hooks. I’m not avoiding anything, I’m just claiming this won’t happen, and you haven’t given any arguments to counter this.

I’ve never seen Immutable/Backbone/etc adding view library-specific hooks, and React is not in some sort of unique position to dictate such integrations. What about Angular, Ember, Vue, Preact, etc? It doesn’t make sense for Immutable or Backbone authors to make special concessions for React. If you think it does please explain how this will happen. Talk to those maintainers, invite them to this discussion, demonstrate that you can coordinate this effort. It is your proposal after all. I don’t want to merge something without at least a basic plan, which you have not provided.

Your proposal is basically “let’s merge and see what happens”. My answer is the that won’t work—not because I’m avoiding or ignoring you, but because it rests on unrealistic assumptions. I explained why they are unrealistic right above. If you don’t agree please let me know via counter-arguments rather than via assuming bad faith behavior on my end. Thanks!

benjamingr:
@gaearon - @gaperton is a known problematic contributor and we've had issues with him before in other projects. The group of people who commented are all from the same Facebook group where he ranted about it - and you're being the responsible adult here.

I know you know this - but I just thought it's worth to have someone else say this out loud. Rock on.

gaperton:
> is a known problematic contributor and we've had issues with him before in other projects

Oh, yes, I am :). I don't remember we ever intersected on github, Benjamin, but I perfectly remember that I blocked you on FB for continuous personal insults :). Let's fix it on github as well.

gaearon:
Going to lock this thread as it doesn’t seem very productive.
I believe I mentioned the possible follow up items:

* If you’re confident with the approach in this PR, please demonstrate commitment from third party projects you’re targeting
* Feel free to send PRs to send specific cases (e.g. Dates)
* Let’s not attack each other or assume bad intentions

Cheers!

Issue #429, opened 2016-09-12
Test Firefox multiprocess/e10s compatibility


ngyikp:
Seems to work fine on my Firefox 51 nightly with multiprocess enabled, just need to add [`"multiprocess": true`](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json#permissions) to the [package.json](https://github.com/facebook/react-devtools/blob/master/shells/firefox/package.json) after more confirmation.

<img width="895" alt="screen shot 2016-09-12 at 4 59 57 pm" src="https://cloud.githubusercontent.com/assets/6135313/18430355/c5729c4c-790a-11e6-943f-25389c52bf44.png">

jnachtigall:
Works for me with Firefox Developer Edition 50 which is e10s enabled. So I agree that the manifest should be updated to report the addon as compatible so that it does not block the e10s rollout:

* https://blog.mozilla.org/addons/2016/08/02/multi-process-firefox-and-add-ons-a-call-to-action/
* https://developer.mozilla.org/en-US/Add-ons/Install_Manifests#multiprocessCompatible

jlandersen:
Firefox 50 was just released where e10s is enabled if all addons have the manifest property - would be nice to have this in :-)

PR #471, opened 2017-01-16
Enable multiprocess mode in Firefox

pluma:
See #429 for discussion.

Not supporting multiprocess mode literally means "Firefox UI and all tabs run on the same CPU" which is really bad. The comments suggest that the add-on seems to work fine if multiprocess mode is forcefully enabled so this change should not break anything.


gaearon:
OK, let's give it a try!

PR #475, opened 2017-01-17
Support Fiber

gaearon:
This works on top of https://github.com/facebook/react/pull/8818.
We diff trees on every commit and create a list of events to send to the hook.


backend/attachRendererFiber.js
```diff
@@ -0,0 +1,244 @@
...
+      const event = events[i];
+      hook.emit(event._event, event);
+    }
+  }
+
+  function handleFiberCommit(root) {
+    const current = root.current;
+    const alternate = current.alternate;
+    const events = [];
+    if (alternate && subscription != null) {
```
sebmarkbage: What if you're attaching to an existing tree that already has an alternate?

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,244 @@
...
+    // Just like we do in the Fiber child reconciler.
+    // We shouldn't be allocating Maps all the time.
+    const allKeys = new Set();
+    const prevChildren = mapChildren(prevFiber, allKeys);
+    const nextChildren = mapChildren(nextFiber, allKeys);
+    allKeys.forEach(key => {
+      const prevChild = prevChildren.get(key);
+      const nextChild = nextChildren.get(key);
+      if (prevChild != null && nextChild != null) {
+        // TODO: are there more cases when we can bail out?
```
sebmarkbage: Currently, either all children are cloned or none are cloned. So you really only need to check `nextFiber.child === prevFiber.child` to know if all are the same or all are different.

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,244 @@
...
+    if (isRoot) {
+      events.push({
+        element: fiber._debugID,
+        renderer: rid,
+        _event: 'root',
+      });
+    }
+  }
+
+  function enqueueUpdate(fiber, events) {
```
sebmarkbage: If props, state and the others have reference equality, you shouldn't need to push any events here. Helpful for deep updates since nothing along the path will have been updated.

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,244 @@
...
+        }
+      }
+      return;
+    }
+  }
+
+  function updateFiber(nextFiber, prevFiber, events) {
+    // TODO: optimize for the common case of children with implcit keys.
+    // Just like we do in the Fiber child reconciler.
+    // We shouldn't be allocating Maps all the time.
```
sebmarkbage: You shouldn't even need to diff on keys and types (which is incomplete anyway) since you can match on object equality between the fiber alternates.

If the first child is equal, then all the children are equal. If the first child is non-equal, you know that you will need to enqueue and update on everything that has an alternate and a mount on anything without an alternate.

The only thing you might need diffing for is unmounts. One solution is to just loop through all the next children with alternates. Add all the alternates to a Set. Then loop through all the previous children, if they're not in the Set, they're deleted.

A better solution might be to just expose a hook for deletions instead.

backend/getDataFiber.js
```diff
@@ -0,0 +1,114 @@
...
+    state,
+    context,
+    children,
+    text,
+    updater,
+    publicInstance,
+  };
+}
+
+module.exports = getFiberData;
```
bvaughn: nit: File name is `getDataFiber`; exports function `getFiberData`. The latter sounds less awkward.

sebmarkbage: same thing with `attachFiberRenderer`.

backend/attachRendererFiber.js
gaearon: Then `subscription` would be null (since we haven't finished subscribing, and the React side calls the commit hook synchronously during subscription the first time).

backend/getDataFiber.js
gaearon: We have existing `getData013`, `getData012`. I was striving for the all to appear together with their non-Fiber equivalents. But I don't care, can change for readability too.

gaearon: (Didn't notice the inconsistency—probably because I changed it some point. Which way do you prefer?)

backend/attachRendererFiber.js
sebmarkbage: When React calls it the first time (synchronously), then there can already be an alternate though, right?

backend/getDataFiber.js
jaredly: I'd maybe vote for consistency w/r/t getData12? so `getDataFiber`

backend/attachRendererFiber.js
gaearon: Right. But it's `&&` so we would go into the other branch which always mounts.

backend/attachRendererFiber.js
gaearon: Is this an implementation detail that will hold true? I thought we might change this in the future as optimization but wasn't sure.

backend/attachRendererFiber.js
gaearon: Nice, thanks for explaining!

gaearon: >You shouldn't even need to diff on keys and types (which is incomplete anyway) 

Why is it incomplete though? What did I miss?

backend/attachRendererFiber.js
gaearon: I'd still need to compare children though, don't I? So that would be reference equality for everything except child ID array (which I'd compare shallowly).

gaearon:
Ready for another review.

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,252 @@
...
+    } else {
+      // Non-root fibers are deleted during the commit phase.
+      // They are deleted in the child-first order. However
+      // DevTools currently expects deletions to be parent-first.
+      // This is why we unshift deletions rather than push them.
+      pendingEvents.unshift(event);
+    }
+    opaqueNodes.delete(fiber);
+    if (fiber.alternate != null) {
+      opaqueNodes.delete(fiber.alternate);
```
bvaughn: If you stored the value value returned on line 127 you would only have to delete that (and could avoid checking `fiber.alternate` again).

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,252 @@
...
+    }
+  }
+
+  function mountFiber(fiber) {
+    // Depth-first.
+    // Logs mounting of children first, parents later.
+    let node = fiber;
+    outer: while (true) {
+      if (node.child) {
+        node.child.return = node;
```
bvaughn: Why does this function modify `return` pointers? That doesn't seem right. Am I reading this wrong? (Also when would `node.child.return` ever _not_ equal `node`?)

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,252 @@
...
+  function enqueueMount(fiber) {
+    pendingEvents.push({
+      // TODO: the naming is confusing. `element` is *not* a React element. It is an opaque ID.
+      element: getOpaqueNode(fiber),
+      data: getDataFiber(fiber, getOpaqueNode),
+      renderer: rid,
+      _event: 'mount',
+    });
+
+    const isRoot = fiber.tag === 3;
```
bvaughn: nit Because of references like this and the ones in `getDataFiber` I wonder if it wouldn't be clearer to define a duplicate `ReactTypeOfWork` const in this project. At least this way if a value ever changes it's clear where all of the mirrored values are (and we aren't just greping for numbers).

backend/getDataFiber.js
```diff
@@ -0,0 +1,148 @@
...
+      publicInstance = fiber.stateNode;
+      props = fiber.memoizedProps;
+      state = fiber.memoizedState;
+      if (publicInstance != null) {
+        context = publicInstance.context;
+        if (context && Object.keys(context).length === 0) {
+          context = null;
+        }
+      }
+      const inst = publicInstance;
```
bvaughn: Why is the `inst` var needed?

backend/installGlobalHook.js
```diff
@@ -57,6 +65,33 @@ function installGlobalHook(window: Object) {
           this._listeners[evt].map(fn => fn(data));
         }
       },
+      // Fiber-only:
+      supportsFiber: true,
```
sebmarkbage: Is this unused now?

backend/installGlobalHook.js
```diff
@@ -57,6 +65,33 @@ function installGlobalHook(window: Object) {
...
+        return this._fiberRoots[rendererID] || (this._fiberRoots[rendererID] = new Set());
+      },
+      onCommitFiberUnmount: function(rendererID, fiber) {
+        // TODO: can we use this for roots too?
+        if (this.helpers[rendererID]) {
+          this.helpers[rendererID].handleCommitFiberUnmount(fiber);
+        }
+      },
+      onCommitFiberRoot: function(rendererID, root) {
+        const mountedRoots = this.getFiberRoots(rendererID);
```
sebmarkbage: It's unfortunate that the use of `this` instead of closure capture invalidates the suggestion I had in https://github.com/facebook/react/pull/8818/files#r98101132

backend/installGlobalHook.js
```diff
@@ -57,6 +65,33 @@ function installGlobalHook(window: Object) {
...
+        const current = root.current;
+        const isKnownRoot = mountedRoots.has(root);
+        const isUnmounting = current.memoizedState == null || current.memoizedState.element == null;
```
sebmarkbage: cc @acdlite This is regard to roots but I don't really have a better idea.

We could possibly make `onCommitFiberRoot` accept the direct children of roots instead of the actual root but it's nice to be able to have the mount container target here too for various debug data.

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,252 @@
...
+      });
+    }
+  }
+
+  function enqueueUpdateIfNecessary(fiber) {
+    const nextData = getDataFiber(fiber, getOpaqueNode);
+    const prevData = getDataFiber(fiber.alternate, getOpaqueNode);
+    // Avoid unnecessary updates since they are common
+    // when something changes deep in the tree due to setState.
+    if (!hasDataChanged(prevData, nextData)) {
```
sebmarkbage: Would it be better to do this at the fiber level instead? Maybe it's more fragile to updates but faster since you don't have to create the other objects.

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,252 @@
...
+    }
+    return false;
+  }
+
+  function hasDataChanged(prevData, nextData) {
+    return (
+      prevData.ref !== nextData.ref ||
+      prevData.source !== nextData.source ||
+      prevData.props !== nextData.props ||
+      prevData.state !== nextData.state ||
```
sebmarkbage: During a forceUpdate I don't think state object reference has necessarily changed and yet the component can have been mutated deep below.

backend/attachRendererFiber.js
gaearon: In React source, we do this in all traversals to avoid inifinite loops. They happen if return points to alternate of the parent you're traversing.

I'm not sure if it's necessary here but I did it for extra safety. It should be fine but maybe @spicyj or @acdlite can verify.

backend/getDataFiber.js
gaearon: To avoid next lines breaking over 80 character lint :-)
I can restructure.

backend/installGlobalHook.js
gaearon: No:

```js
if (
  typeof __REACT_DEVTOOLS_GLOBAL_HOOK__ !== 'undefined' &&
  typeof __REACT_DEVTOOLS_GLOBAL_HOOK__.supportsFiber
) {
```

Necessary so that Fiber doesn't try to attach to non-Fiber DevTools and crash them.

backend/attachRendererFiber.js
gaearon: :-(

backend/attachRendererFiber.js
bvaughn: > They happen if return points to alternate of the parent you're traversing.

Wouldn't this only happen if there was a bug in fiber?

I expect these `return` assignments in devtools are basically neutral since the pointers should be correct to begin with, but it still seems weird to see something _outside of_ `React` setting values on fibers.

backend/getDataFiber.js
bvaughn: Fair enough! 🤣

backend/attachRendererFiber.js
gaearon: >Wouldn't this only happen if there was a bug in fiber?

I don't know if Fiber guarantees that `return`s will match up anywhere. Fiber itself *always* reassigns them when traversing so that's what I did. I understand it seems a bit fishy but I don't see any harm in this.

backend/installGlobalHook.js
sebmarkbage: If we release a new version of the devtools that auto-updates everyone then before we release Fiber we can revert this check with the assumption that devtools gets updated before renderers?

backend/attachRendererFiber.js
sebmarkbage: It wouldn't be too bad to use recursion here instead if you wanted to avoid the concern and the implementation details of return. But then again, maybe this could should live colocated with the React Fiber source code so it feels less smelly.

backend/attachRendererFiber.js
sebmarkbage: cc @acdlite

So, we need to know this information for updates too. Currently we schedule an update effect tag that will be consumed during the commit. You might be able to use that. Ideally we would only do that if there exists a `componentDidUpdate` life-cycle but maybe we can keep doing that for this purpose too.

gaearon: In 040eaa040596ce3c8084bbbc4a5d626281ee52da, I check `.updateQueue.hasForcedUpdate` on the committed work and it seems to be fine. Can I rely on this?

backend/attachRendererFiber.js
gaearon: Addressed by 09517f2da876f69cc2ce6d266aed795cd424107b.

backend/attachRendererFiber.js
gaearon: Meh, we rely on implementation details a lot anyway. I think it's fine if you don't mind. I don't want to further bloat the bundle since we decided to enable this in prod.

backend/attachRendererFiber.js
```diff
@@ -0,0 +1,252 @@
...
+    }
+  }
+
+  function enqueueMount(fiber) {
+    pendingEvents.push({
+      // TODO: the naming is confusing. `element` is *not* a React element. It is an opaque ID.
+      element: getOpaqueNode(fiber),
+      data: getDataFiber(fiber, getOpaqueNode),
+      renderer: rid,
+      _event: 'mount',
```
gaearon: Mostly I wasn't sure if any downstream code handling these events gives any meaning to `event.type` or other property names. I'll check and rename. 😄

backend/attachRendererFiber.js
gaearon: Addressed

backend/attachRendererFiber.js
gaearon: Addressed.

backend/installGlobalHook.js
gaearon: Yes I think so.

PR #480, opened 2017-02-01
Fix number input of new attribute for React Native style editor

jhen0409:
I didn't fix the part of new attribute in #392, this is a continuation.


spicyj:
cc @kevinbarabash -- this should fix your issue.

PR #483, opened 2017-02-02
Hack around displayName issues in FB products (Simple)

gaearon:
Same purpose as #482, but much less hacky.

This will work best, but it requires that we also move `ReactComponent` and `ReactPureComponent` exports behind an object so that they don't get a `displayName` internally (https://github.com/facebook/react/pull/8918).

### DIFF
--- a/backend/getDisplayName.js
+++ b/backend/getDisplayName.js
@@ -11,6 +11,15 @@
 'use strict';
 
 const FB_MODULE_RE = /^(.*) \[from (.*)\]$/;
+
+// Hack: base class names are hardcoded because of internal FB transform that puts
+// displayName on all modules. It resulted in displayName of React base classes shadowing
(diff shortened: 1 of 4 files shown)


backend/getDisplayName.js
```diff
@@ -0,0 +1,42 @@
...
+  // Facebook-specific hack to turn "Image [from Image.react]" into just "Image".
+  // We need displayName with module name for error reports but it clutters the DevTools.
+  const match = displayName.match(FB_MODULE_RE);
+  if (match) {
+    const componentName = match[1];
+    const moduleName = match[2];
+    if (componentName && moduleName) {
+      if (
+        moduleName === componentName ||
+        moduleName.startsWith(componentName + '.')
```
gaearon: This is ES6 but we use WeakMap already everywhere so I assume it's fine

gaearon:
#movefast

PR #485, opened 2017-02-04
[Electron] Don't attach highlighter on React Native

gaearon:
This fixes an issue I'm seeing trying to hook up electron DevTools to RN.

screen shot 2017-02-04 at 21 13 15


jaredly:
this is a good start. we'll have to do a bit more to get highlighting to work on RN, but at least it won't break

Issue #178, opened 2015-09-03
Submitting to Opera Addons Catalog


shwetank:
This can also be submitted to Opera users. 

The Opera addons catalog also accepts crx files too, so its just a matter of the official creators just submitting the chrome extension to the https://addons.opera.com and it will be available to Opera users too. 

(I can help out with any others things related to this issue).

ryentzer:
I'd like to see this added to Opera also.

omgaz:
Bump

b2m9:
I would also like tot see it in the Opera store. I'm running v0.14.11 on Opera 38 without major problems (there are some minor issues, but I have experienced them in Chrome too).

StephanBijzitter:
bump

gaearon:
We won't be maintaining it in yet another store, sorry!
Hopefully this works for you: https://github.com/facebook/react-devtools/pull/486.

PR #486, opened 2017-02-06
Add installation information for Opera users

StephanBijzitter:
All Chrome extensions work natively on Opera, but the Chrome webstore checks the useragent string before it shows the download button. This instructs Opera users how they can enable the Chrome webstore for Opera, and thus use the official Chrome release.

Auto-updating and everything works.

I quickly wrote this down while #178 seems to have a low priority.


gaearon:
Thanks.

PR #492, opened 2017-02-07
use idiomatic syntax for npm scripts commands

everdimension:
An insignificant change, but it kind of confused me when I glanced at the `scripts` section. Made me wonder if there was any reason for writing out command paths explicitly, but as I see these commands are just for linting and testing within the repository itself, so the current code is superfluous. An idiomatic npm way is to just use the names of the commands and the local package will be used by default.


gaearon:
Yep, thanks.

PR #494, opened 2017-02-07
Add connectToDevTools(options) for React Native

gaearon:
Currently RN essentially [uses a fork](https://github.com/facebook/react-native/blob/75c14e3674e90a9b67335b896c39d3618fb37ac1/Libraries/Core/Devtools/setupDevtools.js#L25-L45) of `backend/websocketConnect.js`.

RN has a few special needs:

* Customizing host and port.
* Injecting RN style resolver.

In this PR, I am adding a new API to `react-devtools-core` that works like `require('react-devtools')` but provides additional customization for sophisticated consumers like RN. My work-in-progress rewrite of RN's `setupDevtools` now looks like this:

```js
function setupDevtools() {
  var {connectToDevTools} = require('react-devtools-core');
  connectToDevTools({
    // Special case: Genymotion is running on a different host.
    host: (Platform.OS === 'android' && NativeModules.AndroidConstants) ?
      NativeModules.AndroidConstants.ServerHost.split(':')[0] :
      'localhost',
    // Read the optional global variable for backward compatibility.
    // It was added in https://github.com/facebook/react-native/commit/bf2b435322e89d0aeee8792b1c6e04656c2719a0.
    port: window.__REACT_DEVTOOLS_PORT__,
    resolveRNStyle: require('flattenStyle'),
  });
}
```

It is still very flaky and often doesn’t connect, but the module boundaries themselves seem fine to me, and I’ll be cherry-picking some changes RN has made to the server so that it’s less flaky. (Might be some race condition or connection issue.)

So this is work in progress but I’d like feedback in case I’m going in a wrong direction.


backend/websocketConnect.js
```diff
@@ -10,13 +10,14 @@
  */
 'use strict';
 
-function websocketConnect(uri: string, WebSocket?: (val: string) => Object) {
-  WebSocket = WebSocket || window.WebSocket;
```
gaearon: I couldn't find where this is being used.

jaredly: ok - I think it was when I was experimenting with embedding in a webview or somewhere that didn't have a websocket impl.

jaredly:
Looks good!

PR #497, opened 2017-02-08
[Electron] Add loadingStatus handler for replace `#loading-status` text set

jhen0409:
Currently we need to set `#loading-status` div to make `react-devtools-core/standalone` work, provide a way to replace it will be better for me:

```js
require('react-devtools-core/standalone')
  .setContentDOMNode(document.getElementById('container'))
  .loadingStatus(message => console.log(message))
  .startServer(port);
```


gaearon:
Awesome, thanks! I tweaked the API a tiny bit in https://github.com/facebook/react-devtools/commit/dd9d9411e6c9b3f597f56c529c6d24ee2f10e854.

PR #498, opened 2017-02-08
[Electron] Remove indirection

gaearon:
This vastly simplifies the Electron shell.
Changes:

* Remove the `eval` dance and merge backend and embed together
* Remove the code that was only necessary when RN packager proxied socket connection
* Backport reconnection logic and memory leak fix from RN


packages/react-devtools-core/src/backend.js
```diff
@@ -8,60 +8,134 @@
...
+    setupBackend(wall, resolveRNStyle);
+  };
+
+  var hasClosed = false;
+  function handleClose() {
+    if (!hasClosed) {
+      hasClosed = true;
+      setTimeout(() => {
+        connectToDevTools(options);
+      }, 2000);
```
jaredly: we might want to drop this to 500 -- 2 seconds feels like a long time to wait for reconnection

gaearon: This was intentional change from RN team because they had some issues with 200ms reconnection (it was making things slow or something). 500ms is better than 200ms but I don't want to touch it without understanding the issues they had in the first place. Maybe @yungsters or @javache can provide more details.

jaredly: cool sounds good

Issue #489, opened 2017-02-06
Consolidate packages


gaearon:
Currently there's a lot of forking going on which makes it harder to figure out how everything fits together. For example:

* Injecting into the hook
  - React tries to do this [here](https://github.com/facebook/react/blob/b49611e643fd6baa5214b9ebaae347f185c4ad5b/src/renderers/native/ReactNativeStack.js#L53-L79)
  - But actually the hook doesn't exist yet until [injected](https://github.com/facebook/react-native/blob/75c14e3674e90a9b67335b896c39d3618fb37ac1/Libraries/Core/Devtools/setupDevtools.js#L70) so RN just copies the same code [here](https://github.com/facebook/react-native/blob/75c14e3674e90a9b67335b896c39d3618fb37ac1/Libraries/Core/Devtools/setupDevtools.js#L76-L96)

* Setting up the connection
  - Happens [here](https://github.com/facebook/react-native/blob/75c14e3674e90a9b67335b896c39d3618fb37ac1/Libraries/Core/Devtools/setupDevtools.js#L25-L45) in RN
  - However Electron shell has [very similar code](https://github.com/facebook/react-devtools/blob/5261d3af67050154e2c021db18aa0bad2b951af0/backend/websocketConnect.js#L14-L42)

* Nuclide [vendors the Electron build](https://github.com/facebook/nuclide/tree/master/pkg/nuclide-react-inspector/VendorLib/dev-tools/build) for the Inspector (although it only uses the client while RN runs its fork of the server).

* [Backends](https://github.com/facebook/react-devtools/tree/master/backend) depend on React internals but aren’t colocated with React

---

The goals for consolidation:

* Reduce the number of forks to make it easier to make changes across them (e.g. to support Fiber).
* Make it easy for downstream packages (RN, Nuclide) to accept changes related to DevTools by packaging them.
* Consolidate the code relying on React internal data structure inside the React repo.

The plan for consolidation:

- [x] Publish `react-devtools` package added in #487 
- [x] Publish `react-devtools-core` separately
- [x] Make [`nuclide-react-inspector`](https://github.com/facebook/nuclide/tree/master/pkg/nuclide-react-inspector) depend on it instead of vendoring its build (https://github.com/facebook/nuclide/pull/1016)
- [x] Rewrite RN [`setupDevtools()`](https://github.com/facebook/react-native/blob/master/Libraries/Core/Devtools/setupDevtools.js)
[...]

PR #501, opened 2017-02-09
[Electron] Add `[React DevTools]` head for logs

jhen0409:
Related to https://github.com/facebook/react-devtools/issues/489#issuecomment-278260740, but I think keep logs will be better, add `[React DevTools]` head for logs may be more clearly in the [rndebugger](https://github.com/jhen0409/react-native-debugger).


gaearon:
Thanks.

PR #504, opened 2017-02-10
Call teardown() when Panel will unmount

jhen0409:
(Explanation is now [here](https://github.com/facebook/react-devtools/pull/504#issuecomment-278985858))

### DIFF
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -96,6 +96,7 @@ class Panel extends React.Component {
     if (this._unsub) {
       this._unsub();
     }
+    this.teardown();
   }
 
   pauseTransfer() {
--- a/packages/react-devtools-core/src/standalone.js
+++ b/packages/react-devtools-core/src/standalone.js
@@ -85,7 +85,9 @@ function initialize(socket) {
       listeners.push(fn);
     },
     send(data) {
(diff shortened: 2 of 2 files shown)


gaearon:
Why does it error in the first place? I'd like to fix the cause instead of silencing errors.

jhen0409:
I take a time to check it, looks like it haven't clear all event listeners when `unmount` panel, I got the old `socket` instance:

```js
window.sock = socket;
wall = {
    ...
    send: (data) => {
      // sometimes false, it should got the old socket instance
      console.log(window.sock === socket)
      socket.send(JSON.stringify(data));
    },
   ...
}
```

I updated the PR, call `this.teardown()` when `componentWillUnmount` in Panel will fix the problem, but we may need to avoid send `shutdown` with closed socket, so I still add the `readyState` check.

gaearon:
@jaredly Can you review this please?

gaearon:
LGTM

jhen0409:
@gaearon it seems not built in `react-devtools-core@2.0.5`, I can't seen this patch in the bundle. 😅

gaearon:
Apologies, I probably forgot to rebuild. Should be in 2.0.6.

PR #508, opened 2017-02-12
[Electron] Add missing ws server.close()

jhen0409:
Currently the client socket not closed when called `standalone.close()`, it only closed http.


gaearon:
Awesome, thank you!

PR #509, opened 2017-02-13
Avoid `forward slash` key handle when typing on contentEditable element

jhen0409:
It's not affect the current extension. In case of I use `contentEditable` div and  `react-devtools-core/standalone` together, it will break the `/` input on div.


gaearon:
Thanks!

PR #510, opened 2017-02-13
Allow reassignment of the relay install hook

jquense:
Lets the property be reassigned, so other consumers beside the devtools can instrument relay.

I realize this is mostly secret, but It's super helpful as a general instrumentation hook. In particular I'm tying into this to improve e2e testing. Moving tests along when Relay is done making requests is a good bit nicer then arbitrary `.sleep()` commands.

:)

### DIFF
--- a/plugins/Relay/installRelayHook.js
+++ b/plugins/Relay/installRelayHook.js
@@ -132,6 +132,7 @@ function installRelayHook(window: Object) {
 
   var _relayInternals = null;
   Object.defineProperty(hook, '_relayInternals', ({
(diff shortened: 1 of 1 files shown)


gaearon:
Hush hush.

gaearon:
Oops, that didn't work.

screen shot 2017-02-15 at 01 11 32

gaearon:
I think your intention was https://github.com/facebook/react-devtools/commit/eb96aaf8ad3c6188f6b7e08131f0d703670d2bd5.

Issue #481, opened 2017-02-02
TreeView not visible if too many components in hierarchy


Cypcer:
Hi,

Is it possible to hide or drag panel under the TreeView? If many components are nested is not possible to see TreeView (see attachment).

![treeview not visible if many components are inspected](https://cloud.githubusercontent.com/assets/9745663/22546766/b549c648-e93e-11e6-87e7-d19db228a8d4.jpg)

Kind regards
Rafal

gaearon:
Would you like to contribute to fix this? I agree it’s annoying.
Maybe we could make that area scrollable.

Cypcer:
Currently I'm switching between 3 projects, so at the moment I cannot look at that, but in about 3 months I want to finish 1 of them, so then (I hope) I could find some time to look at this enhancement.

gaearon:
This was mostly fixed in https://github.com/facebook/react-devtools/issues/511.

PR #511, opened 2017-02-15
Breadcrump - adding scroll and max-height

mskec:
Fix for https://github.com/facebook/react-devtools/issues/481
This approach is the simplest. There will be a maximum of 4 breadcrumb rows and scroll will appear if needed.
For testing I used this [branch](https://github.com/mskec/react-devtools/tree/example-breadcrumb-scroll). I'm not sure if it makes sense to commit that example in master.

More complicated approach would be to create top/bottom `SplitPane` similar like there is left/right.

### DIFF
--- a/frontend/Breadcrumb.js
+++ b/frontend/Breadcrumb.js
@@ -72,6 +72,8 @@ var styles = {
     listStyle: 'none',
     padding: 0,
     margin: 0,
+    maxHeight: '80px',
+    overflow: 'scroll',
   },
 
   selected: {


gaearon:
This seems like a good start.
Would you be interested in implementing a smarter version like the one in Chrome?

By default Chrome shows only the innermost tree that fits:

screen shot 2017-02-15 at 16 07 04

If you press "..." it takes you to the very root:

screen shot 2017-02-15 at 16 07 09

However, now that you're at the root, you can navigate by a single page forward and back:

screen shot 2017-02-15 at 16 07 14

gaearon:
I'll get this in for now as an incremental improvement. Thank you!

mskec:
Awesome!
I will give it a try implementing smarter version.

PR #513, opened 2017-02-15
Highlight selected line correctly

### LINKED ISSUE #240: Line Highlighting with Horizontal Scrollbars
In Chrome (45.0) the line highlighting for selected items and hover only extends as far as the props panel on the right. When there are horizontal scrollbars, the highlighting gets cut off.

This happens with react-devtools 0.14.3 (current Chrome Store version) and `master`.
It can also be seen in Chrome & Firefox (41) when using `/shells/plain`.
## Initial view:

image
## Scroll right:

image
## `/shells/plain`:

image

MethodGrab:
Fixes #240.

### DIFF
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -73,7 +73,7 @@ class Node extends React.Component {
 
     if (node.get('nodeType') === 'Wrapper') {
       return (
-        <span>
+        <span style={styles.wrappedNodeContainer}>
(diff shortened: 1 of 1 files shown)


MethodGrab:
@facebook-github-bot CLA is now signed.

gaearon:
Sweeeeet.
Thank you.

Issue #514, opened 2017-02-15
Fix Firefox shell


gaearon:
It turns out that it's actually been broken since updating to React 15.
I need to figure out how to test Firefox plugins to get it fixed.

PR #516, opened 2017-02-15
Fix Firefox addon

gaearon:
This should fix https://github.com/facebook/react-devtools/issues/514.

The update to React 15 broke it, and later when I tested with FF I just assumed it doesn't work locally, but it was actually crashing.


spicyj:
(y)

Issue #472, opened 2017-01-16
Checkbox labels and search box overflow the column


albell:
MacOS 10.12.2, Chrome 55. This doesn't look so hot:

![screen shot 2017-01-16 at 2 52 14 pm](https://cloud.githubusercontent.com/assets/1142025/22001699/8afb9c04-dbfb-11e6-8b79-36e9d0203b97.png)

Checkbox options should probably flow to a single vertical column when narrow. Thanks for a great tool!

gaearon:
Yes, the design there is not very polished.
Would you like to help fix this?

timarney:
I can take a look at this if you like.

gaearon:
Sounds great!

timarney:
Did a quick test - 

![dev-tools](https://cloud.githubusercontent.com/assets/62242/22997015/cd0eb3b4-f39e-11e6-86a4-deca8837345b.png)

I'll see about getting the labels to wrap nicer later.

Is there a way to get the dev tools panel to auto-refresh? 

I seem to have to close it manually and re-open to see the updated view.

gaearon:
Hmm what do you mean by auto-refresh? Which shell are you using?

timarney:
Just getting familiar with things ... I was running the chrome shell should I be using the plain one?

Auto-refresh - I did the webpack watch command from the chrome dir, which updates the code but I wasn't seeing changes until I closed the dev tool panel and than re-opened it.

gaearon:
Plain is simpler to develop against since you just refresh to see the updates.

timarney:
How's this?
![checkboxes](https://cloud.githubusercontent.com/assets/62242/23003791/103ceb7a-f3c1-11e6-901e-2cca3dd72b8a.gif)

kp25:
@timarney that looks better than overflow thing, but I'm thinking about what to do to avoid the text 'Use Regular Expressions' hiding behind.

What about reducing the column width further and replace the labels with some kind of ellipsis? Not sure about how it looks, just a thought.

jaredly:
looks like auto-refer didn't work https://github.com/facebook/react-devtools/pull/519

jaredly:
[fixed by #519]

PR #519, opened 2017-02-16
Fix issue #472: Stack checkboxes in settings pane

helenvholmes:
Before:
8afb9c04-dbfb-11e6-8b79-36e9d0203b97

After:
Items now stack when you resize window
)

Wasn't certain what y'alls wanted to happen here/how important being able to expand the sidebar to 100% was important, so I threw on a `min-width` and a `flex-wrap`.

### DIFF
--- a/frontend/SearchPane.js
+++ b/frontend/SearchPane.js
@@ -167,7 +167,7 @@ var styles = {
   input: {
     flex: 1,
     fontSize: '18px',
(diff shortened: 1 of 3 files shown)


jaredly:
This looks great! Thanks for fixing the jank w/ the search input too

gaearon:
Thaaank you.

PR #521, opened 2017-02-16
Alternative fix for #240

### LINKED ISSUE #520: Sibling tag doesn't start with a newline
Seeing a weird issue on Chrome:

screen shot 2017-02-16 at 00 36 16

One child ends, and then the another child immediately begins without a newline.

I wonder if this is some edge case introduced after https://github.com/facebook/react-devtools/pull/513.

@MethodGrab Do you have any thoughts?

gaearon:
Also fixes #520 which was introduced in #513.
The whole wrapper thing is gross but unavoidable right now.

I also noticed another Fiber-only bug related to that (will file later).

### DIFF
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -73,7 +73,7 @@ class Node extends React.Component {
 
     if (node.get('nodeType') === 'Wrapper') {
       return (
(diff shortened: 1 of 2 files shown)


gaearon:
Nah, that breaks multiple roots.

PR #522, opened 2017-02-16
Fix line highlighting when scrolling horizontally

### LINKED ISSUE #240: Line Highlighting with Horizontal Scrollbars

jaredly:
Fixes #240, replaces #513.

Turns out flexbox in the presence of scrolling is a weird beast.
Having "flex-direction: column" doesn't mix well with filling the width while
horizontally scrolling.
Therefore, we use `flex-direction: row` with a `flex: 1` child, and the child
has `flex-direction: column` and `align-items: stretch`.

### Test Plan:
rebuild in `shells/plain`, get it running (open up `shells/plain/index.html`),
and shrink the tree view's width to be small enough that you get some
horizontal scrolling when things are expanded.

Verify that:
- when all is collapsed & there's not horizontal scrolling, the highlight is
  full-width
- when things are expanded and there's horizontal scrolling, the highlight
  extends all the way to the right of the inner content.

For good measure, go into `tests/example/target.js` and add another react root at the bottom.

image

image

### DIFF
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -298,6 +298,7 @@ var styles = {
   // in case we need them later, and the corresponding divs refernce them. But
   // I could remove them if desired.
   container: {
+    flexShrink: 0,
   },
 
   children: {
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -73,7 +73,7 @@ class Node extends React.Component {
 
     if (node.get('nodeType') === 'Wrapper') {
       return (
-        <span style={styles.wrappedNodeContainer}>
+        <span>
           {children.map(child =>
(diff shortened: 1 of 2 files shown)


gaearon:
This works great. Thank you!

PR #528, opened 2017-02-16
[WIP] Fixes for ReactNative on Fiber

gaearon:
WIP

### DIFF
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -115,15 +115,10 @@ function getData(element: Object): DataType {
   if (typeof element.setNativeProps === 'function') {
     // For editing styles in RN
     updater = {
-      setState() {},
-      forceUpdate() {},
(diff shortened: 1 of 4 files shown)


backend/getData.js
```diff
@@ -112,6 +112,15 @@ function getData(element: Object): DataType {
     }
   }
 
+  if (typeof element.setNativeProps === 'function') {
```
gaearon: Note to self: this depends on how we expose host instances in the end.

It might be that we'll just expose this from RN renderer instead, and it will have to inject `setNativeProps(tag)` just like it does with `resolveStyle()`.

bvaughn:
Closing in favor of a new branch, pushed to facebook/react-devtools so Dom and I can both work on it, with conflicts from master resolved. 😄

PR #533, opened 2017-02-16
Removed duplicated css margin property in plain shell index.html file

tomaszlakomygp:
It's my first (hopefully not last) contribution to react-devtools so it has to be a massive change, right?

Anyway, there was a duplicated `margin` in `shells/plain/index.html` and it's gone now. All is well.

### DIFF
--- a/shells/plain/index.html
+++ b/shells/plain/index.html
@@ -22,7 +22,6 @@
                 left: 0;
                 right: 0;
                 bottom: 0;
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks! Let's remove the first one instead so it stays close to padding?

tomaszlakomygp:
Done 👍

gaearon:
Thanks!

PR #535, opened 2017-02-16
Fixed inconsistencies and typos in some frontend components

tomaszlakomygp:
I've been looking through the codebase in order to get familiar with it and I've noticed some small inconsistencies/typos that can be fixed easily, hence this PR. 

(I realise now that I should have included change from https://github.com/facebook/react-devtools/pull/533 in this PR, let me know if I should combine those two PRs.)

### DIFF
--- a/frontend/Props.js
+++ b/frontend/Props.js
@@ -16,10 +16,7 @@ var PropVal = require('./PropVal');
 class Props extends React.Component {
   props: Object;
   shouldComponentUpdate(nextProps: Object): boolean {
-    if (nextProps === this.props) {
-      return false;
-    }
-    return true;
+    return nextProps !== this.props;
   }
 
   render() {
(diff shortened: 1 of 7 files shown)


frontend/Props.js
```diff
@@ -16,10 +16,7 @@ var PropVal = require('./PropVal');
 class Props extends React.Component {
   props: Object;
   shouldComponentUpdate(nextProps: Object): boolean {
-    if (nextProps === this.props) {
-      return false;
-    }
-    return true;
+    return nextProps !== this.props;
```
gaearon: This was actually a pretty weird comparison. I wonder if @jaredly meant to compare `this.props.props` with `nextProps.props` instead.

tomaszlakomygp: I was a bit confused myself but since I'm new to the codebase I've decided not to change the logic beneath it, only polish the code a little

gaearon: Right. Can you file a new issue to investigate this?

tomaszlakomygp: Sure, I'll create it a bit later

Issue #349, opened 2016-03-19
Sparse arrays display "null" for missing indices


m-kondrateva-parc:
Hi,

Undefined array elements are shown as `null` in `React Dev Tool` chrome extension and it may be misleading when debugging.

Please refer to my test case:

I have array `units` with defined 1,2,3 and 1000 elements. Rest of the elements are `undefined`.
When I check state of Array with console:

``` javascript
$r.props.units
```

I get following:

```
Array[1001]

0: Object
1: Object
2: Object
1000: Object
length: 1001
__proto__: Array[0]
```

Ok, double-checking:

``` javascript
$r.props.units[5]
```

is `undefined`.
With `React Dev Tool` chrome extension I get `null` instead of `undefined`:

![alt text](http://snag.gy/553ob.jpg)

jaredly:
yeah, so we don't currently support sparse arrays.
two options here.
1) don't use sparse arrays :) I find them a bit of an anti-pattern, and your use case would probably be just as well served by using an object here. If you don't want a list of things, but rather a mapping of numbers to values, use an object.
2) change devtools to support sparse arrays :) I imagine the only thing needed would be to switch somewhere from a `for (var i=0; i<arr.length; i++)` to an `arr.forEach(`.

If you want to investigate 2, let me know and I can give you some pointers.

102:
![selection_002](https://cloud.githubusercontent.com/assets/5839225/23033931/238d2198-f4a4-11e6-9eb9-5cbe4a462036.png)

Can't reproduce it: `undefined` items are just omitted

tomaszlakomygp:
I've created a PR for this issue although I'm not entirely sure which one is the "correct" approach - whether the `undefined` items should be omitted or not.

gaearon:
Unlabeling since there's an open PR: https://github.com/facebook/react-devtools/pull/536.

gaearon:
This was fixed in https://github.com/facebook/react-devtools/pull/536.

PR #536, opened 2017-02-16
Added displaying all elements in a sparse array

tomaszlakomygp:
This fixes the issue described in https://github.com/facebook/react-devtools/issues/349

Currently when inspecting a component which has a sparse array passed in as a prop the user will only see the elements that are defined, the rest is omitted.

I've ensured that the inspector will show the sparse array in its entirety (although I'd argue against using sparse arrays at all), see the gif below for reference:

react_dev_tools

### DIFF
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -37,10 +37,11 @@ class DataView extends React.Component {
 
   render() {
     var data = this.props.data;
+    var isArray = Array.isArray(data);
     if (!data) {
       return <div style={styles.missing}>null</div>;
     }
-    var names = Object.keys(data);
+    var names = !isArray ? Object.keys(data) : [...data.keys()];
     if (!this.props.noSort) {
       names.sort(alphanumericSort);
     }
@@ -48,7 +49,7 @@ class DataView extends React.Component {
     if (!names.length) {
       return (
         <div style={styles.empty}>
-          {Array.isArray(data) ? 'Empty array' : 'Empty object'}
+          {isArray ? 'Empty array' : 'Empty object'}
         </div>
       );
     }
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -35,8 +35,8 @@ type DataViewProps = {
 class DataView extends React.Component {
   props: DataViewProps;
 
-  parseSparseArray(arr) {
-    var acc = [];
+  parseSparseArray(arr: Object) {
+    var acc: Array<any> = [];
     var length = arr.length;
     var numOfUndefsInRow = 0;
     var pushUndefValsNumber = () => acc.push('undefined x ' + numOfUndefsInRow);
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -39,17 +39,16 @@ class DataView extends React.Component {
     var acc: Array<any> = [];
     var length = arr.length;
     var numOfUndefsInRow = 0;
-    var pushUndefValsNumber = () => acc.push('undefined x ' + numOfUndefsInRow);
 
     for (var i = 0; i < length; ++i) {
-      if (arr[i] === undefined) {
+      if (!arr.hasOwnProperty(i)) {
         numOfUndefsInRow++;
         if (i === length - 1) {
-          pushUndefValsNumber();
+          acc.push('undefined x ' + numOfUndefsInRow);
         }
       } else {
         if (numOfUndefsInRow) {
-          pushUndefValsNumber();
+          acc.push('undefined x ' + numOfUndefsInRow);
         }
         numOfUndefsInRow = 0;
         acc.push(i);
@@ -66,10 +65,9 @@ class DataView extends React.Component {
 
     var isArray = Array.isArray(data);
     if (isArray) {
-      var itemCounter = 0;
-      data.forEach(() => ++itemCounter);
+        var realItemCount = data.reduce((acc) => {return ++acc});
     }
-    var isSparseArray = isArray && data.length !== itemCounter;
+    var isSparseArray = isArray && data.length !== realItemCount;
     var names = !isSparseArray ? Object.keys(data) : this.parseSparseArray(data);
     if (!this.props.noSort && !isSparseArray) {
       names.sort(alphanumericSort);
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -35,26 +35,30 @@ type DataViewProps = {
 class DataView extends React.Component {
   props: DataViewProps;
 
-  parseSparseArray(arr: Object) {
-    var acc: Array<any> = [];
-    var length = arr.length;
-    var numOfUndefsInRow = 0;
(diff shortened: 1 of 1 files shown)


spicyj:
Chrome does this:

image

Any strong reason we should deviate?

tomaszlakomygp:
I don't have any personal reasons for/against - I saw the issue https://github.com/facebook/react-devtools/issues/349 marked as `good first bug` and decided to give it a go.

One argument could be that in your example is visible that in an array there are in fact elements which are `undefined` whereas in current React Dev Tools the same array looks like this:

screen shot 2017-02-16 at 23 53 09

It might not be obvious to the user that this array is a sparse one and contains `undefined` elements inside of it. Of course I won't insist, just trying to help a bit :)

gaearon:
Thank you for taking this!
How hard would it be to combine these representations in something like

```
undefined × 2
2: 3
undefined × 1
4: 7
5: 9
undefined × 2
```

?

tomaszlakomygp:
Shouldn't be too hard, I'll see what I can do

tomaszlakomygp:
I've pushed a new commit, here's how it looks like in action:

sparse_array_fixed

Let me know if that's satisfactory 😄

gaearon:
We just don't need the ": undefined" there, it's a bit confusing. Also I'd like to show it in italics, like Chrome does, to make it visually distinct.

tomaszlakomygp:
Sure, no problem. I wanted not to introduce too much lines to handle this edge case (after all, sparse arrays are quite rare) but I see your point.

gaearon:
Shouldn't be too much code, it's just another component and an if branch, right?
I'm not sure how it's actually implemented but that's what I'd expect.

tomaszlakomygp:
Comments applied, see below:

screen shot 2017-02-17 at 17 15 14

gaearon:
Very nice, I love it! Can you look at why CI is failing?

tomaszlakomygp:
CI passes now 👍

frontend/DataView/DataView.js
```diff
@@ -35,20 +35,51 @@ type DataViewProps = {
...
   props: DataViewProps;
 
+  parseSparseArray(arr: Object) {
+    var acc: Array<any> = [];
+    var length = arr.length;
+    var numOfUndefsInRow = 0;
+    var pushUndefValsNumber = () => acc.push('undefined x ' + numOfUndefsInRow);
+
+    for (var i = 0; i < length; ++i) {
+      if (arr[i] === undefined) {
```
gaearon: It's a bit trickier that this.
Just because something is undefined, doesn't mean there is a hole in the array.

```
const a = [,,,]; // sparse
const b = [undefined, undefined, undefined]; // not sparse
```

I think the real check here is whether `arr.hasOwnProperty(i)`. But I might be wrong.

frontend/DataView/DataView.js
```diff
@@ -35,20 +35,51 @@ type DataViewProps = {
 class DataView extends React.Component {
   props: DataViewProps;
 
+  parseSparseArray(arr: Object) {
+    var acc: Array<any> = [];
+    var length = arr.length;
+    var numOfUndefsInRow = 0;
+    var pushUndefValsNumber = () => acc.push('undefined x ' + numOfUndefsInRow);
```
gaearon: Let's write this inline instead of an extra closure. Just duplicate this line if you need to.

frontend/DataView/DataView.js
```diff
@@ -35,20 +35,51 @@ type DataViewProps = {
...
     if (!data) {
       return <div style={styles.missing}>null</div>;
     }
-    var names = Object.keys(data);
-    if (!this.props.noSort) {
+
+    var isArray = Array.isArray(data);
+    if (isArray) {
+      var itemCounter = 0;
+      data.forEach(() => ++itemCounter);
```
gaearon: Let's avoid closures in tight loops.

tomaszlakomygp:
FYI: I'll get this PR finished this weekend, it's been a busy week

frontend/DataView/DataView.js
tomaszlakomygp: Looks like you're right 👍

tomaszlakomygp:
@gaearon Comments addressed, take a look again when you'll have the time. Thanks for review 👍

gaearon:
Can you check why CI fails?

tomaszlakomygp:
Done, CI passes now - all is well :)

gaearon:
Thanks! Going to look at this and a few other PRs later next week.

gaearon:
I tried this with Electron version of DevTools and realized this only works in plain shell.

The problem is that the array loses sparseness when it gets converted to JSON, which happens in real extensions (and the Electron version).

I started adding tracking of holes [to `dehydrate.js`](https://github.com/facebook/react-devtools/blob/ca647bcc6d8508ca119db2754c6837ebac20cb42/agent/dehydrate.js#L73-L76) so that we could restore them after receiving JSON but couldn't figure out where to add the "restoration" code that would hydrate this back:

```js
    var holes = [];
    var result = [];
    var lastIndex = -1;
    data.forEach((item, i) => {
      for (var j = lastIndex + 1; j < i; j++) {
        holes.push(j);
      }
      lastIndex = i;
      // $FlowFixMe path is not undefined.
      result[i] = dehydrate(item, cleaned, path.concat([i]), level + 1);
    });
    for (var j = lastIndex + 1; j < data.length; j++) {
      holes.push(j);
    }
    return {
      // ?
      type: 'array',
      value: result,
      holes: holes
    };
```

Overall at this point I’m not sure it’s worth it 😞 . It seems harder than I thought.
@jaredly Maybe you can provide any guidance?

gaearon:
Hmm. I see now. This is a bigger issue: while Chrome/Firefox/plain shells use `postMessage` (and so this seems to work), Electron shell sends data as JSON. It doesn't only affect sparse arrays, but also objects with `undefined` values (those get stripped out). So I guess it’s a separate unrelated issue.

gaearon:
@tomaszlakomygp 

I think this works (I wrote my concern about Electron in https://github.com/facebook/react-devtools/issues/582 but it is unrelated to the issue).

I found the logic a bit hard to read because of the special cases in `DataItem` so I slightly reshuffled the code. I also think it was a bit hard to understand that `name` was both used for keys and for `undefined` labels so I restructured that. Finally, I also tried to avoid extra allocations and array traversals where possible.

Can you please review my changes for mistakes? (I pushed up two commits to your branch.) Thanks!

tomaszlakomygp:
Looks good to me 👍 Thanks for the improvements, much appreciated!

gaearon:
Yay! Thanks for your work on this 👍

tomaszlakomygp:
No problem @gaearon, I'm glad to help a bit since I use react-devtools daily :)

gaearon:
Do you want to take a look at https://github.com/facebook/react-devtools/issues/582? I wonder how to solve that..

tomaszlakomygp:
Sure! I cannot think of anything obvious that can solve that issue but I'm sure we'll figure something out.

Issue #525, opened 2017-02-16
Make empty UI look nicer


gaearon:
Lots of low hanging fruit here:

![screen shot 2017-02-16 at 12 48 17](https://cloud.githubusercontent.com/assets/810438/23021903/3b6fe9e8-f446-11e6-82ae-e2ced7b57a43.png)

Both left and right pane could use some UI love.

gaearon:
Here’s another one:

<img width="383" alt="screen shot 2017-02-16 at 12 54 17" src="https://cloud.githubusercontent.com/assets/810438/23022114/107dabc0-f447-11e6-93e1-774f288f3e44.png">

lukejacksonn:
I'd be happy to take a look at this.. is there any particular style in mind?

gaearon:
Just making it look cleaner would be nice!

I was thinking about adding some paddings and maybe not using monospace font for text like "No search results" because it looks odd there.

lukejacksonn:
Something as simple as this could work..

![screen shot 2017-02-16 at 11 22 27 am](https://cloud.githubusercontent.com/assets/1457604/23030013/682b7a9e-f43a-11e6-9ed8-9f89bb278312.png)

Changes that effect specifically, the components outlined in this issue:
- Adds padding to main container
- Vertically and horizontally centers the span in the sidebar
- Increases size of error icon and fixes "x" character alignment

Changes that effect components outside the scope of this issue:
- Adds padding to the search input
- Adds padding to the top checkbox container
- Puts some space between the checkboxes and label

gaearon:
How do you feel about using the same positioning/styling for "No search results" as for "No selection"?

lukejacksonn:
I gave that a go at first but felt the snap from left aligned main content, to absolutely centered main content was quite jarring. Also, the slight difference in height of the main container and sidebar container, causes the "No search results" and the "No selection" text to not quite align vertically, which feels awkward too. See here:

![screen shot 2017-02-16 at 12 40 01 pm](https://cloud.githubusercontent.com/assets/1457604/23033221/45bc73e0-f445-11e6-8621-f802a29c6cbd.png)

gaearon:
Yea, it's a bit awkward.
Maybe let's keep it where you put it at first, but make fonts consistent?

lukejacksonn:
Here is what identical font styles look like:

```
font-family: Menlo, Consolas, monospace;
font-size: 12px;
```

[...]

PR #538, opened 2017-02-17
Styling Empty UI

lukejacksonn:
In response to https://github.com/facebook/react-devtools/issues/525 specifically addressing the following:

- Absolutely centers the `No selection` text in the right pane.
- Makes the `No search results` text sans-serif and increases font-size/padding.
- Styles `No state/props` text in the right pane.
- Pads slightly the search input and fixes alignment of the cancel button.
- Fixes spacing between items in settings panel to better accommodate wrapping.
- Fixes the alignment of console hint that appears in the details pane header

screen shot 2017-02-16 at 7 30 20 pm
screen shot 2017-02-16 at 7 30 45 pm


gaearon:
I would like to make both the search pane and the top bar height smaller.

The top bar is very rarely used (and IMO not very thought-out UI-wise). So it would be nice to make it as compact as possible.

The search bar is used often but it doesn't have any interactive elements except the text box so it's fine to make it smaller. You can make the font a little bit smaller if it's bumping into the edges too much in it.

lukejacksonn:
Sounds sensible! Is this better?

screen shot 2017-02-17 at 2 15 30 pm

gaearon:
Great job.
This looks *way* better than it used to!

PR #539, opened 2017-02-17
Support specify maximum connection attempt

kwonoj:
This PR expands options for `connectToDevTools()` to support `maxConnectAttempt`, allows to specify number to retry when web socket connection attempt doesn't succeed.

This is for usecases like who doesn't install react devtools extension but use Electron based react devtools application only - those user'll see console log is filled with continuing connection attempt if devtools are not executed. In my case our application itself is Electron-based application not able to use devtools extension due to certain circumstances, would like to selectively use electron standalone devtools but also wanted to avoid repeated connection attempt as needed.

### DIFF
--- a/packages/react-devtools-core/README.md
+++ b/packages/react-devtools-core/README.md
@@ -1,18 +1,18 @@
 # `react-devtools-core`
 
-A standalone React DevTools implementation.  
+A standalone React DevTools implementation.
 
-This is a low-level package.  
+This is a low-level package.
 **If you're looking for the Electron app you can run, use `react-devtools` package instead.**
 
 ## Exports
 
 ## `require('react-devtools-core').connectToDevTools(options)`
 
-This is similar to `require('react-devtools')` in another package but providing more control.  
+This is similar to `require('react-devtools')` in another package but providing more control.
 Unlike `require('react-devtools')`, it doesn't connect immediately, but exports a function.
 
-Run `connectToDevTools()` in the same context as React to set up a connection to DevTools.  
+Run `connectToDevTools()` in the same context as React to set up a connection to DevTools.
 Make sure this runs *before* any `react`, `react-dom`, or `react-native` imports.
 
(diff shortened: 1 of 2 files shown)


packages/react-devtools-core/README.md
```diff
@@ -1,25 +1,26 @@
...
 
-This is a low-level package.  
+This is a low-level package.
 **If you're looking for the Electron app you can run, use `react-devtools` package instead.**
 
 ## Exports
 
 ## `require('react-devtools-core').connectToDevTools(options)`
 
-This is similar to `require('react-devtools')` in another package but providing more control.
```
gaearon: Spaces are significant in Markdown. Here, they cause the line to break. Could you please remove this change?

kwonoj:
Here it is, it's keep growing and eating all logs in terminal & log file both.

image

I'll update PR as suggested to remove spaces.

gaearon:
I see. Could you explain again why the connection is failing?

Would specifying `isAppActive()` satisfy your use case?
If you return `false` there, it will not connect once (and will check again next time).

kwonoj:
Given application is Electron application doesn't have installed react devtools extension, and dev would like to use it executes electron-based devtools standaone separately. So usual dev doesn't running standalone devtools are seeing those connection error.

So suggestion is use isAppActive instead? I just noticed this after bump it up devtools and haven't used importing core until (was using require('react-devtools') simply)

gaearon:
If `isAppActive()` works for your use case (that is, if you know specify in the code whether now is a good time to reconnect or not), then I'd suggest using it. Can you try?

kwonoj:
I gave a few try and found at least it could work way as I intended, so I'll close this PR for now as this doesn't have strong usecases over it. Though I think it'd better to expose those option in readme?

Appreciate for support, @gaearon .

gaearon:
Feel free to send a PR to expose it in the README! I haven't had time yet.

Issue #534, opened 2017-02-16
Add icon and title to standalone Electron app


gaearon:
The [Electron app](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools) currently has a generic title and icon:

![screen shot 2017-02-16 at 20 02 42](https://cloud.githubusercontent.com/assets/810438/23038699/e4d68104-f482-11e6-8d5b-3d4c21c66af5.png)
<img width="126" alt="screen shot 2017-02-16 at 20 02 33" src="https://cloud.githubusercontent.com/assets/810438/23038700/e4d7dd1a-f482-11e6-8609-b2e61b54d073.png">

We should fix it to use our existing extension icon (it's somewhere in this repo) and say "React Developer tools" in the title (and app name in the dock too!)

tomaszlakomygp:
I'm on it! 😄

tomaszlakomygp:
@gaearon Fixing the title/icon for Windows/Linux is trivial, it's just a matter of adding the following in app.js:

```
mainWindow = new BrowserWindow({width: 800, height: 600, icon: path.join(__dirname, 'icons/icon128.png')});
```

The thing is that it doesn't work on MacOS, the only solution I was able to find requires using electron-packager, see here: http://stackoverflow.com/questions/31529772/how-to-set-app-icon-for-electron-atom-shell-app

gaearon:
Sure, separate PRs sound good.

gaearon:
OK, fixing macOS icon wasn't so easy: https://github.com/facebook/react-devtools/pull/558
If somebody has ideas on how to change our bundling workflow to allow it, please chime in!

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #540, opened 2017-02-17
Fixed title & icon for Windows/Linux in Electron standalone app

tomaszlakomygp:
This PR is related to https://github.com/facebook/react-devtools/issues/534

Fixing the electron icon on Windows/Linux is just a matter of adding the first param to `new BrowserWindow` call, unfortunately for MacOS it's an another piece of cake and I'd have to dig in a bit further to solve that.

Anyway, I hope that at least this will be useful 😄


gaearon:
Thanks!

Issue #295, opened 2015-12-11
Persist previous state upon page refresh


rayshan:
Hello, it would be nice if react-devtools can remember user's search term / selection / tree expansion state / etc. upon refreshing the page in browser.

kassens:
This would be a nice addition and doesn't seem super complicated. Contributions welcome :)

jaredly:
Here's how I would do it:
- store that state in the chrome extension's "background page", keyed by the tabid
- on refresh, try to expand, and if it doesn't work (e.g. nodes not rendered yet) try again in a second or so...
- you'd probably also want to make sure to clean up the stored data when the tab closes

khanghoang:
I'd like to give it a try

khanghoang:
I'm facing the problem that I don't know what to store because the id of the node changes when we refresh the page. My first thought was we shall store the path of the current expanded nodes but it seems doesn't work properly in some cases, for example, the list of children. 

cc @gaearon @jaredly

gaearon:
I would probably store displayName and key path. This is as accurate as it gets IMO. I wouldn't rely on any IDs.

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

gaearon:
FWIW the rewrite has this feature. :-)

PR #546, opened 2017-02-17
Feat: Pin component

khanghoang:


gaearon:
Can you provide a gif or vide of how it works?

khanghoang:
@gaearon here it is 

pin

FezVrasta:
Does it work between page reloads?

gaearon:
Not currently but if we implement something like https://github.com/facebook/react-devtools/issues/295 then it should be easy to make it work for pinning too. Would you like to try to tackle https://github.com/facebook/react-devtools/issues/295?

FezVrasta:
I haven't free time for this but it would be awesome to have this PR and #295 work together, this PR alone doesn't look much interesting to me because I would have to redo everything on each reload 😓

gaearon:
This is a good start. Can you make the UI in the panel prettier?

khanghoang:
@gaearon which part that you want me to change and how will it look like?

gaearon:
I think having two identical panels is a bit confusing. I would expect the pinned panel to not have its own footer, I think. I would also expect some sort of header explaining what it is, or another way to separate them.

khanghoang:
@gaearon I made some changes, please take a look if they're what you're looking for :)

container example 2017-02-26 13-35-25

gaearon:
Can we leave the checkmarks at the top bar?

khanghoang:
@gaearon like this?

container example 2017-02-28 11-15-44

If it's not what you want then please let me know what it's gonna look like.

khanghoang:
ping @gaearon

FezVrasta:
I think he means to have the "Trace React Updates", "Highlight Search" etc on the top bar.

khanghoang:
@FezVrasta you're right 👍

khanghoang:
@gaearon @FezVrasta here it is

container example 2017-03-04 00-05-48

FezVrasta:
Just my 2cents, I'm not part of the project, but that "Pinned Components" header is huge...

gaearon:
Can you make it styled more blandly? Perhaps similar to how "No search results" is styled right now, but a little bit larger. Also, it is weird that the main tree is centered. I would expect that after pinned components, there is a horizontal line, and then the normal trees go after them as normal, without centering.

khanghoang:
@gaearon sorry for the late reply. How is about this?
container example 2017-03-16 21-29-23

khanghoang:
@gaearon ping

khanghoang:
@gaearon can you help me review this one?

gaearon:
I'll review it—as I mentioned in other threads, I'm in the middle of a big React change and will jump on other tasks and repos after that's done.

khanghoang:
@gaearon this is just a friendly reminder if you can take a look at this PR when you have time.

gaearon:
I will look at it—as you can see I’ve been merging more urgent bugfixes, and there are still more bugs that I need to fix before cutting a release. This is more experimental stuff, and while I’m excited you worked on this, it’s less urgent and I can only look at it when I have some free time (currently I’m jumping between several large task lists in several projects).

It would be really helpful if you could look at the persistence of the collapsed/expanded tree state in the meantime separately, as that’s a pre-requisite to making this feature truly useful.

khanghoang:
hi @gaearon, I finally figure out a way to persist pinned components, I think we can use the same technique for collapsed/expanded tree state. Take a look at this PR if you have time.

frontend/Store.js
```diff
@@ -128,6 +130,7 @@ class Store extends EventEmitter {
     this._parents = new Map();
     this._nodesByName = new Map();
     this._bridge = bridge;
+    this.pinnedComponents = JSON.parse(localStorage.getItem('foo', JSON.stringify(this.pinnedComponents))) || [];
```
petetnt: Should be `JSON.parse(localStorage.getItem('foo')) || []`

frontend/Store.js
```diff
@@ -564,6 +567,50 @@ class Store extends EventEmitter {
...
+    ];
+    localStorage.setItem('foo', JSON.stringify(this.pinnedComponents));
+    this.emit('unpinComponent');
+  }
+
+
+  getPath(componentId: string): string {
+    const currentNode = this.get(componentId);
+    if (!currentNode) return null;
+    const key = currentNode.get('key') === null ? '' : currentNode.get('key');
```
petetnt: `currentNode.get('key') || ''` might be clearer.

frontend/Store.js
```diff
@@ -564,6 +567,50 @@ class Store extends EventEmitter {
...
+    let path = `${currentNode.get('name')}${key}`;
+    let temp = this.getParent(componentId);
+    while (this.getParent(temp)) {
+      const parentNode = this.get(this.getParent(temp));
+      const k = parentNode.get('key') === null ? '' : parentNode.get('key');
```
petetnt: `parentNode.get('key') || ''` might be clearer.

frontend/SearchPane.js
```diff
@@ -0,0 +1,186 @@
...
+  reload: PropTypes.func,
+  searchText: PropTypes.string,
+  selectFirstSearchResult: PropTypes.func,
+  onChangeSearch: PropTypes.func,
+  placeholderText: PropTypes.string,
+};
+
+var Wrapped = decorate({
+  listeners(props) {
+    return ['searchText', 'placeholderchange'];
```
petetnt: What is placeholderchange? 🤔

frontend/Store.js
petetnt: The key should probably be something else than `foo` 😸

frontend/Store.js
```diff
@@ -103,6 +103,8 @@ class Store extends EventEmitter {
   placeholderText: string;
   preferencesPanelShown: boolean;
   refreshSearch: boolean;
+  // pinned components
```
petetnt: Comment not needed

bvaughn:

PR #548, opened 2017-02-18
Exit edit mode on enter press

pastelsky:
Finish editing when enter is pressed when editing props / state.

react devtools


gaearon:
Yes. Thank you!

Issue #545, opened 2017-02-17
Unclear condition in shouldComponentUpdate in frontend/Props.js


tomaszlakomygp:
This issue is a result of my findings in https://github.com/facebook/react-devtools/pull/535/files, @gaearon suggested to create a separate issue for that

Currently the `shouldComponentUpdate` function in https://github.com/facebook/react-devtools/blob/master/frontend/Props.js#L18 looks a bit uncommon:

```
  shouldComponentUpdate(nextProps: Object): boolean {
    return nextProps !== this.props;
  }
```

gaearon:
This is probably unintentional. Worth verifying and fixing.

ioantm:
There is one more place where this kind of check is done https://github.com/facebook/react-devtools/blob/master/frontend/Node.js#L44 , here `shouldComponentUpdate` probably should be removed.

gaearon:
Fixed via #549.

PR #549, opened 2017-02-18
update comparison to props key on props

kyleshevlin:
The Props component has a key `props` on its `props`. The previous comparison was for all props passed to the component instead of checking strictly for the `props` key passed on the props object. Given that the values on this key are used in the render method, it seemed safe to assume that was the intended comparison.

This resolves issue #545


gaearon:
Yea, this seems good. Thank you!

Issue #288, opened 2015-11-19
Show non-enumerable properties in the right pane


tquetano-r7:
I want to be able to trace back the inline styles I apply to the keys in the object they relate to, so I tried adding a non-enumerable, non-configurable property to my style object that just contains the key, thinking if it was non-enumerable then React's style creator would ignore it. This is true, however the key also does not appear in devtools, basically just disappears once the component is created.

My question is if its possible to surface that hidden property in DevTools? Or perhaps I'm going about it the wrong way, is there another good practice for tracing back the styles for a rendered component to the object you instantiated them with?

jaredly:
neat trick! I've actually also been thinking about how to make inline styles easier to reason about. 
For hidden properties, we currently just do a `for ... in` on the object [as seen here](https://github.com/facebook/react-devtools/blob/master/agent/dehydrate.js#L86). It shouldn't be hard to include non-enumerables, but I wonder if there are any downsides to doing so.
@spicyj are there typically non-enumerables that we'd rather not send up to the devtools?

spicyj:
I think it would be fine to show nonenumerable keys (via Object.getOwnPropertyNames).

tquetano-r7:
Sweet! Do you guys need me to do a PR? It's a fairly minor change, so didn't know if you guys just wanted to churn it out or if you want the audit trail.

jaredly:
go ahead and PR!

gaearon:
Fixed in https://github.com/facebook/react-devtools/pull/376.

gaearon:
Reopening because I reverted #376 in https://github.com/facebook/react-devtools/pull/517.

gaearon:
We can fix this again after https://github.com/facebook/react-devtools/issues/390#issuecomment-280360149 is implemented.

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #551, opened 2017-02-18
Don’t allow expansion of empty arrays or objects

pastelsky:
Previously, we used to allow clicking on empty complex values, only to discover that the value was empty. Now we show it upfront.

screen shot 2017-02-18 at 2 34 11 pm

### DIFF
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -139,8 +139,8 @@ class DataItem extends React.Component {
   isEmpty(value) {
     return (
       (Array.isArray(value) && !value.length) ||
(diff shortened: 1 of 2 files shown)


gaearon:
How does this handle objects with stuff on the prototype but no own properties?

gaearon:
I think we should probably wait for https://github.com/facebook/react-devtools/issues/390 which should pave the way for https://github.com/facebook/react-devtools/issues/288, and then decide what to do with this one.

gaearon:
I’ll close because this seems very low priority, and there are other issues (as explained above) which make this harder than we’d like. Thanks for the PR though!

PR #553, opened 2017-02-18
Enhance the treeview with better selection color and indent guides

pastelsky:
indent


gaearon:
I really like this direction. Can you rebase on master?
I would like to make the "matching" indentation guide a little bit more subtle.

Feel free to tune the right panel too if some recent changes don’t work very well with it.

pastelsky:
Have rebased against master. As for the color for the active indent line, I've used the same shade that chrome dev tools uses (to make the looks consistent). Could make it lighter if you wish to.

gaearon:
What Chrome seems to do differently is it doesn’t actually display all guides all the time. It only shows the guide for the selected item, and the currently hovered item. Can you do the same?

gaearon:
Ping 😄

gaearon:
Unfortunately there are a few issues with the new layout here. For example clicking on the row from the left side no longer selects it, and navigating rows with keys no longer scrolls to them. So I don’t think this matches the requirements. Closing per lack of follow up.

gaearon:
Just wanted to give you a heads up as I did some work on this inspired by your PR.
https://github.com/facebook/react-devtools/pull/618

PR #556, opened 2017-02-18
update example: add deeply nested component; switch docking with Alt+D

alexkuz:
A couple of example updates for easier testing:

* toggle bottom/right DevTools docking with `Alt+D`;
* added deeply nested component.

image


gaearon:
Thanks! Would appreciate a separate PR that documents this somewhere.

PR #557, opened 2017-02-19
fix(getData): fix Stateless Functional Component source code doesn't …

ShiXiaoyYu:
fix Stateless Functional Component source code doesn't show up correctly

### DIFF
--- a/shells/chrome/src/panel.js
+++ b/shells/chrome/src/panel.js
@@ -42,7 +42,7 @@ var config: Props = {
     // the constructor. Otherwise, inspect the render function.
     var code = `Object.getOwnPropertyDescriptor(window.${vbl}.__proto__.__proto__, 'isMounted') &&
       Object.getOwnPropertyDescriptor(window.${vbl}.__proto__.__proto__, 'isMounted').value ?
-        inspect(window.${vbl}.render) : inspect(window.${vbl}.constructor)`;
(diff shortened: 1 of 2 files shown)


ShiXiaoyYu:
react-devtool-stateless

gaearon:
I don't think this is the right fix. You are mutating the instance which we should avoid.
Can we change the "Show Source" function to always use `type` instead of `constructor`?

shells/chrome/src/panel.js
```diff
@@ -42,7 +42,7 @@ var config: Props = {
     // the constructor. Otherwise, inspect the render function.
     var code = `Object.getOwnPropertyDescriptor(window.${vbl}.__proto__.__proto__, 'isMounted') &&
       Object.getOwnPropertyDescriptor(window.${vbl}.__proto__.__proto__, 'isMounted').value ?
-        inspect(window.${vbl}.render) : inspect(window.${vbl}.constructor)`;
+        inspect(window.${vbl}.render) : inspect(window.${vbl}._reactInternalInstance._currentElement.type)`;
```
gaearon: This is a bit gross because we're looking at internals but I don't really have a better alternative. Let's just fix this to work in Fiber: `inspect(typeof window.${vbl}._reactInternalInstance.tag === 'number' ? window.${vbl}._reactInternalInstance.type : window.${vbl}._reactInternalInstance._currentElement.type)`

gaearon:
Closing as review comments have not been addressed.

PR #558, opened 2017-02-20
Fix #534: Add MacOS Electron app icon

### LINKED ISSUE #534: Add icon and title to standalone Electron app

helenvholmes:
screen shot 2017-02-20 at 12 28 31 pm

Adds the `.icns` file necessary to package a Mac Electron app, second part of #534. Also adds a package name and electron dependencies to the `package.json` file—I was running through [this](https://www.christianengvall.se/electron-packager-tutorial/) tutorial which seemed to suggest those changes were necessary for bundling it properly:

> Electron packager looks for a product name in package.json, so lets go ahead and add one. We also need to add what version of electron to package the app with.

 ¯\_(ツ)_/¯


helenvholmes:
Seems like Travis failed on a `package.json` inside a node_module:

```
The command "npm run lint" exited with 0.
6.89s$ flow check
node_modules/electron-packager/test/fixtures/infer-malformed-json/package.json:3
  3: 
     ^ Unexpected end of input
Found 1 error
```

Any pointers on how to fix this? I'm a little out of my depth.

tomaszlakomygp:
There, I fixed it for ya: ¯\\_(ツ)_/¯

Thanks for finishing this issue, I wasn't quite sure how to create a proper `.icns` icon

helenvholmes:
@tomaszlakomygp I had to buy [IconSlate](http://www.kodlian.com/apps/icon-slate) a few years ago so I generated it with that—but [this article](https://developer.apple.com/library/content/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Optimizing/Optimizing.html#//apple_ref/doc/uid/TP40012302-CH7-SW2) suggests using `iconutil` on the command line if you have the Xcode command line tools installed (although I haven't ever done it that way, GUI ftw 😴)

Thank you!

gaearon:
Hey, thank you for this!

I think the `package.json` you meant to edit is `package/react-devtools/package.json`, not the root one. We only use the root one for common devDependencies. Does this make sense?

helenvholmes:
@gaearon 😑 eep, yes, new PR coming

packages/react-devtools/package.json
```diff
@@ -1,7 +1,9 @@
 {
   "name": "react-devtools",
+  "productName": "React Developer Tools",
   "version": "2.0.12",
   "description": "Use react-devtools outside of the browser",
+  "main": "./packages/react-devtools/app.js",
```
gaearon: Is this still the right path? I'm actually thinking we intentionally have a different entry point because we want people to `require('react-devtools')` and get [the client](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools/index.js) running.

helenvholmes: Meaning I should point to `app.js` relative to the directory it's in (which I've updated... whoops) or that I should be pointing this to `index.js`?

gaearon: I think this should stay pointing at `index.js` (which it does by default when the key is absent). Is there a reason why this doesn't work? When you run from the command line [the bin script runs instead](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools/bin.js) anyway.

helenvholmes: Seems like `electron-packager` gets upset when it can't find those window definitions, which is why I pointed it to the `app.js`:

screen shot 2017-02-22 at 3 10 53 pm

That bin script just requires it, though, no idea why `electron-packager` is barfing on it ☹️

gaearon:
What is `electron-packager`? I don't see it being used here. I just run `npm start` in the app directory to run it.

gaearon:
Hmm I think I'm starting to understand it. You are using `electron-packager` to produce a bundled app so that it gets the icon. But `electron-packager` makes certain assumptions about `package.json`. Is that right?

How did you call `electron-packager`?

helenvholmes:
@gaearon Ah, yes, that's exactly it—sorry, I shouldn't have made any assumptions. 

This is the command I've been passing to create the binary, which I got from [here](https://www.christianengvall.se/electron-packager-tutorial/#macos):

`electron-packager . --overwrite --platform=darwin --arch=x64 --icon=icons/icon.icns --prune=true --out=release-build`

Dumps the binary in a new folder.

Um, if this is getting too complicated, I can modify this PR to just include the `icon.icns` file so you can make decisions about how you want to package up the application for Macs when that's closer on the horizon. (I don't even know if releasing this as a binary in say, the App Store is even the goal, or if you imagine developers will be spinning this up on the command line normally.)

gaearon:
So, ideally the plan is that it should still "just work" as an npm command.
Right now we launch Electron like this:

```
npm i -g react-devtools
react-devtools
```

So we want to preserve this behavior. Maybe it's easy to do with precompiled packages—I'm not sure.

Do you think you could search for an example Electron app that ships through npm but has a custom icon? Then we could learn from it.

helenvholmes:
> Do you think you could search for an example Electron app that ships through npm but has a custom icon? Then we could learn from it.

I looked around for one for a bit—I wasn't able to find any that would properly display the icon with an `npm start` (or whatever their project equivalent was 😪)

Asked around: apparently MacOS icons are part of the app bundle metadata, not the executable, so setting the Mac icon from, say, `app.js` might not be possible. (I guess to never say never, maybe there's some private Apple private API fanciness, but I imagine we wouldn't be able to use it here.)

I guess an option (I think what you meant with precompiled packages?) is that `packages/react-devtools` could have a precompiled version of Electron shell with the icon bundle that's committed, instead of `npm install`'d. I think. _I think._ You might have some gaps in knowledge I'm missing there, or be able to make a judgement call if that could be too much maintenance to have a non-Electron Mac app icon.

gaearon:
I see. Welp. I guess we'll live without an icon then. I don't know how to make the Node binary script use an OS X precompiled bundle.

helenvholmes:
I imagine it could be figured out, main issue would be rebundling for every electron release—I imagine that would get annoying

gaearon:
Yea. If you do figure it out, please send a PR. Otherwise I think we'll have to close this one. Sorry 😞 but thanks for investigating it!

jhen0409:
It should can be exec with `React Developer Tools.app/Contents/MacOS/React Developer Tools` binary, but this will add extra ~100mb to an npm package. 😱

gaearon:
Yea, probably won't do it.

xqwzts:
I don't have a mac machine available but here's what I've found, maybe someone can run with it:
I _think_ this can be achieved by changing the icon in the electron app's Info.plist via the [CFBundleIconFile key](https://developer.apple.com/library/prerelease/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102043)

The Info.plist file should be located under `node_modules/electron/dist/Electron.app/Contents/Info.plist`

The plist values can be changed in the terminal using the `defaults write` command like so: [[via](http://stackoverflow.com/questions/26986874/how-do-i-programmatically-change-icon-of-a-windows-executable-and-mac-applicatio)]
```
defaults write /full/path/to/Info.plist CFBundleIconFile 'app/img/my-icon.icns'
```

Alternatively there are npm modules that expose an API for modifying the file like [plist.js](https://github.com/TooTallNate/plist.js)

gaearon:
Hmm. It seems a bit fishy to do something to your own node_modules.

tomaszlakomygp:
@mjaniszew perhaps you could help? You know a thing or two about electron 😄

helenvholmes:
@xqwzts's solution was one I had looked at for this, but I had the same reaction as @gaearon. Would require checking in a `react-devtools` electron shell.

PR #559, opened 2017-02-20
Fix #388 :  Displaying date appropriately

### LINKED ISSUE #388: Show more info for Date objects.

tarang9211:
PR for #388


gaearon:
This is not fixing the problem. You fixed the example to *not* use a date as a prop. Instead we need to fix the DevTools to display dates correctly. Once that is fixed, it's fine to revert any changes to ` test/example`. I only suggested adding a date prop there so that you can reproduce the underlying problem.

tarang9211:
Oh, im confused. Where should I be making the changes?

gaearon:
DevTools uses a bridge to serialize objects and later deserialize them. Deserialization happens in [`agent/dehydrate.js`](https://github.com/facebook/react-devtools/blob/master/agent/dehydrate.js). So I think you might want to add a special case for Dates there, similarly to how we have a special case for Arrays.

It is probably worth introducing a new type (`'date'`) in addition to `'function'`, `'object'`, and `array`. For a serialized representation, you can probably use `date.toJSON()`. Then you'd need to find the places in the DevTools UI that look at the data type before displaying it. Based on #459 (which is much bigger, as it adds support for Iterables, but similar in scope), it is possible you’ll need to touch `frontend/DataView/previewComplex.js` and `frontend/PropVal.js`.

I hope this helps!

gaearon:
(Note I'm not sure myself. I'd start with adding a special case to `dehydrate` for dates, and seeing how it works and what breaks.)

tarang9211:
No worries, I'll dig into this. Should I remove the changes i made above?

gaearon:
You'll want to keep a date prop there so that you can verify your next fixes work. But you should remove `JSON.stringify` in `test` since that will prevent you from testing your feature.

tarang9211:
Got it. Question: why are you referring to it as props when it's in the `this.state`?

gaearon:
Ah, I missed that. I suggested adding it to props initially so that's what I was referring to. It doesn't matter really—the way you added it to state also works for testing.

tarang9211:
@gaearon is that a step in the right direction? I put a console log on line 44 of `dehydrate.js`which was basically `console.log('reached here')` and it never got printed after I create a new to do by opening the `index.html` in `/shells/plains`

agent/dehydrate.js
```diff
@@ -40,6 +40,13 @@ function dehydrate(data: Object, cleaned: Array<Array<string>>, path?: Array<str
       type: 'function',
     };
   }
+  //check if data is of type date. typeof doesn't support Date, so instanceof is being used
+  if (data instanceof Date) {
```
gaearon: I'm not sure this will work across iframes. Let's use `data != null && typeof data == 'object' && Object.prototype.toString.call(data) === '[object Date]'`. See explanation [here](http://stackoverflow.com/questions/643782/how-to-check-whether-an-object-is-a-date#comment2644531_643827). This is what [Lodash does](https://github.com/lodash/lodash/blob/master/isDate.js), too.

agent/dehydrate.js
```diff
@@ -40,6 +40,13 @@ function dehydrate(data: Object, cleaned: Array<Array<string>>, path?: Array<str
       type: 'function',
     };
   }
+  //check if data is of type date. typeof doesn't support Date, so instanceof is being used
```
gaearon: Let's remove this comment because it just duplicates the code.

agent/dehydrate.js
tarang9211: Done. Should I be able to console.log and see where it reaches?

gaearon:
>I put a console log on line 44 of dehydrate.js which was basically console.log('reached here') and it never got printed 

Here's what I did:

screen shot 2017-02-21 at 16 31 54

I am running:

* `./build.sh --watch` in `shells/plain`
* `../../node_modules/.bin/webpack --watch` in `test/example` (actually it's annoying, can you send another PR to add `build.sh` there too?)

In the console, I see both logs:

screen shot 2017-02-21 at 16 33 04

I am also seeing the date in the inspector:

screen shot 2017-02-21 at 16 33 24

So I think you're going in a right direction, and now need to add code to the `frontend` (as I mentioned above) to handle the new `date` type.

tarang9211:
> ../../node_modules/.bin/webpack --watch in test/example (actually it's annoying, can you send another PR to add build.sh there too?)

simply copy the `build.sh` in `test/example` right?

tarang9211:
@gaearon Thanks for the suggestion. I got really confused with all the scripts to start the dev environment which is why I could not verify if the code was working. Just pushed a few changes. also added a random color, if you want a specific one let me know. (is hex ok?)

gaearon:
>simply copy the build.sh in test/example right?

If that works then yes.

tarang9211:
Do you want another PR for that?

gaearon:
YEa

tarang9211:
Sounds good, will do so. Any thoughts on this PR? Now that we're toward the end of this, I can remove the `time` property from the state right?

jaredly:
I think it's fine to leave `time` in the state as an example

tarang9211:
Sounds good! @gaearon Does this look good to you? I'll pull master and create another PR for the `build.sh` file?

gaearon:
I'm a bit busy this week but I should be able to look at this next week.
Yes, please do.

gaearon:
I am still seeing this:

screen shot 2017-02-24 at 10 20 47 pm

Am I doing something wrong? How do you test the change? Can you give me a screenshot?

tarang9211:
I tested this by creating a new to do item and inspecting the todos array.

gaearon:
This is what I’m seeing in the array:

screen shot 2017-02-24 at 10 38 16 pm

tarang9211:
This is exactly what I see. Oh one second, is the `{...}` the problematic part?

gaearon:
Well, the problem is this doesn’t really show Dates specially, it just shows an object instead of them. What we want is to show the *date itself*—this is why we need to modify the frontend to check for this special type, and display it specially. I see you added the code there, but have you verified it actually runs? Maybe there’s some earlier conditions that prevent it from running.

The end goal is to display Dates same way Chrome displays them:

screen shot 2017-02-24 at 10 44 32 pm

tarang9211:
Oh ok. Sorry about that, I'll get to this tomorrow if it's ok. It's 4 15 am where I am. 😄

gaearon:
Yea, no worries. Thank you!

gaearon:
Do you need any help?

tarang9211:
Hi Dan, sorry for the delay. Got caught up in work, looking at this now.

tarang9211:
@gaearon is the `.toJSON()` really needed? that's converting it to a string type.

gaearon:
It's up to you—all I care about is the end result :-). I just suggested one way of producing [a string like this](https://github.com/facebook/react-devtools/pull/559#issuecomment-282426180) out of a date.

tarang9211:
Alright, so I've been able to debug that it doesn't hit any of the checks in `propVal.js` and `previewComplex.js`. When we do this check
```
  var type = data[consts.type];
```
it returns `object`.

Should I modify `consts.js`?

gaearon:
Would editing it give you anything? I presume it returns `object` because that's what `data` says in its type field. The question is, why.

tarang9211:
Hmm. So `data.toJSON()` returns a string. It comes into `previewComplex.js` as this 
```
Object {type: "date", name: "2017-03-02T20:12:15.389Z"}
```

gaearon:
Maybe? I don’t know 😄 . You have more context on this code than I do.

gaearon:
All I’m saying is: try to see how other special types (e.g. array) work, and follow the same principle. I don’t know this code well so following my suggestions verbatim is probably not the best idea.

tarang9211:
Sounds good. 

> try to see how other special types (e.g. array) work, and follow the same principle.

Could you tell me how to debug arrays, objects?

gaearon:
I'm not sure what you're asking. I think `agent/dehydrate.js` has examples of treating arrays.

tarang9211:
Right, I saw that. For example, to test the date formatting I would enter a new todo item in the input box and then view the todos state (and inspect the time property).

Similarly, I was wondering how would I test the object or array? I know it's being handled, but I want to attach a debugger and step through

gaearon:
You can set a breakpoint in those places in `dehydrate.js`. Since todos are already arrays, I'm not sure what else you could need.

gaearon:
I’m closing since this got stale. Thanks for your effort!

PR #560, opened 2017-02-21
Update comparison logic

### LINKED ISSUE #545: Unclear condition in shouldComponentUpdate in frontend/Props.js

nicksp:
Fixes #545

### DIFF
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -40,10 +40,6 @@ class Node extends React.Component {
   props: PropsType;
   static contextTypes: Object;
 
-  shouldComponentUpdate(nextProps: PropsType) {
-    return nextProps !== this.props;
-  }
-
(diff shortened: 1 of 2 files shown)


gaearon:
#549 was there earlier, sorry!

gaearon:
(Maybe we should still do the `Node` one though. Any ideas as to why it was added?)

PR #561, opened 2017-02-21
Validate the source of global messages

Rob--W:
The global "message" event can be activated by pages from a different origin. Blindly accepting such messages is a security risk.

A message channel in React Devtools always have exactly two endpoints. At each side, the target of `postMessage` is also the expected source of the global `message` event. To resolve the security risk, I added a check to the message event handler to reject messages from unexpected sources.

### DIFF
--- a/shells/chrome/src/backend.js
+++ b/shells/chrome/src/backend.js
@@ -26,7 +26,7 @@ setInterval(function() {
 
 window.addEventListener('message', welcome);
 function welcome(evt) {
-  if (evt.data.source !== 'react-devtools-content-script') {
+  if (evt.source !== window || evt.data.source !== 'react-devtools-content-script') {
     return;
   }
 
@@ -40,7 +40,7 @@ function setup(hook) {
   var wall = {
     listen(fn) {
       var listener = evt => {
-        if (evt.data.source !== 'react-devtools-content-script' || !evt.data.payload) {
+        if (evt.source !== window || !evt.data || evt.data.source !== 'react-devtools-content-script' || !evt.data.payload) {
           return;
         }
         fn(evt.data.payload);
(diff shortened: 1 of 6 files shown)


Rob--W:
@gaearon This PR fixes the most severe issue that I identified in my review of your add-on (I will leave the other issues up to you). This issue is not limited to the Firefox add-on, it is also present in your Chrome extension and non-extension (test?) page. I recommend that you publish a new version of the add-on and extension after merging (and testing) this PR.

gaearon:
Wow, thank you so much for this PR!
cc @jaredly

gaearon:
This looks good.

gaearon:
@Rob--W 

I found some bandwidth to look at the PRs in this repo, and it seems like Firefox addon no longer works for me locally. It just gets stuck "Connecting..." as if it never receives the message. The problem bisects to this commit. Do you have any idea as to what might be causing this issue?

Rob--W:
@gaearon If this PR is the culprit, then I suggest to add logging (e.g. `console.error`) to the if-then-return-blocks that were modified by this PR. That will tell you when a previously accepted message becamse rejected.

gaearon:
Apparently it looks like `window` is a `Sandbox` when loading addons in Debug mode. I’m not sure if this is intentional? The issue is that `evt.source` is *a* `Window` but it’s not *the* `window`.

Rob--W:
Try changing `window` to `document.defaultView`. Does that fix the issue?

gaearon:
Yes, this works. Do you think it is reasonable to do this in all environments, or does it seem Firefox-specific to you?

gaearon:
I’d appreciate if you could take a quick glance at https://github.com/facebook/react-devtools/pull/609.

Rob--W:
It is reasonable to do it in all environments. It shouldn't matter either way, but in Firefox's Add-on SDK, the "window" global is not really the Window.

gaearon:
Thanks. I’ll keep it localized to Firefox then because I guess it would be confusing to people why we use `window` in some places and `defaultView` in other places in the same file. Keeping it confined to FF is more explicit and easily blames.

PR #565, opened 2017-02-24
fixes #306 - Detect when file:// access is not enabled

### LINKED ISSUE #306: Detect when file:// access is not enabled
We can check in code if the plugin has access to `file://` URLs and warn if not. We could implement a warning that people have to manually check the box on the `chrome://extensions` page.

see #172

https://developer.chrome.com/extensions/extension#method-isAllowedFileSchemeAccess

Tushkiz:
- shows devtools tab always
- warns if 'file URLs' permission is not enabled


gaearon:
Do you know whether this problem exists in Firefox or not?

Tushkiz:
It does not exist in Firefox

Tushkiz:
This is how it will look.

devtools-warning-chrome

Tushkiz:
amigos, any feedback on this?

shells/chrome/src/main.js
```diff
@@ -21,9 +21,6 @@ function createPanelIfReactLoaded() {
   chrome.devtools.inspectedWindow.eval(`!!(
     Object.keys(window.__REACT_DEVTOOLS_GLOBAL_HOOK__._renderers).length || window.React
   )`, function(pageHasReact, err) {
-    if (!pageHasReact || panelCreated) {
```
suchipi: Why do you return early here? With this change in place, it looks like the devtools will load the React tab on pages where React isn't present. This could be undesirable.

Tushkiz: There are two cases as of now when the devtools does not display React tab
1. If React is not present
2. If React is present but the devtools extension do not have permission to access 'file://' urls

I agree that it is undesirable to show React tab for 1st case,
but for 2nd case (which behaves same as 1st, when there are no permissions) we have no clue why the tab was not displayed

That is why with this change, devtools will always display React tab.
It will show a message "**Unable to find React…**" for 1st case,
and a more specific message "**Please enable 'Allow access to file URLs' permission for the devtools extension**" for 2nd case

suchipi: Gotcha, that makes sense now. I think this change in behavior is fine, then.

frontend/Panel.js
```diff
@@ -310,6 +324,12 @@ var styles = {
     padding: 30,
     flex: 1,
   },
+  errorMessage: {
+    color: '#f38b00',
```
gaearon: I don't think we use any color in other places. Let's keep the style consistent?

shells/chrome/src/main.js
gaearon: Do we need to leave the `if (panelCreated)` check? Was it guarding against some sort of race condition?

frontend/Panel.js
```diff
@@ -234,7 +239,16 @@ class Panel extends React.Component {
       );
     }
     if (!this.state.isReact) {
-      return <div style={styles.loading}><h2>Looking for React…</h2></div>;
+      return (
+        <div style={styles.loading}>
+          <h2 style={styles.errorMessage}>Unable to find React…</h2>
```
gaearon: This does not necessarily mean it’s unable to find React. As you can see it could also mean it is “looking for React” (doesn’t it always take a while to connect?) Especially with React Native.

We need to avoid showing a strongly negative message when we still might get a connection. Can you distinguish loading state from failed state somehow? Or maybe reword to be more ambiguous and imply it takes a few seconds to load if React is there.

shells/chrome/src/main.js
Tushkiz: Yes, it is guarding against [`onNavigated`](https://github.com/Tushkiz/react-devtools/blob/195ecd46e2fb1fbbf1bf7afd392f8f53c06fd2ea/shells/chrome/src/main.js#L46-L48) and [`setInterval`](https://github.com/Tushkiz/react-devtools/blob/195ecd46e2fb1fbbf1bf7afd392f8f53c06fd2ea/shells/chrome/src/main.js#L52-L54) callbacks

frontend/Panel.js
Tushkiz: oh I see, no colors.

frontend/Panel.js
Tushkiz: I thought the code [here](https://github.com/Tushkiz/react-devtools/blob/195ecd46e2fb1fbbf1bf7afd392f8f53c06fd2ea/frontend/Panel.js#L209-L222) could distinguish loading vs failed state, but it seems that the failed state isn't actually 'failed'.

How about we reword the messages as
- Connecting to React…
- Looking for React…
  - Please check if the devtools extension has 'Allow access to file URLs' permission

gaearon: Yea, it's a though one. I'll try to look at other solutions as this can potentially get more confusing.

Tushkiz: yeah, do let me know.

frontend/Panel.js
```diff
@@ -234,7 +239,16 @@ class Panel extends React.Component {
       );
     }
     if (!this.state.isReact) {
-      return <div style={styles.loading}><h2>Looking for React…</h2></div>;
+      return (
+        <div style={styles.loading}>
+          <h2>Looking for React…</h2>
+          {showFileUrlWarningMessage &&
+            <div style={styles.troubleshootingHint}>
+              Please check if the devtools extension has 'Allow access to file URLs' permission
```
gaearon: Can we change this to [this wording](https://github.com/facebook/react-devtools/blob/60c3e1bfa3ce60a22f1de1f3d8ea2beda3df2a2e/shells/chrome/popups/not-detected.html#L6) and somehow open that link in a new tab if it's clicked?

Tushkiz: I have changed the message, please check.

Tushkiz:
I completely forgot about this PR. Btw I have updated it with the latest master, @gaearon can you look into this (or suggest on the confusing bits)

bvaughn:

PR #568, opened 2017-02-25
Wrap source file location for JSX

### LINKED ISSUE #567: Wrap source file location for JSX
If you run an app with `babel-plugin-transform-react-jsx-source` plugin (which is enabled by default in Create React App in development mode), you'll see source locations in the bottom right pane.

However currently, if the location is long, it causes the whole pane to be scrollable horizontally:

screen shot 2017-02-24 at 10 24 04 pm

We should instead fix it to have the same width as the right pane, and wrap accordingly.

rubenmoya:
Fixes #567.

Here's what it looks like now:

screen shot 2017-02-25 at 14 07 58

Feedback is welcome :)

### DIFF
--- a/frontend/PropState.js
+++ b/frontend/PropState.js
@@ -37,12 +37,10 @@ class PropState extends React.Component {
     }
     return (
       <div style={styles.source}>
-        <div style={styles.sourceName}>
-          {source.fileName}
-        </div>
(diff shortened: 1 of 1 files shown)


gaearon:
Looks great, thanks!

Issue #572, opened 2017-03-02
The value of `Infinity` displays as `null`.


hyperh:
The issue with the React Chrome devtools and Redux Chrome devtools, they both display `Infinity` as `null`. But when I do a `console.log(...)` of my prop value, I get `Infinity`.

Looks like this might be the issue: http://stackoverflow.com/questions/16644597/json-stringify-converting-infinity-to-null

Or do you just suggest writing a custom transform?

tomaszlakomygp:
I'll try to get this fixed, shouldn't be too hard.

tomaszlakomygp:
Done, PR opened.

gaearon:
Fixed in #576, thanks @tomaszlakomygp!

tomaszlakomygp:
No worries @gaearon !

PR #576, opened 2017-03-04
Ensured that Inifnity value will be properly displayed in react dev tools

tomaszlakomygp:
This fixes the issue reported in https://github.com/facebook/react-devtools/issues/572, which is caused by the fact that `JSON.stringify(Infinity)` returns `null`.

### DIFF
--- a/frontend/DataView/Simple.js
+++ b/frontend/DataView/Simple.js
@@ -200,6 +200,8 @@ function textToValue(txt) {
 function valueToText(value) {
   if (value === undefined) {
     return 'undefined';
+  } else if (value === Infinity) {
+    return 'Infinity';
   }
   return JSON.stringify(value);
 }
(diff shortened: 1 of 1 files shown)


tomaszlakomygp:
CI failed when installing npm modules, unfortunately I cannot restart it manually.

frontend/DataView/Simple.js
```diff
@@ -200,6 +200,8 @@ function textToValue(txt) {
 function valueToText(value) {
   if (value === undefined) {
     return 'undefined';
+  } else if (value === Infinity) {
```
spicyj: Can we do `typeof value === 'number'` instead and cast to string? Then it will automatically cover `-Infinity` and `NaN` as well. (Though `-0` will still come out as `0`…)

tomaszlakomygp: Good idea, and I wouldn't bother too much about `-0` since one can argue that it's exactly equal to `0` :)

tomaszlakomygp:
@gaearon Changes applied

gaearon:
Thanks!

PR #577, opened 2017-03-04
Ensured that ellipsis will be added to very long prop names

tomaszlakomygp:
I've noticed a bit of dead code while going through the codebase - an ellipsis was supposed to be added at the end of very long prop names, but the `name` variable created as a result of

```
     var name = this.props.name;
     if (name.length > 50) {
       name = name.slice(0, 50) + '…';
     }
```

was never used anywhere.

Before:
screen shot 2017-03-04 at 17 28 57

After:
screen shot 2017-03-04 at 17 30 15


tomaszlakomygp:
CI fails when executing `lerna bootstrap` command, unfortunately I cannot do much about that :/

gaearon:
Looks good, thanks!

Issue #579, opened 2017-03-06
Mapping a function returning JSX to a list containing 0 returns a weird element in devtools.


szhongren:
Not sure what this is, have only seen it in the React devtools extension on Firefox 51.0.1 (64-bit). I'm using devtools 2.0.12, webpack 2.2.1 and babel-core 6.23.1.

Code [here](https://jsfiddle.net/p16qux88/6/). HTML of the actual page is fine, but in the devtools the

    nums.map(
        (number) =>
            <li key={number.toString()}>
                {number}
            </li>)

turns into

    <li key="-1">-1</li>
    <li key="0"/>0
    <li key="1">1</li>

I'd be happy to put up the full repo if it can't be reproduced.

gaearon:
This is likely a bug where it thinks `0` is an empty child. Probably some check for falsiness in a component that should've been checking for `null` or `undefined` explicitly instead. Would you like to send a PR to fix this?

szhongren:
So I've checked it and it's there in Chrome too, is there something else I need to change to build the extension for debugging? I changed `babel` to `babel-loader` in the `webpack.backend.js` and `webpack.config.js`. 

However, `webpack --config .\webpack.backend.js` gives me a `callback is not a function` error. This is for the chrome extension on Windows.

gaearon:
You should be able to follow the instructions here:

https://github.com/facebook/react-devtools/tree/master/shells/plain

gaearon:
This was fixed by @szhongren in https://github.com/facebook/react-devtools/pull/585!

szhongren:
First PR! Whoop.

PR #585, opened 2017-03-09
Issue #579.

szhongren:
Small issue with the display where if a React element's inner value was
0 it did not close properly in the React devtools. Let me know if there's something else I should do, first PR.


gaearon:
This looks great. I made a few small tweaks to make it nicer. Thanks!

PR #588, opened 2017-03-19
[Electron] Support click `Show Source` to open file in editor

jhen0409:
This PR made click `Show Source` to open file in editor in Electron part, if we used [`babel-plugin-transform-react-jsx-source`](https://www.npmjs.com/package/babel-plugin-transform-react-jsx-source)):

> Example for RN

* Add `launchEditor`, it's mirror from [react-native packager](https://github.com/facebook/react-native/blob/master/local-cli/server/util/launchEditor.js), have changes:
  - Remove `projectRoots`, we shouldn't get it here
  - Clear logs
* Add showComponentSource prop for open file in editor

### DIFF
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -36,7 +36,7 @@ export type Props = {
   reload?: () => void,
 
   // optionals
-  showComponentSource?: () => void,
+  showComponentSource?: (vbl: string, source?: Object) => void,
   reloadSubscribe?: (reloadFn: () => void) => () => void,
   showAttrSource?: (path: Array<string>) => void,
   executeFn?: (path: Array<string>) => void,
@@ -146,14 +146,14 @@ class Panel extends React.Component {
     this.props.showComponentSource(vbl || '$r');
   }
 
-  viewSource(id: string) {
+  viewSource(id: string, node?: Object) {
     if (!this._bridge) {
       return;
     }
     this._bridge.send('putSelectedInstance', id);
     setTimeout(() => {
       invariant(this.props.showComponentSource, 'cannot view source if props.showComponentSource is not supplied');
-      this.props.showComponentSource('__REACT_DEVTOOLS_GLOBAL_HOOK__.$inst');
+      this.props.showComponentSource('__REACT_DEVTOOLS_GLOBAL_HOOK__.$inst', node && node.get('source'));
     }, 100);
   }
 
@@ -265,7 +265,7 @@ class Panel extends React.Component {
             return [this.props.showComponentSource && node.get('nodeType') === 'Composite' && {
               key: 'showSource',
               title: 'Show Source',
-              action: () => this.viewSource(id),
+              action: () => this.viewSource(id, node),
             }, this.props.selectElement && this._store.capabilities.dom && {
               key: 'showInElementsPane',
               title: 'Show in Elements Pane',
--- a/packages/react-devtools-core/package.json
+++ b/packages/react-devtools-core/package.json
@@ -24,6 +24,7 @@
   "author": "Jared Forsyth",
   "license": "BSD-3-Clause",
   "dependencies": {
+    "shell-quote": "^1.6.1",
     "ws": "^2.0.3"
   },
   "devDependencies": {
--- a/packages/react-devtools-core/src/launchEditor.js
+++ b/packages/react-devtools-core/src/launchEditor.js
@@ -1,0 +1,150 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+'use strict';
+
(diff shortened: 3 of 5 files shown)


packages/react-devtools-core/src/backend.js
```diff
@@ -15,6 +15,7 @@ type ConnectOptions = {
   port?: number,
   resolveRNStyle?: (style: number) => ?Object,
   isAppActive?: () => boolean,
+  openFileInEditor?: () => {},
```
suchipi: This type doesn't look right; [openFileInEditor](https://github.com/facebook/react-native/blob/master/Libraries/Core/Devtools/openFileInEditor.js) has no return value. It should probably be `() => void` instead.

plugins/ReactNativeOpenFile/setupBackend.js
```diff
@@ -0,0 +1,34 @@
...
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+'use strict';
+
+import type Bridge from '../../agent/Bridge';
+
+module.exports = function setupRNOpenFile(bridge: Bridge, openFileInEditor: () => {}) {
```
suchipi: The type should be `() => void` here, too.

packages/react-devtools-core/src/backend.js
jhen0409: @suchipi you're right, thank you! 👍

plugins/ReactNativeOpenFile/setupBackend.js
```diff
@@ -0,0 +1,34 @@
...
+      (prev, curr) => prev ? prev[curr] : null,
+      window
+    );
+    if (!el ||
+      !el._reactInternalInstance ||
+      !el._reactInternalInstance._currentElement
+    ) {
+      return;
+    }
+    var source = el._reactInternalInstance._currentElement._source;
```
gaearon: This code is coupled to React internal representation (which is changing as we speak). Can you rewrite it so that it uses the “data” (see `getData()`) just like the frontend code? This way this won’t break when we switch to Fiber.

jhen0409: Got it. I rewrote it with `getData` and `getDataFiber`, I refer to [this implementation](https://github.com/facebook/react/blob/4b2eac3de7e1dbf5c2dd742fd9989974a83972cb/src/shared/utils/getComponentName.js) for check Fiber.

plugins/ReactNativeOpenFile/setupBackend.js
```diff
@@ -0,0 +1,41 @@
...
+import type Agent from '../../agent/Agent';
+
+module.exports = function setupRNOpenFile(bridge: Bridge, agent: Agent, openFileInEditor: () => void) {
+  bridge.on('rn-open-file', val => {
+    // '__REACT_DEVTOOLS_GLOBAL_HOOK__.$inst'
+    var el = val.split('.').reduce(
+      (prev, curr) => prev ? prev[curr] : null,
+      window
+    );
+    var inst = el._reactInternalInstance;
```
gaearon: Sorry I wasn't clear, I didn't mean importing `getData` itself. I meant accessing the `node` data structure (that is generated by `getData` when renderer attaches). You are still accessing the "opaque nodes" (`el`) which is a problem (they're implementation details).

Agent has access to "data" in `this.elementData`. However I don't think it's meant to be a public field.

I would look at `showComponentSource` in `Panel` instead. This looks like related code (not sure if it's used though). Perhaps this is a better place to implement it, and somehow inject an implementation there?

jhen0409: > I would look at showComponentSource in Panel instead.

I tried `node.get('source')` just worked in `Panel`, obviously I did something unnecessarily. 😂

Also, I think it would be great if we just do [`launchEditor`](https://github.com/facebook/react-native/blob/master/local-cli/server/util/launchEditor.js) in Electron part, this will don't need go to RN process (call openFileInEditor) and support non-RN platforms, what do you think?

gaearon: >I tried node.get('source') just worked in Panel, obviously I did something unnecessarily. 😂

Sounds awesome!

>I think it would be great if we just do launchEditor in Electron part, this will don't need go to RN process (call openFileInEditor) and support non-RN platforms, what do you think?

I don’t fully understand but if it’s better, go for it!

jhen0409:
@gaearon I changed this PR as https://github.com/facebook/react-devtools/pull/588#discussion_r112542394.

* Add `launchEditor`, it's mirror from [react-native packager](https://github.com/facebook/react-native/blob/master/local-cli/server/util/launchEditor.js), have changes:
  - Remove `projectRoots`, we shouldn't get it here
  - Clear logs
* Add showComponentSource prop for open file in editor

Due to I removed `projectRoots` so I guess some editors cannot open file to the correct workspace, but currently I can sure it working on Atom.

gaearon:
This looks awesome. I love it. Thanks!

gaearon:
Please see also: #625.

PR #589, opened 2017-03-20
SearchPane.js: press 'Enter' for keyboard nav

bmalehorn:
Before, pressing 'Enter' on the search pane would highlight the first
search result, but wouldn't do anything else. You had to click on the
results in order to navigate them with the arrow keys or hjkl.

Now, pressing 'Enter' on the search pane will select the results,
allowing keyboard navigation without any clicking involved.

----

gif example:

### DIFF
--- a/frontend/SearchPane.js
+++ b/frontend/SearchPane.js
@@ -82,6 +82,8 @@ class SearchPane extends React.Component {
 
   onKeyDown(key) {
     if (key === 'Enter' && this.input) {
+      // switch focus to tree view
+      this.input.blur();
       this.props.selectFirstSearchResult();
     }
   }


bmalehorn:
Thoughts on this CR? Anything I should change?

suchipi:
As written, the `blur` call looks like it would not be a big deal to remove, so there is a chance this functionality could be broken by someone who doesn't know better. It might be good to add a comment saying "blur the input so that the tree view can receive focus".

bmalehorn:
@suchipi Thanks for the feedback, I've added a comment explaining the blur.

gaearon:
Looks great. Thanks!

PR #590, opened 2017-03-20
[React Native] Add box inspect to NativeStyler

jhen0409:
In React Native, it have the built-in inspector can show the component width / height / margin / padding information:

The built-in inspector contents will hidden when we use `react-devtools`, this PR made the box inspector showing in React Native Style Editor:

### DIFF
--- a/packages/react-devtools-core/src/backend.js
+++ b/packages/react-devtools-core/src/backend.js
@@ -14,6 +14,7 @@ type ConnectOptions = {
   host?: string,
   port?: number,
   resolveRNStyle?: (style: number) => ?Object,
+  resolveBoxStyle?: (name: string, style: Object) => ?Object,
   isAppActive?: () => boolean,
 };
 
@@ -47,6 +48,7 @@ function connectToDevTools(options: ?ConnectOptions) {
     host = 'localhost',
     port = 8097,
     resolveRNStyle = null,
+    resolveBoxStyle = null,
     isAppActive = () => true,
   } = options || {};
 
@@ -81,7 +83,7 @@ function connectToDevTools(options: ?ConnectOptions) {
         ws.send(JSON.stringify(data));
       },
     };
-    setupBackend(wall, resolveRNStyle);
+    setupBackend(wall, resolveRNStyle, resolveBoxStyle);
   };
 
   var hasClosed = false;
@@ -123,7 +125,7 @@ function connectToDevTools(options: ?ConnectOptions) {
   }
 }
 
-function setupBackend(wall, resolveRNStyle) {
+function setupBackend(wall, resolveRNStyle, resolveBoxStyle) {
   wall.onClose(() => {
     if (agent) {
       agent.emit('shutdown');
@@ -142,7 +144,7 @@ function setupBackend(wall, resolveRNStyle) {
   agent.addBridge(bridge);
 
   if (resolveRNStyle) {
-    setupRNStyle(bridge, agent, resolveRNStyle);
+    setupRNStyle(bridge, agent, resolveRNStyle, resolveBoxStyle);
   }
 
   setupRelay(bridge, agent, window.__REACT_DEVTOOLS_GLOBAL_HOOK__);
--- a/packages/react-devtools-core/src/backend.js
+++ b/packages/react-devtools-core/src/backend.js
@@ -14,7 +14,6 @@ type ConnectOptions = {
   host?: string,
   port?: number,
   resolveRNStyle?: (style: number) => ?Object,
-  resolveBoxStyle?: (name: string, style: Object) => ?Object,
   isAppActive?: () => boolean,
 };
 
@@ -48,7 +47,6 @@ function connectToDevTools(options: ?ConnectOptions) {
     host = 'localhost',
     port = 8097,
     resolveRNStyle = null,
-    resolveBoxStyle = null,
     isAppActive = () => true,
   } = options || {};
 
@@ -83,7 +81,7 @@ function connectToDevTools(options: ?ConnectOptions) {
         ws.send(JSON.stringify(data));
       },
     };
-    setupBackend(wall, resolveRNStyle, resolveBoxStyle);
+    setupBackend(wall, resolveRNStyle);
   };
 
   var hasClosed = false;
@@ -125,7 +123,7 @@ function connectToDevTools(options: ?ConnectOptions) {
   }
 }
 
-function setupBackend(wall, resolveRNStyle, resolveBoxStyle) {
+function setupBackend(wall, resolveRNStyle) {
   wall.onClose(() => {
     if (agent) {
       agent.emit('shutdown');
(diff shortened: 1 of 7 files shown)


jaredly:
coool I've definitely wanted this

plugins/ReactNativeStyle/BoxInspector.js
```diff
@@ -0,0 +1,110 @@
...
+    top: number,
+    left: number,
+    right: number,
+    bottom: number,
+  },
+  children: React$Element,
+};
+
+var Box = (props: BoxProps) => {
+  var box = props.box;
```
suchipi: It's kinda weird to have a `Box` that takes a prop called `box`. It seems like there is a name for a domain concept here that wants to be found. Maybe creating a type for `{ top: number, left: number, right: number, bottom: number }` will help us find a clearer name to use for this prop.

Alternatively, perhaps `top`, `left`, `right`, and `bottom` could be made top-level props on the element itself:
```js
<Box top={4} left={5} right={6} bottom={7} />
```

plugins/ReactNativeStyle/ReactNativeStyle.js
```diff
@@ -40,53 +48,87 @@ class NativeStyler extends React.Component {
...
+    this.setState(this.state.style);
   }
 
   _handleStyleRename(oldName: string, newName: string, val: string | number) {
     var style = shallowClone(this.state.style);
     delete style[oldName];
     style[newName] = val;
     this.props.bridge.send('rn-style:rename', {id: this.props.id, oldName, newName, val});
-    this.setState({style});
+    this.setState(this.state.style);
```
suchipi: The above comment applies to this line as well (same situation).

plugins/ReactNativeStyle/ReactNativeStyle.js
```diff
@@ -40,53 +48,87 @@ class NativeStyler extends React.Component {
...
     super(props);
-    this.state = {style: null};
+    this.state = {style: null, measureLayout: null};
   }
 
   componentWillMount() {
-    this.props.bridge.call('rn-style:get', this.props.id, style => {
-      this.setState({style});
-    });
+    (this:any)._styleGet = this._styleGet.bind(this);
```
jhen0409: Just avoid `Property not found` error of Flow, but I should define `_styleGet: () => void;` instead.

gaearon:
Do you think there is any harm in just inlining `resolveBoxStyle`? It doesn’t seem to has anything RN-specific per se (correct me if I’m wrong!) I’d like to avoid expanding API surface if we can just get by with some duplication.

jhen0409:
Ahh, yeah it haven't anything RN-specific, mirror it will be more cleaner and support current RN version.

I pushed new commits.

packages/react-devtools-core/src/backend.js
```diff
@@ -138,6 +138,7 @@ function setupBackend(wall, resolveRNStyle) {
   var bridge = new Bridge(wall);
   var agent = new Agent(window, {
     rnStyle: !!resolveRNStyle,
+    rnStyleMeasure: !!resolveRNStyle,
```
gaearon: We don't need this anymore, do we? Seems like it would always be supported.

jhen0409: I added [`measureSupport`](https://github.com/jhen0409/react-devtools/blob/4a332c2e9becadb3d2774f5971b09f836b88ec4b/plugins/ReactNativeStyle/ReactNativeStyle.js#L57) prop in NativeStyler (from [`store.capabilities.rnStyleMeasure`](https://github.com/jhen0409/react-devtools/blob/4a332c2e9becadb3d2774f5971b09f836b88ec4b/frontend/Panel.js#L242)) to backward compatibility for [the backend vendor](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools-core/vendor).

plugins/ReactNativeStyle/setupBackend.js
```diff
@@ -22,15 +28,57 @@ module.exports = function setupRNStyle(bridge: Bridge, agent: Agent, resolveRNSt
...
+    return;
+  }
+
+  instance.measure((x, y, width, height, left, top) => {
+    var margin = (style && resolveBoxStyle('margin', style)) || blank;
+    var padding = (style && resolveBoxStyle('padding', style)) || blank;
+    bridge.send('rn-style:measure', {
+      id, style,
+      measureLayout: {
+        x, y,
```
gaearon: Please format each property on next line (above too).

packages/react-devtools-core/src/backend.js
gaearon: Can you explain why you can’t call `measureStyle` directly in the frontend where you display it instead of going through the bridge? (I’m not sure I understand this part of React DevTools so bear with me.) It would be cool if it “just worked” with old RN versions too.

jhen0409: The [`instance.measure`](https://github.com/jhen0409/react-devtools/blob/d4d9ea7cba2c9f47427007d612455d88863d687e/plugins/ReactNativeStyle/setupBackend.js#L67) is from RN [`UIManager.measure`](https://github.com/facebook/react-native/blob/bc1ea548d0017f131c36a30ce06bf4d512cb2f8c/React/Modules/RCTUIManager.m#L1215), it's native method.

gaearon: OK, this makes sense. You can’t call it because it lives in the RN process.

plugins/ReactNativeStyle/ReactNativeStyle.js
```diff
@@ -25,38 +26,69 @@ type Props = {
   // TODO: typecheck bridge interface
   bridge: any;
   id: any;
+  measureSupport: boolean;
```
gaearon: Let's call this `supportsMeasure`.

plugins/ReactNativeStyle/ReactNativeStyle.js
```diff
@@ -25,38 +26,69 @@ type Props = {
...
 };
 
 type DefaultProps = {};
 
 type State = {
   style: ?Object;
+  measureLayout: ?Object;
```
gaearon: Let's call this `measuredLayout`

plugins/ReactNativeStyle/ReactNativeStyle.js
```diff
@@ -25,38 +26,69 @@ type Props = {
...
+};
+
+type StyleResult = {
+  style: Object;
+  measureLayout: ?Object;
```
gaearon: Let's call this `measuredLayout`

plugins/ReactNativeStyle/setupBackend.js
```diff
@@ -22,15 +28,61 @@ module.exports = function setupRNStyle(bridge: Bridge, agent: Agent, resolveRNSt
...
+    bridge.send('rn-style:measure', {style});
+    return;
+  }
+
+  instance.measure((x, y, width, height, left, top) => {
+    var margin = (style && resolveBoxStyle('margin', style)) || blank;
+    var padding = (style && resolveBoxStyle('padding', style)) || blank;
+    bridge.send('rn-style:measure', {
+      style,
+      measureLayout: {
```
gaearon: `measuredLayout`

plugins/ReactNativeStyle/ReactNativeStyle.js
jhen0409: @gaearon done. :)

jhen0409:
It seems like have an async change in `rename` and `set`, so `measureStyle` will get old layout (apart from `margin` an `padding`), just set `setTimeout` for `measureStyle` it works fine.

gaearon:
Seems to work well. Thank you!

PR #593, opened 2017-03-22
Allow state that is boolean to be toggled

joecritch:
Folks! I've added a checkbox for boolean state, e.g.

👍 Saves having to manually type "true/false"

screen shot 2017-03-22 at 21 17 30

It's my first PR on this repo, so nit away.

### DIFF
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -181,6 +181,10 @@ class DataItem extends React.Component {
     });
   }
 
+  toggleBooleanValue(e) {
+    this.context.onChange(this.props.path, e.target.checked);
+  }
+
   render() {
     var data = this.props.value;
     var otype = typeof data;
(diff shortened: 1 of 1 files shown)


frontend/DataView/DataView.js
```diff
@@ -306,6 +323,13 @@ var styles = {
     top: 4,
   },
 
+  toggler: {
+    marginLeft: -12,
```
jaredly: if we're already doing `position: absolute`, might as well just do `left: -12` here, to align with `top: -1`. And we probably don't need `paddingRight: 3`?
In fact, I don't think the `padding` will have any effect on a checkbox

jaredly:
Neat idea!

joecritch:
@jaredly Thanks man. Fixed up the styling as per your suggestion. Also, I've ensured the checkbox is a controlled component. (I was passing in something that was evaluating as `undefined`, so therefore wasn't being controlled. Doh!)

jaredly:
solid!

Issue #581, opened 2017-03-08
Update code to latest Flow version and fix violations


gaearon:
We are an a very old version of Flow. I’d appreciate if somebody could update the version and then go through all violations, either fixing the code or adding suppressions where reasonable.

tomaszlakomygp:
I'm on it 👍

gaearon:
Sounds good. It won’t be a lot of fun but I bet you’ll learn something. 😄

gaearon:
You’ll need to bump the version both in `.flowconfig` and `package.json`.

tomaszlakomygp:
```
... 37 more errors (only 50 out of 87 errors displayed)
To see all errors, re-run Flow with --show-all-errors
```

What have I gotten myself into 😭

tomaszlakomygp:
I better dig into the documentation first 👍

gaearon:
One important thing to keep in mind is that an error usually has two sides in Flow. Either something is annotated (or inferred) incorrectly, or used incorrectly. So fixing one fixes the other, but you need to decide which side is wrong. Explicitly annotating things sometimes helps to discover where inference produced something you didn’t expect.

tomaszlakomygp:
I got a case like that right from the start:
```
if (!this.ids.has(element)) {
      this.ids.set(element, guid());
      this.reactElements.set(this.ids.get(element), element);
    }
```

`reactElements` is defined as `reactElements: Map<ElementID, OpaqueNodeHandle>;` and flow throws an error because it is not aware that `this.ids.set(element, guid());`will set a value and we will not try to amend the `Map` with an `undefined` key

tomaszlakomygp:
Could you tell me if this is the proper approach in this case before I dig in to apply fixes for ~80 errors?

```
getId(element: OpaqueNodeHandle): ElementID {
    if (typeof element !== 'object' || !element) {
      return element;
    }
    if (!this.ids.has(element)) {
      this.ids.set(element, guid());
      // $FlowIssue flow is not aware that the line above ^ will set the value to this.ids
      this.reactElements.set(this.ids.get(element), element);
    }
      // $FlowIssue same as above, this will be always defined
    return this.ids.get(element);
  }
```

gaearon:
Remove indirection helps.

```js
const id = guid();
this.ids.set(element, id);
this.reactElements.set(id, element);
```

gaearon:
In case of return, you could add an explicit check, e.g.
```js
[...]

PR #595, opened 2017-03-25
[WIP] Upgraded flow and solved most of flow violations in codebase

tomaszlakomygp:
This is a Work In Progress PR which aims to fix https://github.com/facebook/react-devtools/issues/581

Currently most of the flow errors are addressed (13 are remaining out of 81 initial errors after updating the flow version). 

As this was my first experience with flow, some of the error messages were rather criptic to me (especially those related to eventListeners). Temporary solution was to of course use `any` but I find it to be suboptimal, perhaps someone can share some advice how to improve that.

All unit tests and eslint checks are passing properly


frontend/Draggable.js
```diff
@@ -15,8 +15,8 @@ var ReactDOM = require('react-dom');
 import type {DOMEvent} from './types';
 
 class Draggable extends React.Component {
-  _onMove: (evt: DOMEvent) => void;
-  _onUp: (evt: DOMEvent) => void;
+  _onMove: (evt: any) => void;
+  _onUp: (evt: any) => void;
```
tomaszlakomygp: Those caused me the most problems. I'm not quite sure how to solve it properly, because if you substitute `any` with `DOMEvent` you'll get a huge error message about not being able to call `addEventListener`/`removeEventListener` on all elements of the intersection

tomaszlakomygp:
BTW: @gaearon you were right, this was an exact opposite of fun 😄

plugins/DepGraph/crawlChildren.js
```diff
@@ -13,7 +13,7 @@
 var sep = '\x1f'; // separator
 // var MAX_DEPTH = 100;
 
-function crawlChildren(ptype: string, children: Array<string>, nodes: Map<string, Map>, depth: number, graph: Object) {
+function crawlChildren(ptype: string, children: Array<string>, nodes: Map<string, Map<any>>, depth: number, graph: Object) {
```
suchipi: Adding `any` here just to make flow happy is not good. What are the values in `nodes` really? Do we need to define a new type for it?

plugins/ReactNativeStyle/StyleEdit.js
```diff
@@ -39,7 +39,7 @@ class StyleEdit extends React.Component {
 
   onChange(name: string, val: string | number) {
     var num = Number(val);
-    this.props.onChange(name, num == val ? num : val);
+    this.props.onChange(name, num === val ? num : val);
```
suchipi: This looks like it will change behavior in an unintended way; since `val` can be either a string or number, the `==` check was probably intentional (since eg `4 == "4"`, and we are dealing with user input here). If flow isn't happy about the typing here, then maybe we need to coerce one of the results of the ternary into either a string or number.

plugins/ReactNativeStyle/setupBackend.js
```diff
@@ -45,6 +45,7 @@ function renameStyle(agent, id, oldName, newName, val) {
   if (!data || !data.updater || !data.updater.setInProps) {
     var el = agent.reactElements.get(id);
     if (el && el.setNativeProps) {
+      // $FlowFixMe we know that setNativeProps can be called here
```
suchipi: You can change the if condition to `el && typeof el.setNativeProps === 'function'` to make flow happy here, instead of using `$FlowFixMe`.

plugins/ReactNativeStyle/setupBackend.js
```diff
@@ -86,6 +87,7 @@ function setStyle(agent, id, attr, val) {
...
```
suchipi: Same as above; a `typeof` check will make flow happy.

shells/chrome/src/panel.js
```diff
@@ -37,7 +37,7 @@ var config: Props = {
       chrome.devtools.inspectedWindow.eval('inspect(window.__REACT_DEVTOOLS_GLOBAL_HOOK__.$node)');
     }, 100);
   },
-  showComponentSource(vbl) {
+  showComponentSource(vbl: any) {
```
suchipi: You can safely type `vbl` as string.

shells/chrome/src/panel.js
```diff
@@ -103,7 +103,9 @@ var node = document.getElementById('container');
 function reload() {
   setTimeout(() => {
     ReactDOM.unmountComponentAtNode(node);
-    node.innerHTML = '';
+    if (node) {
```
suchipi: `node` should always exist here, especially since we're about to call `ReactDOM.render` with it. This `if` here is confusing for a reader because it seems like it should be impossible; it's not clear that you're just checking to make flow happy. It might be less confusing to have an early throw if `node` doesn't exist, if you're just trying to satisfy flow.

shells/firefox/src/contentScript.js
```diff
@@ -46,14 +46,19 @@ function connectToBackend() {
...
 function injectBackend() {
   var node = document.createElement('script');
+  var documentElement = document.documentElement;
 
   node.onload = function() {
     window.postMessage({source: 'react-devtools-reporter'}, '*');
 
     connectToBackend();
-    node.parentNode.removeChild(node);
+    if (node.parentNode) {
```
suchipi: It should be impossible to not have a parent when `onload` is called; throwing an error would probably be more readable than adding a seemingly-unnecessary condition check to the code to just make flow happy.

shells/plain/container.js
```diff
@@ -20,7 +20,7 @@ window.React = React;
 
 var Panel = require('../../frontend/Panel');
 
-var target: Object = document.getElementById('target');
+var target: any = document.getElementById('target');
```
suchipi: This should probably be `?HTMLElement` instead

shells/plain/container.js
```diff
@@ -36,10 +36,13 @@ if (iframeSrc) {
 
 window.addEventListener('keydown', function(e) {
   if (e.altKey && e.keyCode === 68) { // Alt + D
-    if (document.body.className === 'devtools-bottom') {
-      document.body.className = 'devtools-right';
-    } else {
-      document.body.className = 'devtools-bottom';
+    var documentBody = document.body;
```
suchipi: `document.body` should be safe and reliable in practice. Rather than adding indirection to the code, an early throw if it is not present is a cleaner way to satisfy flow.

tomaszlakomygp:
@suchipi thanks for the review 👍 I'm learning a lot, I'll amend this PR once I have some extra time :)

frontend/Draggable.js
tomaszlakomygp: See the error output here:

<details>
```
frontend/Draggable.js:18
 18:   _onMove: (evt: MouseEvent) => void;
                      ^^^^^^^^^^ Event. Covariant property `preventDefault` incompatible with invariant use in
 44:   onMove(evt: DOMEvent) {
                   ^^^^^^^^ object type

frontend/Draggable.js:18
 18:   _onMove: (evt: MouseEvent) => void;
                      ^^^^^^^^^^ Event. Covariant property `stopPropagation` incompatible with invariant use in
 44:   onMove(evt: DOMEvent) {
                   ^^^^^^^^ object type

frontend/Draggable.js:19
 19:   _onUp: (evt: MouseEvent) => void;
                    ^^^^^^^^^^ Event. Covariant property `preventDefault` incompatible with invariant use in
 49:   onUp(evt: DOMEvent) {
                 ^^^^^^^^ object type

frontend/Draggable.js:19
 19:   _onUp: (evt: MouseEvent) => void;
                    ^^^^^^^^^^ Event. Covariant property `stopPropagation` incompatible with invariant use in
 49:   onUp(evt: DOMEvent) {
                 ^^^^^^^^ object type

frontend/Draggable.js:44
 44:   onMove(evt: DOMEvent) {
                   ^^^^^^^^ property `cancelBubble`. Property not found in
 18:   _onMove: (evt: MouseEvent) => void;
                      ^^^^^^^^^^ MouseEvent

frontend/Draggable.js:44
 44:   onMove(evt: DOMEvent) {
                   ^^^^^^^^ property `keyCode`. Property not found in
 18:   _onMove: (evt: MouseEvent) => void;
                      ^^^^^^^^^^ MouseEvent

frontend/Draggable.js:44
 44:   onMove(evt: DOMEvent) {
                   ^^^^^^^^ property `key`. Property not found in
 18:   _onMove: (evt: MouseEvent) => void;
                      ^^^^^^^^^^ MouseEvent

frontend/Draggable.js:49
 49:   onUp(evt: DOMEvent) {
                 ^^^^^^^^ property `cancelBubble`. Property not found in
 19:   _onUp: (evt: MouseEvent) => void;
                    ^^^^^^^^^^ MouseEvent

frontend/Draggable.js:49
 49:   onUp(evt: DOMEvent) {
                 ^^^^^^^^ property `keyCode`. Property not found in
 19:   _onUp: (evt: MouseEvent) => void;
                    ^^^^^^^^^^ MouseEvent

frontend/Draggable.js:49
 49:   onUp(evt: DOMEvent) {
                 ^^^^^^^^ property `key`. Property not found in
 19:   _onUp: (evt: MouseEvent) => void;
                    ^^^^^^^^^^ MouseEvent
```
</details>

gaearon:
@suchipi Thanks for a comprehensive review!

suchipi:
Sure thing 😃

gaearon:
@tomaszlakomygp Do you intend to pick this up?

tomaszlakomygp:
@gaearon 

I do, I haven't had enough time lately to finish it. Sorry for the delay :/

frontend/Draggable.js
tomaszlakomygp: I tried that, and it doesn't seem to do the trick. 

See the error here: https://gist.github.com/tomaszlakomygp/71153238662af3781732a0ff82597058

tomaszlakomygp:
In the meantime I've rebased this branch off master and bumped flow to the newest, current version - 0.44.1

tomaszlakomygp:
@gaearon I'm going on a business trip next week and I might have hard time squeezing enough free time to finish this up. If someone else is willing to take over this issue I won't mind :)

tomaszlakomygp:
I'm sorry but I need to close this PR. Due to some personal important affairs I won't be able to finish this in a foreseeable future and it's branch is way behind `master` already.

gaearon:
I see, thanks for giving it a try!

tomaszlakomygp:
No worries, I just need to focus on some personal matters at the moment :) Expect more PRs to dev-tools after I sort some stuff

Issue #580, opened 2017-03-08
React devtools tab not showing up on FF nor on Chrome


istvano:
Hi,

I have installed the extension on Chrome, but it stays gray no matter what website I visit. 
- I tried my own react app ( Redux devtools works! )
- Facebook
- React page ( https://facebook.github.io/react/ )

Also on chrome when I click on the icon the Options menu stays grey and inaccessible

Firefox same thing. I installed it, but never shows up on the developer toolbar.

I am not sure how I could debug this further. Any clues would be gratelly appreciated.

Thanks

istvano:
Hi, I managed to enable it on FF removing it and adding it again. On Chrome same problem unforuntatellty.

gaearon:
Can you remove all other extensions and try again? Maybe some other extensions is interfering with it somehow.

istvano:
Hi,

I tried it in incognito mode same result. Do I need to fully remove my extensions ?

gaearon:
I don't know. Since I can't reproduce it, it sounds very odd and specific to your browser installation.

istvano:
Hi,

No matter what I do I get the same, the options menu is disabled and the extension is not functioning.
on Version 56.0.2924.87 (64-bit).

I guess we can close the issue. I am using FF to debug react specific problems now.

Thanks for your help!

gaearon:
Can you confirm the version of the extension itself?
It would also be great to try with a clean Chrome install.

istvano:
it is the latest: Version: 2.0.12. I will try to do it, I will check how I can save my current settings

istvano:
Hi,

1: I have backuped up profile
2: uninstall chrome from windows 10 64 bit professional
3: deleted anything chrome related
4, installed latest chrome
5, installed React Chrome extension
6, options menu still greyed out. visiting www.reactjs.com shows inactive toolbar

I guess not much else to try :)

gaearon:
OK let's reopen it. I'm going to try it specifically on Windows 10 when I get some time.

gaearon:
Tagging as "good first bug" in case somebody could help reproducing this.

markneumann:
FWIW, Chrome Version 57.0.2987.98 (64-bit), OSX 10.12.3, React Dev 2.0.12,
all extensions disabled except React Dev,

Same problem.  DevTools tab does not appear in Chrome Dev Tools.

zinoviev:
Don't even know icon should be colored sometimes

[...]

PR #597, opened 2017-03-26
Ensured that React devtools icon is displayed in color

tomaszlakomygp:
This solves one of the issues reported in https://github.com/facebook/react-devtools/issues/580 that is - the React devtools icon in Chrome was always greyed out.

Before:
screen shot 2017-03-26 at 15 41 53

After:
screen shot 2017-03-26 at 15 42 28


jaredly:
oh cool, thanks!

gaearon:
One confusing thing here is that it doesn’t actually do anything, but now it draws more attention.

Issue #566, opened 2017-02-24
Weird highlighting over a text node


gaearon:
Normally we highlight matching DOM nodes on hover.

This doesn't work for text nodes because they don't have `getBoundingClientRect`. Just like the browser, we don't highlight anything when you hover over a text node.

Except, if you change [this line](https://github.com/facebook/react-devtools/blob/568b1586ef15069fd024fc1f76d17af17a68de24/shells/plain/container.js#L26) to point at `sink.js` (and build it by running `../../node_modules/.bin/webpack` in `test/example`), and open the Deep Tree example, you'll notice that hovering over the text nodes produces a weird empty highlighter box:

<img width="137" alt="screen shot 2017-02-24 at 9 56 51 pm" src="https://cloud.githubusercontent.com/assets/810438/23322592/98dec5b2-fadc-11e6-808b-138362a71c06.png">
<img width="64" alt="screen shot 2017-02-24 at 9 56 56 pm" src="https://cloud.githubusercontent.com/assets/810438/23322593/98e15430-fadc-11e6-889c-4a4ff8ff8e80.png">

We need to figure out why this is happening, and stop showing it.

spicyj:
You can call getBoundingClientRect on a range of text though:

https://developer.mozilla.org/en-US/docs/Web/API/Range/getBoundingClientRect

Just note that it's buggy in Chrome; I had to write a new version to work around the issue:

https://github.com/facebook/draft-js/blob/e4f87b04b3ed22134961c8ba7be4e8c66dadd4c4/src/component/selection/getRangeBoundingClientRect.js

Maybe we should move that to fbjs and use it in this project.

tomaszlakomygp:
Issue resolved in https://github.com/facebook/react-devtools/pull/598, I believe we can close it

srkchandru:
As of now we do not show highlighter for text node. We ignore the hover if it is not a element node. Can we can show a highlighter similar to chrome or firefox for text nodes. I would like to contribute. Can you re-open this issue? @gaearon

PR #598, opened 2017-03-26
Ensured that an empty overlay won't be displayed when hovering over a text node

tomaszlakomygp:
This should solve the issue reported in https://github.com/facebook/react-devtools/issues/566. An empty `Overlay` box was displayed because `Overlay` component has default styles applied when it is created. When hovering over a text node the `inspect` function in `Overlay` component would return without adding any content and because of that user was left with an empty yellow box.

Before:
before

After:
after


jaredly:
looks good, thanks!

PR #602, opened 2017-03-31
allow standlone to listen on custom port

### LINKED ISSUE #600: Changing the port from 8097?
Hi everyone, 

Noob question here (and not sure if this is the right place to ask it), but how do you go about changing the default port from 8097 to a custom one, say 3001?

[I've asked the question on StackOverflow](http://stackoverflow.com/questions/43074677/react-devtools-by-default-devtools-listen-to-port-8097-on-localhost-how-to-cus), although figured I'd ask here too.

Sorry if this isn't the correct place to ask - it's the first time I've created an issue on GitHub.

jaredly:
There was a way to make `react-devtools-core` listen on a custom port, but no way for the frontend to start the server on that custom port!
This fixes #600

### DIFF
--- a/packages/react-devtools/app.html
+++ b/packages/react-devtools/app.html
@@ -78,7 +78,7 @@
             </div>
         </div>
         <script>
-            var port = 8097;
+            var port = process.env.PORT || 8097;
             var localIp = require('ip').address();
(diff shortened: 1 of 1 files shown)


tomaszlakomygp:
Great idea!

PR #609, opened 2017-04-19
Fix Firefox message source validation

gaearon:
#561 is too strict and doesn’t work in Debug mode for Firefox since it stubs out `window`.

We can get around it by checking against `document.defaultView` instead.
This gives us the real window.

It shouldn’t matter for other `window` references since this is the only place where we rely on referential equality.

See the discussion in https://github.com/facebook/react-devtools/pull/561#issuecomment-295364114 for more context.


Rob--W:
lgtm

PR #610, opened 2017-04-19
Don't call walkNode for null nodes

### LINKED ISSUE #584: Failing when a child is null
I have an element one of which childs is `null`. React devtools fails terribly (do not display anything/is just broken).

The problem occurs in getData function in backend.js, line 8440, when argument `element` is `null` (the line that causes the problem is 8457).

(getData is invoked from walkNode func (line 8367).

doytch:
This fixes #584 , where the React devtools would fatally crash when a node's child was `null`. Examples of impact projects included things like react-select (at least while being rendered by inferno-compat).

Poked around a bit and this looked like the best place to put this check but it's my first exposure to the code base so feel free to send it back for rework.

### DIFF
--- a/backend/attachRenderer.js
+++ b/backend/attachRenderer.js
@@ -175,7 +175,9 @@ function walkRoots(roots, onMount, onRoot, isPre013) {
 function walkNode(element, onMount, isPre013) {
   var data = isPre013 ? getData012(element) : getData(element);
   if (data.children && Array.isArray(data.children)) {
-    data.children.forEach(child => walkNode(child, onMount, isPre013));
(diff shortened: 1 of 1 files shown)


gaearon:
I think `data.children` is not expected to contain `null`s in the first place. Even if we fixed it here, it could crash somewhere else. Are you sure this is not an `inferno-compat` issue?

gaearon:
I’ll close because this does appear to be a problem with `inferno-devtools`, and is best fixed there. If you can reproduce this in vanilla React please let me know and we can revisit.

PR #616, opened 2017-04-21
Only show "$r in the console" for composites

gaearon:
Fixes it to not show up for `div`s.
@spicyj noticed this IIRC.


spicyj:
Can we fix it instead? I've wanted the DOM node several times.

gaearon:
https://github.com/facebook/react-devtools/issues/617

PR #618, opened 2017-04-21
Nicer color scheme

gaearon:
Inspired by https://github.com/facebook/react-devtools/pull/553, I decided to revamp the colors to match what Chrome does. The main change is that the selected node is properly highlighted even on low contrast displays. Some work still pending.

Before:

After:

- [x] Change highlighting color when panel is inactive
- [x] Add vertical guide for currently selected tag pair

### DIFF
--- a/frontend/Breadcrumb.js
+++ b/frontend/Breadcrumb.js
@@ -119,7 +119,7 @@ module.exports = decorate({
   props(store, props) {
     return {
       select: id => store.selectBreadcrumb(id),
-      hover: (id, isHovered) => store.setHover(id, isHovered),
+      hover: (id, isHovered) => store.setHover(id, isHovered, false),
       selected: store.selected,
       path: getBreadcrumbPath(store),
     };
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -567,7 +567,7 @@ var styles = {
   guidelineSelect: {
     backgroundColor: '#a9c5ef',
     opacity: 1,
-  }
+  },
 
 };
 
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -372,6 +372,11 @@ class Node extends React.Component {
     var guidelineStyle = assign(
       {
         left: leftPad.paddingLeft - 7,
+        // Bring it in front of the hovered children, but make sure
+        // hovering over parents doesn't draw on top of selected
+        // guideline even when we've selected the closing tag.
+        // When unsure, refer to how Chrome does it (it's subtle!)
+        zIndex: selected ? 1 : 0,
       },
       styles.guideline,
       this.props.hovered && styles.guidelineHover,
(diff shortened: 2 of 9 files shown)


rauchg:
Dark mode support ⬛️

rauchg:
The component tree first shows up completely collapsed when turning on the tools. Seems like at least a few levels expanded would be a better default? 🌲

gaearon:
I don't think I'm ready to go for a dark mode since it's a lot of changes. Contributions welcome here.

For auto expanding, I'm not sure. Doesn't make sense if you have a lot of roots. I'd prefer we persist the last expansion path (also a contribution opportunity!)

stryju:
re: expanding: having the selected node +1 level expanded sounds reasonable (and desirable) IMO

Kerumen:
I can set up a PR for the dark mode!

cezarneaga:
I'd like a lock component feature. Much like how you add a debug breakpoint and then after i make code changes the expansion persists there. 

Or persist the search component until i clear it to achieve the same

dylang:
A minor idea, in case you were interested. I won't be offended if this is outside what you are hoping to accomplish with this work.

The `breadcrumbs` and `search` don't match Chrome's styling. It's probably fine for most developers, but for me, my eyes don't pick up on those two features. It's like my brain assumes that if it doesn't look as good, it's not going to work as well. 

React Dev Tools
screen shot 2017-04-22 at 3 42 27 pm

Chrome 
screen shot 2017-04-22 at 3 42 45 pm

Chrome with search open
screen shot 2017-04-22 at 3 51 33 pm

It seems that Chrome's own search feature here doesn't really match their style, such as the bubble-style cancel button.

dylang:
A different font to help distinguish property names?
operator
(The font would have to be one either already installed, or one available to be distributed with the extension.)

My IDE (WebStorm) doesn't let me use multiple fonts, so I can't comment on the usefulness, but it seems nice when I see it in screenshots and screencasts.

gaearon:
@Kerumen PR for dark more would be rad!

Kerumen:
@gaearon I'm working on it.

radubrehar:
`== $0` is also nice and suble :) => `== $r` for react dev-tools
screenshot 2017-04-25 17 04 14

manuelbieh:
Watch expressions for props and state to trace where a props or state change has its origin. Rightclick a prop in the sidebar of the selected element and then an option "Add watch expression" and then trigger `debugger` once a prop has changed. I think that could be a nice alternative to adding `debugger` to `componentWillUpdate`.

gaearon:
>== $0 is also nice and suble :) => == $r for react dev-tools

Ooh this is nice. Could you file a new issue for this?

gaearon:
>Watch expressions for props and state to trace where a props or state change has its origin. 

Interesting. Please file a new issue too?

radubrehar:
@gaearon sure - will add the issue about `$0`.

vjeux:
A dream of mine would be to make the colors better, right now it's a christmas tree!

ansumanshah:
@Kerumen Would like to help with Dark mode too, is there a PR/Fork in place to get started?

Kerumen:
@ansumanshah I've just opened a PR: https://github.com/facebook/react-devtools/pull/669!

PR #620, opened 2017-04-22
Refactor split pane bugfix to avoid magic numbers 

### LINKED ISSUE #615: Refactor split pane bugfix to avoid magic numbers
See https://github.com/facebook/react-devtools/issues/611 and https://github.com/facebook/react-devtools/pull/614.

I’d like to avoid magic numbers, and instead properly fix it. (Maybe we should read something from the right pane DOM node directly.)

shriah:
Fixed #615

The fix is to create a wrapper element around the right pane without any padding or margin so we can set the width or height we calculate directly to it. http://recordit.co/TU3CHcHNbN

I also changed the min width of the right pan to 150px because at 100px it was getting too squashed to the right.

### DIFF
--- a/frontend/SplitPane.js
+++ b/frontend/SplitPane.js
@@ -61,49 +61,43 @@ class SplitPane extends React.Component {
   onMove(x: number, y: number) {
     var node = ReactDOM.findDOMNode(this);
 
-    // TODO: I don't understand what this is doing.
-    // It's probably related to margin and padding but it's necessary to avoid jumps:
-    // https://github.com/facebook/react-devtools/issues/611
-    // I can live with this for now but we should fix it.
-    // How I verify that it works:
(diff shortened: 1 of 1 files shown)


frontend/SplitPane.js
```diff
@@ -61,49 +61,43 @@ class SplitPane extends React.Component {
...
-          style={draggerStyles}
-          onStart={() => this.setState({moving: true})}
-          onMove={(x, y) => this.onMove(x, y)}
-          onStop={() => this.setState({moving: false})}>
-          <div style={styles.draggerInner} />
-        </Draggable>
-        <div style={rightStyle}>
-          {this.props.right()}
+        <div style={rightStyles}>
+            <Draggable
```
gaearon: Nitpick: please use two space indentation

gaearon:
This looks great, thanks. I won’t pretend to understand it, as CSS is not my strong side. But it does look cleaner. 😃

Issue #591, opened 2017-03-20
(Suggestion) Expand / collapse node and all its children 


jerikson:
As possible using Chrome Developer Tools:

![suggestionjerikson](https://cloud.githubusercontent.com/assets/4104014/24120264/45e0ba1a-0db4-11e7-8667-3b4e99177e0c.png)

By clickling either `Ctrl` + `Alt `+ &#9654; using Windows. Or `Opt` + &#9654;  using a Mac would
toggle the html nodes and all of its children &#9660;

I know Chrome gives the opportunity to manually configure short-commands..
[chrome://extensions/configureCommands]()

However the end-user of react-devtools would expect the same behavior anyway? 
Obviously, this is not a high prior issue, and I have not completely thought this through myself, maybe there is a good reason for this.

https://developer.chrome.com/extensions/commands
https://developers.google.com/web/tools/chrome-devtools/shortcuts

Kind regards.
Jimmy

gaearon:
I’d happily review a PR adding this.

shriah:
Added a pull request for this. It works but need to test it on a website with large number of nodes.

gaearon:
Fixed via https://github.com/facebook/react-devtools/pull/621.
Will be out whenever we cut the next patch (probably in a week or so).

jerikson:
Thanks everyone! 👍

PR #621, opened 2017-04-23
Add ability to expand or collapse all the children of selected node

shriah:
Adds #591 

Took inspiration from `_revealDeep`. Not sure how efficient it is when the total nodes are large. Added shortcuts in `keyboardNav` but not sure if it is the right way to do it. This should work in all the shells.

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -63,6 +63,7 @@ const DEFAULT_PLACEHOLDER = 'Search by Component Name';
  * - hideContextMenu
  * - selectFirstSearchResult
  * - toggleCollapse
+ * - toggleAllChildrenNodes
  * - setProps/State/Context
  * - makeGlobal(id, path)
  * - setHover(id, isHovered)
@@ -340,7 +341,7 @@ class Store extends EventEmitter {
     this.emit(id);
   }
 
-  toggle(value: boolean) {
+  toggleAllChildrenNodes(value: boolean) {
     var id = this.selected;
     if (!id) {
       return;
@@ -556,7 +557,7 @@ class Store extends EventEmitter {
     if (children && children.forEach) {
       children.forEach(cid => this._toggleDeepChildren(cid, value));
     }
-    
+
   }
 
   _mountComponent(data: DataType) {
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -557,7 +557,6 @@ class Store extends EventEmitter {
     if (children && children.forEach) {
       children.forEach(cid => this._toggleDeepChildren(cid, value));
     }
-
   }
 
   _mountComponent(data: DataType) {
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -340,6 +340,14 @@ class Store extends EventEmitter {
     this.emit(id);
   }
 
+  toggle(value) {
+    var id = this.selected;
+    if (!id) {
+      return;
+    }
+    this._toggleDeepChildren(id, value);
+  }
+
   setProps(id: ElementID, path: Array<string>, value: any) {
     this._bridge.send('setProps', {id, path, value});
   }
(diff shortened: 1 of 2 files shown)


frontend/keyboardNav.js
```diff
@@ -26,6 +26,9 @@ var keyCodes = {
   '38': 'up',
   '39': 'right',
   '40': 'down',
+
+  '69': 'expand',
```
gaearon: Please change this to match Chrome shortcuts.

frontend/Store.js
```diff
@@ -340,6 +340,14 @@ class Store extends EventEmitter {
     this.emit(id);
   }
 
+  toggle(value: boolean) {
```
gaearon: Can we give this a more descriptive name?

shriah: will `toggleAllChildrenNode` be more descriptive ?

frontend/keyboardNav.js
shriah: I will have to add the shortcut to manifest.json of shells/chrome. Right now there is no support for shortcut in any of the plugins. Will spend some time to try to figure out how to do this.

frontend/Store.js
gaearon: Yea. ("Nodes" if plural.)

frontend/keyboardNav.js
shriah: I added support for using chrome shortcuts by adding it in the manifest file but the shortcut stopped working in the chrome elements panel. Not sure if this is desired.

I also tried this feature in a recently large react app and the expand all on the root node has a small lag and the close all works well. This lag is there even in the elements panel of chrome so this should be acceptable.

shriah:
Updated the PR with support for chrome shortcuts.

frontend/keyboardNav.js
```diff
@@ -43,6 +43,17 @@ module.exports = function keyboardNav(store: Store, win: Object): (e: DOMEvent)
       return;
     }
     e.preventDefault();
+    if ((e.altKey && direction === 'right') || (e.ctrlKey && e.altKey && direction === 'right')) {
```
gaearon: Isn't the second condition extraneous? It seems to me that the left is always true when the right is true.
Same below.

frontend/Store.js
```diff
@@ -535,6 +544,22 @@ class Store extends EventEmitter {
...
+    }
+    if (!(node.get('collapsed') === value)) {
+      this._nodes = this._nodes.setIn([id, 'collapsed'], value);
+      this.emit(id);
+    }
+    var children = node.get('children');
+    if (children && children.forEach) {
+      children.forEach(cid => this._toggleDeepChildren(cid, value));
+    }
+
```
gaearon: Nit: unnecessary newline.

frontend/keyboardNav.js
shriah: I added this to disallow Control + Right Arrow

gaearon: `if (e.altKey && direction === 'right')` is enough to match both Alt+Right and Ctrl+Alt+Right, isn't it? And it wouldn't match Ctrl+Right.

gaearon:
This looks good, thanks!

aspiers:
Cool, thanks a lot! What is the shortcut on Linux?

gaearon:
Alt+Right and Alt+Left should work.

gaearon:
(But the change is not released yet.)

zhanwenchen:
Is this released yet? It's been 4 months

bvaughn:
Yes 😄  Probably faster to just try it out than post 😁

zhanwenchen:
Ah it's Alt+Left_Arrow and Alt+Right_Arrow

PR #629, opened 2017-04-25
Clean up a few hacks

gaearon:
Nothing important.

### DIFF
--- a/packages/react-devtools-core/src/backend.js
+++ b/packages/react-devtools-core/src/backend.js
@@ -17,13 +17,6 @@ type ConnectOptions = {
   isAppActive?: () => boolean,
 };
 
-// TODO: why?
-if (!window.performance) {
-  window.performance = {
-    now: () => Date.now(),
-  };
-}
-
 var Agent = require('../../../agent/Agent');
 var Bridge = require('../../../agent/Bridge');
 var installGlobalHook = require('../../../backend/installGlobalHook');
@@ -36,6 +29,7 @@ installGlobalHook(window);
 installRelayHook(window);
 
 if (window.document) {
+  // This shell is universal, and might be used inside a web app.
   window.__REACT_DEVTOOLS_GLOBAL_HOOK__.on('react-devtools', agent => {
     var setupHighlighter = require('../../../frontend/Highlighter/setup');
     setupHighlighter(agent);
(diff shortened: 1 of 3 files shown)


packages/react-devtools-core/src/backend.js
```diff
@@ -17,13 +17,6 @@ type ConnectOptions = {
   isAppActive?: () => boolean,
 };
 
-// TODO: why?
-if (!window.performance) {
```
gaearon: The only place depending on `performance.now` seems to be Relay plugin so inlined this.

packages/react-devtools-core/src/backend.js
```diff
@@ -36,6 +29,7 @@ installGlobalHook(window);
 installRelayHook(window);
 
 if (window.document) {
+  // This shell is universal, and might be used inside a web app.
```
gaearon: This wasn't obvious.

shells/chrome/src/backend.js
```diff
@@ -15,15 +15,8 @@ var BananaSlugBackendManager = require('../../../plugins/BananaSlug/BananaSlugBa
 var Bridge = require('../../../agent/Bridge');
 var inject = require('../../../agent/inject');
 var setupHighlighter = require('../../../frontend/Highlighter/setup');
-var setupRNStyle = require('../../../plugins/ReactNativeStyle/setupBackend');
 var setupRelay = require('../../../plugins/Relay/backend');
 
-// TODO: check to see if we're in RN before doing this?
-setInterval(function() {
```
gaearon: This is Chrome shell. We can't be in RN, and Chrome shell hasn't work on for a long time.

jhen0409:
I think we can keep the `ReactNativeStyle` plugin in Chrome part, it can be used for [`react-native-web`](https://github.com/necolas/react-native-web) (need assign [`resolveRNStyle`](https://github.com/necolas/react-native-web/blob/master/src/apis/StyleSheet/flattenStyle.js) to hook), that will be useful because the style editor in `Elements` tab of devtools is very complicated for RNW.

gaearon:
Did it work before? Could you open a new issue about it?

jhen0409:
OK, I opened the issue #634.

Issue #384, opened 2016-06-14
Stateless Functional Component source code doesn't show up correctly


keyanzhang:
cc @spicyj

jaredly:
? please elaborate

keyanzhang:
Hey sorry @jaredly I forgot to update the description... 😢 

The repro is at http://s.codepen.io/spicyj/debug/qNOZOP and the corresponding implementation is at http://codepen.io/spicyj/pen/qNOZOP.

When I right click at a `Board` component (which is a class component) in devtools and "show source", it jumps to its implementation correctly. However if I click "show source" at `Square` (which is a stateless functional component), it takes me to the source code of `ReactCompositeComponent` instead.

Thanks!

jaredly:
Gotcha! didn't realize what "SFC" meant :)

tomaszlakomygp:
I'm getting a "This debug view expired." when trying to access the repro link :/

ShiXiaoyYu:
PR https://github.com/facebook/react-devtools/pull/557 try to fixes this issue. @jaredly

PR #630, opened 2017-04-25
Fix Show Source for functional components

gaearon:
This should fix https://github.com/facebook/react-devtools/issues/384.
Verified by adding a functional component and "show source"ing it.

I have not tested the `createClass` case yet but this should be review-able.


gaearon:
Fairly sure this works.

gaearon:
cc @spicyj @jaredly in case I messed up.

I don't understand what the original code was doing with `isMounted()` but seems like jumping to `render()` is more useful in both cases.

spicyj:
I think @jaredly's original idea was that the constructor was more useful for non-createClass components, but I agree that render seems good.

Issue #625, opened 2017-04-25
Make component source link clickable


gaearon:
It would be nice if the component source link in the right bottom corner was clickable.

I'm not sure if this is doable. Unlike "Show Source", we don't want to navigate to *function definition*, but to *`createElement` call*. We can only do so if the user has `babel-plugin-transform-react-jsx-source` plugin, but it is enabled in development for projects generated with Create React App.

All information we'd have is filename and line number. While this may not be enough in Chrome, it might be enough for Electron, so maybe we should at least enable this feature for Electron shell. (See https://github.com/facebook/react-devtools/pull/588 for related PR).

How "Show Source" works:

![](https://d2ppvlu71ri8gs.cloudfront.net/items/451n2u1s2c1Q1e1q1u1u/Screen%20Recording%202017-04-25%20at%2002.51%20PM.gif?v=f4dcd4b2)

Here is the link I want to make clickable:

![](https://d2ppvlu71ri8gs.cloudfront.net/items/1S0G1X0K1k2X1i3B451k/Screen%20Recording%202017-04-25%20at%2002.51%20PM.gif?v=c72dbc33)

nanot1m:
Is it good variant to pass prop with viewSource action through the Panel component to PropState component?

gaearon:
Yes, maybe. Note I slightly tweaked related code in https://github.com/facebook/react-devtools/pull/630 because of https://github.com/facebook/react-devtools/issues/631. So you'd need to rebase on top of that. Tagging issue for you.

gaearon:
Fixed in https://github.com/facebook/react-devtools/pull/633.

PR #633, opened 2017-04-25
[Electron] Make component source link clickable

nanot1m:
Added possibility to click element source link in Electron shell #625


frontend/Panel.js
```diff
@@ -295,6 +295,7 @@ class Panel extends React.Component {
         }}
         extraPanes={extraPanes}
         extraTabs={extraTabs}
+        onViewElementSource={(id, source) => this.viewElementSource(id, source)}
```
gaearon: Should this always be passed? Or only when `this.props.showElementSource` is passed? Note that `viewElementSource` crashes otherwise. For example `showElementSource` doesn't exist in Chrome shell.

nanot1m: Yes, it should be passed only if `showElementSource` prop is defined.
Maybe I should add flow-types to `PropState` component props and for `decorate` function?

gaearon:
Thanks again. I'll review this a bit later—going to make a release with current changes. Feel free to work on other issues in the meantime if you'd like!

gaearon:
This is great, thanks!

Issue #627, opened 2017-04-25
Deleting a style in RN Style Editor puts it under an empty key


gaearon:
I tried deleting a style from the RN style editor by removing its key:

<img width="297" alt="screen shot 2017-04-25 at 3 07 01 pm" src="https://cloud.githubusercontent.com/assets/810438/25389578/ea3233b6-29c8-11e7-8850-130020c2cb82.png">

I would expect that the textbox pair would disappear on blur, and this particular style would be removed. However it stays, and RN complains about `''` being used as a key in the `style` object:

<img width="305" alt="screen shot 2017-04-25 at 3 07 03 pm" src="https://cloud.githubusercontent.com/assets/810438/25389606/01018f4c-29c9-11e7-82a3-92a72a56b4b8.png">

We should treat clearing the key textbox as deleting a style key on blur.

tchaffee:
Not sure how to reproduce this. Where do I find the RN style editor?

soul-wish:
@gaearon can I try to fix this?

soul-wish:
seems like @tchaffee was faster :)

tchaffee:
@soul-wish I don't even know how to reproduce this and it seems like it might be for React Native? If you have more of a clue than me, go for it.

soul-wish:
You are right, @tchaffee. As I can see the problem is in `plugins/ReactNativeStyle` module. @gaearon, correct me if I'm wrong.

gaearon:
Yes, this is for RN. I left some instructions for running RN shell here: https://github.com/facebook/react-devtools/issues/624. You can reproduce this with any RN example (for example UIExplorer in React Native repo).

gaearon:
Please sort it out between yourselves 😉 
Or let me know if you’d like to untag.

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #635, opened 2017-04-25
[React Native] Fix issue with style editor empty key (#627)

soul-wish:
Added additional check for empty key (inside `_handleStyleRename` function). Now, there is no warning message about empty key in `style` object. Please, let me know if I've missed something.

### DIFF
--- a/plugins/ReactNativeStyle/ReactNativeStyle.js
+++ b/plugins/ReactNativeStyle/ReactNativeStyle.js
@@ -102,6 +102,10 @@ class NativeStyler extends React.Component {
   _handleStyleRename(oldName: string, newName: string, val: string | number) {
     var style = shallowClone(this.state.style);
     delete style[oldName];
+    if (newName === '') {
+      this.setState({style});
+      return;
+    }
     style[newName] = val;
     this.props.bridge.send('rn-style:rename', {id: this.props.id, oldName, newName, val});
     this.setState({style});
--- a/plugins/ReactNativeStyle/ReactNativeStyle.js
+++ b/plugins/ReactNativeStyle/ReactNativeStyle.js
@@ -102,10 +102,6 @@ class NativeStyler extends React.Component {
   _handleStyleRename(oldName: string, newName: string, val: string | number) {
     var style = shallowClone(this.state.style);
     delete style[oldName];
-    if (newName === '') {
-      this.setState({style});
-      return;
-    }
     style[newName] = val;
     this.props.bridge.send('rn-style:rename', {id: this.props.id, oldName, newName, val});
     this.setState({style});
--- a/plugins/ReactNativeStyle/setupBackend.js
+++ b/plugins/ReactNativeStyle/setupBackend.js
@@ -93,7 +93,10 @@ function shallowClone(obj) {
 
 function renameStyle(agent, id, oldName, newName, val) {
   var data = agent.elementData.get(id);
-  var newStyle = {[newName]: val};
+  var newStyle = {};
+  if (newName !== '') {
+    newStyle = {[newName]: val};
+  }
   if (!data || !data.updater || !data.updater.setInProps) {
     var el = agent.internalInstancesById.get(id);
     if (el && el.setNativeProps) {
@@ -109,7 +112,9 @@ function renameStyle(agent, id, oldName, newName, val) {
     if (typeof style[style.length - 1] === 'object' && !Array.isArray(style[style.length - 1])) {
       customStyle = shallowClone(style[style.length - 1]);
       delete customStyle[oldName];
-      customStyle[newName] = val;
+      if (newName !== '') {
+        customStyle[newName] = val;
+      }
       // $FlowFixMe we know that updater is not null here
       data.updater.setInProps(['style', style.length - 1], customStyle);
     } else {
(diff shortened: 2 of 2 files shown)


plugins/ReactNativeStyle/ReactNativeStyle.js
```diff
@@ -102,6 +102,10 @@ class NativeStyler extends React.Component {
   _handleStyleRename(oldName: string, newName: string, val: string | number) {
     var style = shallowClone(this.state.style);
     delete style[oldName];
+    if (newName === '') {
+      this.setState({style});
+      return;
```
jhen0409: It looks like will not actually work, should keep the `rn-style:rename` event if it rename to `''`, then remove it in RN client.

See [the backend code](https://github.com/facebook/react-devtools/blob/master/plugins/ReactNativeStyle/setupBackend.js#L94).

soul-wish: Hi, @jhen0409. Thanks for your comment. Can you please help me with build process for backend code? I tried to do some changes in `plugins/ReactNativeStyle/setupBackend.js`, then run `npm run backend` inside `packages/react-devtools-core`, then run `npm start` inside `packages/react-devtools` directory but it seems like my changes doesn't have any affect on react devtools and it uses old/existing code. I'm really new to this project so it looks like I've missed something important :)

jhen0409: The backend script should included in react-native client, so you need to run `npm link <path-of-react-devtools-core>` in your react-native project. If you can't see any changes, just restart react-native packager.

Remember to use RN ^0.43, because v0.42 before are used the old [backend vendor](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools-core/vendor).

soul-wish: Thanks you so much, @jhen0409! Finally got it working. Can you check my last commit? Am I on the right path?

gaearon:
The intended behavior is that the style rule gets deleted.
However if I clear the key, it just reappears:



Can we fix it to delete the key?

soul-wish:
Thanks, @gaearon. With these changes it removes input pairs for the new rules (not the ones that were set by RN app itself). So in case you add the new CSS rules with a help of react-devtools and then try to delete their keys -- the rules will disappear (I mean the pair of inputs and the rule itself). Don't sure how to fix issue with the original rules right now..

gaearon:
I see. This makes sense. Let me think about it.

gaearon:
While this looks okay, I’d appreciate if you could compile a comprehensive list of current UX issues with style editor. It feels very flaky right now: it appears to behave differently for `View`s and `RCTView`s, some rules can deleted, but others can’t, etc. It would be cool to at least have a comprehensive overview of current UX issues, and maybe figure out a plan to addressing them all at once. For example maybe we should disable editing `RCTView`s if it’s not reliable. Would you like to help with this?

soul-wish:
Surely! I'll try to do my best. I'm not sure if I'll be able to fix any of these issues but I'll try to find some of them at least. To be honest, this is my first time with RN, so I can miss some points :). Anyway, give me some time and I'll provide you with some details/issues from my point of view. Thanks for your patience!

gaearon:
>To be honest, this is my first time with RN, so I can miss some points :). 

This is actually great because you aren’t yet used to brokenness and can see things clearly.

soul-wish:
Hi!
Here is my small review.

---

1. There are some inconsistencies in style editor UI (some paddings, input's layout etc.). But I've checked other issues/PR's in the repo and saw that all these issues will be fixed in the nearest future;

---

2. It would be really cool to be able to navigate to the style value by pressing tab key on the keyboard. Now there is only one input field for the new style key, and only after blur event second field appears. So right now there is no way to add new styles by keyboard only - you should always use your touchpad or mouse;

blur

---

3. If I got it correctly, `RCTView` is always a child of `View` so it inherits all its styles (correct me if I'm wrong). And I've noticed that you can not simply edit `RCTView` styles because they will be reset automatically from the parent `View`. So, maybe it would be better just to display all these styles and not allow to change them (readonly fields, or even better - some type of list with style's keys and values);

---

4. As for the `View`s styles I've found few strange issues: 
    - I can not delete original styles (they just keep reappearing, but maybe it should work in this way);
original-styles

    - But I can delete the styles that were added by me manually (with a small fix in this PR);
new-styles

    ---

    - I can not change the keys of the styles (but I can change their values);
original-styles-values-keys

    ---

    - If I try to change the key of the original style it will be added as a new style record (key->value) at the bottom of styles editing form;
original-styles-edit-key

    ---

5. I'm not sure if it is possible to not add new style if the key is incorrect? Right now it works in this way:

incorrect-keys

---

Just a question: is it possible to create the same experience for RN styles editor as we have in native Chrome devtools (I mean inline editing etc.)?

PS. I apologize in case I miss some important points or concepts (as I said earlier - I have zero experience in RN) ;)

gaearon:
>Just a question: is it possible to create the same experience for RN styles editor as we have in native Chrome devtools (I mean inline editing etc.)?

I don’t know, honestly! Thanks for compiling this list. Would you mind creating a new issue with these points, and copy and paste this exploration there?

ansumanshah:
I have pushed a PR with inline editing, working on autosize taking inspiration from [this](https://github.com/JedWatson/react-input-autosize)
Let's track these UX issues on a separate issue and work on it.

soul-wish:
> Would you mind creating a new issue with these points, and copy and paste this exploration there?

Sure! Created #668. Should we close/reject this PR? 
PS. Great work, @ansumanshah!

gaearon:
Let's close in favor of discussing https://github.com/facebook/react-devtools/issues/668 and maybe coming up with a more comprehensive solution.
If this isn't fruitful we can come back to this simpler fix.

Thank you so much for contributing! That writeup is very valuable.

soul-wish:
Thanks @gaearon and @jhen0409 for all your comments, suggestions and help! So welcoming community!

PR #640, opened 2017-04-27
[WIP] Move search bar to the top bar

gaearon:
It belongs better there. TODO: apply proper styling.

PR #641, opened 2017-04-27
Better filter UI

bvaughn:
Builds on top of PR #640 and...
* Adds inline search highlights so it's obvious at a glance that you're viewing a _filtered_ set of the data and not the full tree.
* Tidies up alignment of toolbar
* Makes focused state of search input look more like Chrome search and less aqua
* Make the cancel button a little more subtle and improve its vertical alignment
* Adds search icon to search input
* Removes regex checkbox in favor of auto-detection based on search text (eg "foo" vs "/foo")
* Added error handling for invalid RegExps

## Input states

Default | Focused | Invalid regex
--- | --- | ---
screen shot 2017-04-27 at 7 22 42 pm | screen shot 2017-04-27 at 7 22 59 pm | screen shot 2017-04-27 at 7 23 35 pm

## Plaintext search
plaintext search demo

## RegExp search
RegExp search demo

### Resizing
resize demo

#### Chrome
Chrome screenshot

#### Safari
Safari screenshot

#### Firefox
Firefox screenshot

#### Electron
Electron screenshot


gaearon:
Do you think it's worth making the input font size smaller and tweaking the clear button? Now that there are other indicators that search is active.

bvaughn:
Yup! I actually already shrunk the font size a little. I was fighting the urge to shrink it more because I know I have a tendency to make things too small. But let me try making it a little smaller. :grin:

bvaughn:
Updated demo video.

gaearon:
Wanna add some border radius to the input?

gaearon:
Also what do you think about moving input + search-related checkboxes to the right?

gaearon:
And maybe adding a magnifying glass watermark to the search icon a la

screen shot 2017-04-27 at 21 27 08

bvaughn:
> Also what do you think about moving input + search-related checkboxes to the right?

Thought about this! Seems great when they're all on the same line but a bit weird once they wrap (as they're doing in my video). Might still be worth doing though. Let me try it.

gaearon:
We can lay out them like now if they're small enough, but do a more sensible layout on larger screens.

bvaughn:
How about this?
resize-search-options

gaearon:
A few things still look off (not sure if fixable with small changes or not):

* I'm not sure if Highlight Search is more related to search or not.
* In general it is annoying that a niche feature like search by regex is taking so much space.

I wonder if these two search options would be better as toggleable buttons with icons, like in some IDEs. Not sure if it's going too far though. But I'm not very happy with how we spend this entire top bar on checkboxes people click very rarely.

bvaughn:
Yeah, I don't know. Highlight search _is_ related to the search but... to it's results, more-so than the criteria... so I thought to leave it on its own.

Agreed about space. How about just "RegExp" instead of "Use Regular Expressions"?

What about this:
search-input

iamdustan:
What about using a command like prefixing with `/` to switch to RegExp mode?

* No UI needed.
* Reasonable default experience.
* Power-user feature.
* Low-cost.

bvaughn:
Hm, initial reaction: That isn't very discoverable. (It's not how browser search or IDE search works.) And "[ ] RegExp" is pretty small now so it doesn't take up much space.

gaearon:
Could use a placeholder for this.

bvaughn:
Now the search input takes up way more space than the checkbox + input currently does though

bvaughn:
Let me whip it up quickly and see what it looks like though :smile:

gaearon:
We could also omit "components", it's pretty obvious what it searches for now that we have highlighting.

screen shot 2017-04-27 at 22 17 08

bvaughn:
I'm sold 😁 

regex-in-placeholder

gaearon:
If you feel like bikeshedding a little bit more, another random idea:

screen shot 2017-04-27 at 22 23 33

gaearon:
@bvaughn Anyway, I think your last version is best. Let's just rename "Trace React Updates" to "Highlight Updates".

bvaughn:
Thanks for the feedback folks!

type-ahead-demo-final

frontend/SearchPane.js
```diff
@@ -36,155 +33,27 @@ class SearchPane extends React.Component {
   _key: (evt: EventLike) => void;
```
gaearon: Should probably move this Flow annotation (and maybe rename this file?)
I didn't really completely clean this code up.

gaearon:
While not quite related, can we also debounce search input just a little bit?

gaearon:
It seems like `/` matches all components now. Would be nice to avoid this as I imagine it could be very slow to type.

bvaughn:
Good call. I'll add a short-circuit of the search text is equal to "/" 😁 

I find myself wanting to do a bit of cleanup here but I'd like to limit the footprint of this specific PR. Happy to add the debouncing change and the rename you mentioned above. Maybe we could cap it there and do follow up with another PR?

gaearon:
Agreed, but let's fix `/` here.

bvaughn:
Certainly!

frontend/SearchPane.js
bvaughn: Good call.

bvaughn:
Okay. Tidied up the old `SearchPane` and added better edge-case handling around "/" and "//" strings.

Maybe we can leave off the debounce for now. I think I might like to revisit how our search index and results ranking works in a subsequent PR anyway.

gaearon:
When I click "Highlight Search" the placeholder makes a weird change.
Can we completely remove this behavior? (I assume it was added together with the checkbox.)

screen shot 2017-04-27 at 23 56 06

bvaughn:
Oh. Yeah sure.

I only tested that feature briefly and I didn't notice the placeholder change.

gaearon:
A few other things (for follow up):

* Perhaps `/` along should be treated as an empty search input (no filtering) rather than "no results".
* I think we should change the filtering to only work at word boundaries. For example `to` should find `TodoItem` and `AddTodo`, but not `button`.

bvaughn:
> Perhaps / along should be treated as an empty search input (no filtering) rather than "no results".

Sure.

> I think we should change the filtering to only work at word boundaries. For example to should find TodoItem and AddTodo, but not button.

I'd rather do this as a follow-up.

gaearon:
Yea, totally, separately sounds good.

bvaughn:
slash-handling

gaearon:
Hmm, it doesn't let me fully type a regex. Try typing `/t[`. At this point it stays `/t`.

bvaughn:
Not at my keyboard at the moment but I bet I know the cause of this. Previous UI didn't try to convert the search term into a regex until you clicked the checkbox. Now we convert as you type if there's a forward slash at the start. I bet I need to try catch the conversion for temporarily-invalid regular expressions (like the one you've typed above). Maybe also add a red background color to the search input box to convey that the regular expression is incorrect. I'll take another look at it this evening. Nice catch exclamation!

bvaughn:
lol at the speech to text for writing out the word exclamation

bvaughn:
Hm...actually what you've described ("/" => "/t" => "/t[") works for me.

gaearon:
Weird, it doesn't for me. I'm testing in the plain shell on the default example (`shells/plain`). It seems like I can't type `[` at all if it's in regex mode.

bvaughn:
Curious if you can repro in master (after the "regex" checkbox is selected). I assume you would be able to.

bvaughn:
How about this for invalid regexps?

invalid-regexp

gaearon:
This works really well. The last issue I can spot is search icon in Firefox:

screen shot 2017-04-28 at 10 15 24

I’ll take a look at it.

bvaughn:
Weird! Did you pull before checking out Firefox? I tested it (and Safari) and left screenshots in the PR description yesterday.

gaearon:
Yea, I definitely pulled. Maybe a different version or something. 🤷‍♀️

gaearon:
It was just missing a `top: 0`.

bvaughn:
Weird! 🤷‍♂️ Worked for my Safari.

Glad you caught it!

Issue #490, opened 2017-02-07
Chrome extension icon should activate on sites using React


ogawa0071:
"react-devtools" Chrome Extension icon is unresponsive even on a site using React.
Why it turn on colorful icon and, Inform the page show up `react` tab in Chrome Developer Tools?
I think it update makes even better extension.

Vue.js Chrome Extension icon is responsive.
![image](https://cloud.githubusercontent.com/assets/3258736/22685275/0dcd682e-ed63-11e6-97aa-a6ae67db60ab.png)

gaearon:
Would you like to contribute this? It must be some new API Chrome added.

kp25:
Does any started working on this? I want to pick this up.
Never did much of open-source contribution, would love to contribute back. :-)

gaearon:
@kp25 Feel free to go ahead!

ogawa0071:
@gaearon @kp25 
Sorry to late response. I pointed out, but I have no time to tackle. 
I appreciate @kp25 seems to be working on it. Thank you very match!

kp25:
@ogawa0071 @gaearon haven't started yet; will be picking it up this weekend. Will ping you guys if any help needed.

kp25:
@gaearon found the fix for that, but I need new icon files that need to be showed when it is active; Would be helpful if someone can help me out with that. :-)

gaearon:
Hmm, shouldn't it just use our existing icon? It's green in the repository, so I assume it's Chrome painting it gray by default. There's probably some API to switch it back to colorful when the extension is active.

gaearon:
Our existing icons are here: https://github.com/facebook/react-devtools/tree/master/shells/chrome/icons

kp25:
@gaearon If we just gray it out, a user can't interact with the extension icon, right? 

Looking at https://github.com/vuejs/vue-devtools extension, they are using two sets of icons
default - grayed out ones
active - colored version

By clicking on default icon they are showing a message popup saying that "vue.js not detected". This feature is implemented using "browser action" rather than "page action"; Checked the docs, it suggests to use "browser action". 

Please take a look at the following and let me know which direction should we proceed.

gaearon:
I think we should do the same. 👍 
The extension was written a long time ago so it's likely best practices changed.
I like what Vue is doing.

kp25:
@gaearon cool. Can you ask someone to help me out with the new set of icons, all I need is two sets of icons - default & active one.

[...]

PR #642, opened 2017-04-28
Chrome extension icon should activate on sites using React #490

deepak-jacob:
Colour icon will show up on pages build with react.


gaearon:
This looks like a great start, thank you for jumping on this!

I don’t think we’ll want to use this specific detection strategy because we already should have enough information about whether React is on the page by checking if it attached to the hook we expose.

I’ll try to take it from here because I’m keen to get this out today.

deepak-jacob:
Great, thanks for that !

gaearon:
I added popups!

screen shot 2017-04-28 at 5 50 24 pm

screen shot 2017-04-28 at 5 50 09 pm

I hope this will help people figure out how to use the extension.

>Using page_action instead of browser_action so we don't need to have a gray and colour icon.

I had to revert that part because only `browser_action` lets us show a popup even when extension is inactive (which is important for people confused by the `file://` permission issue).

Overall I’m very grateful for you to showing how to do this, and pushing this change through. 👍

gaearon:
Going to get this in, hope the CI passes. I have some troubles running Flow locally but this should work.

deepak-jacob:
I thought do popup tonight ! but thanks for that... yeah definitely helpful

Issue #582, opened 2017-03-09
Electron shell data transfer is lossy


gaearon:
Since Electron shell uses JSON as transfer format, it loses some data, such as `undefined` values in objects, and sparsity of arrays. See my comments in https://github.com/facebook/react-devtools/pull/536#issuecomment-285235927.

I'm not sure how to solve this but it is confusing to have this disparity between shells.

gaearon:
Maybe we should use something like https://github.com/cognitect/transit-js instead of JSON as transport?

tomaszlakomygp:
transit-js is quite a lot of code (10K gzipped), maybe there's something smaller that can be used to achieve the same effect?

gaearon:
Does it really matter though, if it’s only inside DevTools?

tomaszlakomygp:
Not really, those are my web development habits speaking 😄

jaredly:
yeah, transit could be a good thing to do

davidfurlong:
@gaearon 
Took at a look at transit-js & using it here.
Potential issues with it:
- https://github.com/cognitect/transit-js/issues/10 (I.e. converts undefined to null)
- It converts Objects into its own type of objects, and parsing them back to plain JS objects requires nested calls of http://cognitect.github.io/transit-js/classes/transit.html#methods_transit.mapToObject 

Are there any alternatives people know of?

gaearon:
Interesting. I am not aware of alternatives but I'm fairly sure somebody must have solved this problem before.

davidfurlong:
Theres also http://msgpack.org/ but that seems to be more concerned with compression than with consistency across null/undefined/NaN

davidfurlong:
Okay I've modified the JSON2 parse and stringify functions to support undefined, NaN, Infinity, -Infinity. They were omitted from JSON because they are references to their global respectives (I think properties of the Number type) and hence aren't expressions, but references. I think disregarding the security problems/ spec will work for our use case.

https://github.com/davidfurlong/JSON-js

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

[...]

PR #654, opened 2017-04-29
Adds support for values: undefined, NaN, Infinity and -Infinity

### LINKED ISSUE #653: value: NaN being displayed as null
screen shot 2017-04-29 at 11 13 21
screen shot 2017-04-29 at 11 13 27

davidfurlong:
Fixes #653 #582


gaearon:
Hmm, we might want to be able to change that code (in order to fix more such bugs in the future etc). Are you absolutely sure there are no existing alternatives that are well tested?

If not it would be preferable to keep the source code here (in `packages`).

davidfurlong:
@gaearon I looked fairly thoroughly, though I might have not had the right keywords. 
I too am a bit worried about testing / compatibility. However I am not doing anything radically different than JSON2 already does with null values. Beyond splitting into its own namespace/different package, the whole diff is https://github.com/davidfurlong/JSON-js/commit/9772bd1fbda29e282f8e1536fbcd2c7f66b88780

As for keeping the source code in `packages` I dont think thats necessary & it may be confusing, as its a whole different utility, but I'm fine with that. I can add you as a contributor to the npm package/git repo if you'd like.

gaearon:
>As for keeping the source code in packages I dont think thats necessary & it may be confusing, as its a whole different utility, but I'm fine with that

If it's a core dependency it's fine for it to be here, it's easier to manage issues and do releases that way.

davidfurlong:
Okay I'm modifying my PR to include the code in `packages`, not excited by the amount of lint errors I'm getting 🏖 .

davidfurlong:
@gaearon added it to `packages`, but couldn't understand how you were locally installing the dependencies inside `packages`

tyleriguchi:
@davidfurlong hey I noticed that this is the same issue as the one in [redux-devtools-extension](https://github.com/zalmoxisus/redux-devtools-extension/issues/332). That uses a package called [jsan](https://github.com/kolodny/jsan), which does a very similar thing.

davidfurlong:
@tyleriguchi ah yes thats the sort of thing we were looking for. It does not support NaN, Infinity, -Infinity though. It does support functions & circular structures, and recreates Date Objects rather than leaving them as UTC strings, and recreates ES6 Maps & Sets. Its also more performant & yields smaller strings (u for undefined etc.). I think I could also submit a PR to JSAN to support NaN, Infinity, -Infinity values. @gaearon

gaearon:
I'm okay with using something more battle tested if its feature set works well for us (or if we can contribute to fill in the missing pieces).

davidfurlong:
If we swapped to the JSAN package would we still pull the package into `packages` ~ I put in a PR but have yet to hear anything back. I think we should use the JSAN code, but if the author is not going to maintain the package then I think its better to fork into this repo

zalmoxisus:
The feature is available in the upstream. Let me know if there's anything else we should add there. It's worth to mention that the lib is using WeakMap when available, which makes the serialization really fast, and it's production proved on Redux DevTools Extension for more than an year.

davidfurlong:
@zalmoxisus I am going to modify my PR this evening to use JSAN instead

davidfurlong:
Okay two things:

a) test/example/build doesn't seem to `gitignored`. Since I changed the tests, should I commit the built files
b) jsan has a bug where `parse(stringify(undefined))` is `{$jsan: "u"}`. I've reported it and will look at making a PR there before we merge this.

bvaughn:

Issue #648, opened 2017-04-28
Waiting for roots to load...


samit4me:
Today I saw a tweet by @gaearon *"[Anybody with Windows 10 who could test it?](https://twitter.com/dan_abramov/status/858033145614086145)"* and I have Windows 10, so I thought I'll give it a go. First I tried a couple of my own apps and all was good, the icon would light up and the react tab would appear for sites with react 👍

After that, I switched over to `https://<chatroom>.slack.com` and I noticed the icon was light up, so I opened the devtools and clicked the react tab and I was greeted with the following, so I clicked **click here** to reload but no change.
<img width="178" alt="slack" src="https://cloud.githubusercontent.com/assets/3248531/25546459/870db8c4-2ca6-11e7-8b2a-09ad854e398c.PNG">

I then relayed this to Dan, who asked me to open an issue 🎉 

Unfortunately, I do not work for Slack, nor do I have access to their codebase, so um yeah, not sure what good this issue can bring 😢. Maybe Slack are doing this on purpose? Not sure if there is an API to disable react devtools or if this is even wanted.

I did a quick search and found #354 and #547 which are reporting the same issue. In both of these issues, Electron was mentioned, so I spun up a clone of [electron-with-create-react-app](https://github.com/csepulv/electron-with-create-react-app) to test but the devtools worked as expected.

I apologize for reporting an issue without a minimal project reproducing the issue, but if it helps, I've got the same issue on the following:
* Win10 (v1607) + Chrome (v57)
* macOS (v10.12.3) + Chrome (v57)

gaearon:
Another data point: I think you’ll see this message **if React is on the page but no React components are mounted**. At least that’s my impression from why it fires up on https://facebook.github.io/react/tutorial/tutorial.html.

Regardless, we should change the message to say something like:

>React was detected on the page, but no components are currently mounted.
>If this seems wrong, you can try [reloading the inspector].

samit4me:
Okay thanks for that, after looking at the code it makes sense to fix this in #655. I'll push an update tonight as part of a small refactor.

MrEfrem:
I also have the same issue. But I have a mounted components on the page. Specifics of my case what I to inject React components in websites. I have opened this websites in the separate iframe.

samit4me:
Did you have a URL or some code you can share? Would be good to find the root cause.

MrEfrem:
No. I work on the private repo.

MrEfrem:
[...]

PR #655, opened 2017-04-30
Improve loading and error messages

### LINKED ISSUE #650: The React tab stays in DevTools even when switching to a website not using React
Was able to get the react-devtools stuck with the message "Looking for React..."

Steps to reproduce:
* Visit https://www.facebook.com/ in Chrome
* Open the Chrome devtools
* react-devtools icon should be light up and the react tab visible *(as expected)*
* Now click the react tab and it should work *(as expected)*
* Then change the URL to https://www.google.com/
* react-devtools icon "grayed out" *(as expected)*
* react tab is still visible and selected with the message "Looking for React..." ***(possibly an issue)***

Not sure what the correct behavior should be, but possibly:
* Hide the react tab and select the "default" tab in Chrome devtools *(if that's even possible)*, OR
* Provide a different message in the react tab, maybe something similar to what the icon shows when react is unavailable.

samit4me:
fixes #650 

With [`chrome.devtools`](https://developer.chrome.com/extensions/devtools_panels) you can conditionally create a panel and this is exactly what `react-devtools` does, see [main.js](https://github.com/facebook/react-devtools/blob/8667d793916a0b40a84db0ae2f47000aabf9645d/shells/chrome/src/main.js#L30) . Unfortunately, there is no API to remove a panel, so once the **React tab** has been created there is no way to remove it.

In v2.1.3 and maybe earlier, if you visit a website not using React, no **React tab** will show up *(as expected)*. If you then switch to a website using React *(in the same Chrome tab)* the **React tab** will appear *(as expected)*. If you then go back to the site without React *(in the same Chrome tab)* the **React tab** will still be visible and it shows the message "Looking for React...".

This PR changes that message to more closely align with the message from the `react-devtools` icon, see [`not-detected.html`](https://github.com/facebook/react-devtools/blob/60c3e1bfa3ce60a22f1de1f3d8ea2beda3df2a2e/shells/chrome/popups/not-detected.html). This is also similar to what the [`ember-inspector`](https://github.com/emberjs/ember-inspector/blob/1c34df9e8e7f521b380d09d6f7d3af15c650cc78/app/templates/components/not-detected.hbs#L23-L24) does.

[Prettier](https://github.com/prettier/prettier) was used to style the modified JSX.

### DIFF
--- a/frontend/Loading.js
+++ b/frontend/Loading.js
@@ -1,0 +1,24 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+'use strict';
+
+var React = require('react');
+var Message = require('./Message');
+
+function Loading() {
+  return (
+    <Message>
+      <h2>Looking for React...</h2>
+    </Message>
+  );
+}
+
+module.exports = Loading;
--- a/frontend/Message.js
+++ b/frontend/Message.js
@@ -1,0 +1,50 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+'use strict';
+
+var React = require('react');
+var { sansSerif } = require('./Themes/Fonts');
+
+import type { Theme } from './types';
+
+type Props = {
+  children?: any,
+};
+type Context = {
+  theme: Theme,
+};
+
+function Message(props: Props, context: Context) {
+  const { theme } = context;
+  return (
+    <div style={messageStyle(theme)}>
+      {props.children}
+    </div>
+  );
+}
+
+Message.contextTypes = {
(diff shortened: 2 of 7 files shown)


samit4me:
Not sure why travis-ci is failing the build, but I've made sure it works locally and it passes linting, testing and typechecking. Any help will be appreciated 👍

samit4me:
After doing some further testing, I noticed that if you navigate between pages with and without React *(with the **React tab** selected)* you can get the new "React not detected" message to show on pages that are using React. This is because React on the **inspectedWindow** has not finished loading.

The [`componentDidUpdate`](https://github.com/facebook/react-devtools/blob/8667d793916a0b40a84db0ae2f47000aabf9645d/frontend/Panel.js#L214-L223) method continually looks for React until it is found and this is why the "React not detected" message can show and then disappear when React is found.

The problem appears to be ***detecting when the page has finished loading***, if that can be figured out there will be no need to continually look for React.

If that's not possible, a quickfix would be to modify the message to read something like:
screen shot 2017-05-01 at 10 09 06 am

Keen to hear your thoughts!

gaearon:
Thanks for PR. Just a heads up—I’ll be busy with other things this week so it will take a me a little while to get back to this repo.

samit4me:
fixes #648 #650 

1. Improve message consistency between Firefox and Chrome.
1. Delay showing the "React not detected" message so the page has a chance to load.
 NEW - *"Looking for React..." > DevTools*
 **VS**
 OLD - *"Looking for React..." > "React was not detected on this page." > DevTools*

gaearon:
Thanks for the PR. I’m sorry I couldn’t get to reviewing it yet, and probably won’t be able for a couple of weeks.

PR #656, opened 2017-04-30
webpack to emit file with same name as in source dir

deepak-jacob:
Issue description :

basepath: /shells/chrome/
Seeing "build/inject.js" in  "manifest.json" will think "src/inject.js" is main entry point, as the file with that name exists.

But actually inject.js in manifest.json is src/GlobalHook.js which webpack emits as build/inject.js.

I feel like its little confusing, as I need to look into webpack config as well to see which file is mapped to inject.js.

Reject the PR if this is done on purpose or as a workaround.


deepak-jacob:
closing this PR as deepak-jacob:master is going diverge.

PR #657, opened 2017-04-30
Fix iterator check on dehydrate

shriah:
Fix iterator check on dehydrate

Fix styling of zero as value in data view


gaearon:
Thanks.

PR #660, opened 2017-05-01
[React Native] Keep only 3 decimals for box measurements

jhen0409:
This PR fixed the problem of long layout measurements:

## After

Chrome DevTools also keep 3 decimals for box measurements.

Also increased box width (200 -> 208) because It can keep `left` and `top` on the same line.


gaearon:
Thanks.

Issue #524, opened 2017-02-16
Refactor FireFox extension to use WebExtensions


SpaceK33z:
FireFox now supports the [WebExtensions API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions), which is almost exactly like the Chrome extension API.

Mozilla is planning to drop support for the existing extension API by the end of 2017 ([source](https://blog.mozilla.org/addons/2016/11/23/add-ons-in-2017/)).

So the FF extension will need to be rebuild to be compatible with the WebExtensions API before the end of 2017. However, this is good news, because the current Chrome version will perhaps work out-of-the-box and if not, it would need only small changes. So it shouldn't be too much work.

zalmoxisus:
`devtools.panels` are not supported yet. See http://arewewebextensionsyet.com/#devtools.panels and https://bugzilla.mozilla.org/show_bug.cgi?id=1300587. [Some progress has been made recently](https://reviewboard.mozilla.org/r/76818/diff/13/), but I wasn't able to create a panel in FF Nightly yet.

SpaceK33z:
@zalmoxisus ah, that's a shame. Still might be worth to keep this open to monitor this.

SpaceK33z:
If there are questions about the WebExtensions API or missing features, feel free to ask @rob--w. He's a colleague of mine and contributed much to the WebExtensions API as Mozilla volunteer.

clarkbw:
[bug 1300587](https://bugzilla.mozilla.org/show_bug.cgi?id=1300587) has been fixed and is on target for Firefox 54 which goes live to release on June 13th.  Now is probably a good time to get started on this work.  Let me know how I can help.

clarkbw:
I was checking this out today on Firefox 54 (DevEdition) and things mostly work https://twitter.com/clarkbw/status/857979112119349249

However I'm not sure what would be the best way to proceed. 

@gaearon do you have any thoughts?

I'm able to run the Chrome extension as is in Firefox and it would likely take some modifications (still need to investigate them all) to make sure things look normal and work correctly.  I'm wondering about how we handle any differences continuing down the line.

We could:

A shared source directory like `webextensions` for common code and then the `chrome` and `firefox` directories remain as light wrappers?

Or (my preference) converge on the `chrome` directory and do some of the fixes directly in there.  This hopefully leads to less divergence.  We could possibly rename that directory to `webextension` since it does also work in Opera but it doesn't matter too much to me.

Or some other option...

[...]

PR #663, opened 2017-05-01
Initial set of changes required to get Firefox WebExtension working

clarkbw:
This is the first PR required to merge Firefox and Chrome shells as described in #524 by merging the CSS properties used in both browsers.  The next PR will need to look into the `manifest.json` changes required to maintain the minimum Firefox version required to run this WebExtension.

Without this change Firefox allows the content to expand horizontally beyond the panel to accommodate content.  Chrome appears to continue to behave in the same way as it currently does with this change.

### DIFF
--- a/shells/chrome/main.html
+++ b/shells/chrome/main.html
@@ -1,6 +1,7 @@
 <!doctype html>
 <html>
     <head>
+        <meta charset="utf-8">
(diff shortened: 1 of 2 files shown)


gaearon:
This looks fine, thanks!

PR #664, opened 2017-05-01
Use fuzzy search in a search input field

olegafx:
Component: Search input in `SettingsPane`.
Proposal: allow searching nodes by partial matches.

Current behaviour:
Input: `Higher`. Output: `<HigherOrderComponent> list`
Input: `hoc`. Output: `no results`
Input: `hdeon`. Output: `no results`

Proposed behaviour:
Input: `Higher`. Output: `<HigherOrderComponent> list`
Input: `hoc`. Output: `<HigherOrderComponent> list`
Input: `hdeon`. Output: `<HigherOrderComponent> list`

Changes results highlighting:
`hoc` => <**H**igher**O**rder**C**omponent>

This fuzzy search algo uses simple match approach (no weights for matches applied).

My questions:
1. Should I replace all `this.searchRoots` filtering with `nodeMatchesText`? See my comment in a code section.
2. Is it okay to output `<span>` for every char in a `Node.name`?
3. `flow check` outputs error about code I didn't touch. What should I do about it?

### DIFF
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -15,6 +15,7 @@ var React = require('react');
 var assign = require('object-assign');
 var decorate = require('./decorate');
 var Props = require('./Props');
+var SearchUtils = require('./SearchUtils');
 
 import type {Map} from 'immutable';
 
@@ -25,6 +26,7 @@ type PropsType = {
   depth: number,
   isBottomTagHovered: boolean,
   isBottomTagSelected: boolean,
+  searchText: ?string,
   searchRegExp: ?RegExp,
   wrappedChildren: ?Array<any>,
   onHover: (isHovered: boolean) => void,
@@ -267,14 +269,31 @@ class Node extends React.Component {
     );
 
     var name = node.get('name') + '';
+    var searchText = this.props.searchText;
     var searchRegExp = this.props.searchRegExp;
 
     // If the user's filtering then highlight search terms in the tag name.
     // This will serve as a visual reminder that the visible tree is filtered.
-    if (searchRegExp) {
+    var shouldSearchUseRegex = SearchUtils.shouldSearchUseRegex(searchText);
+    var pieces = [];
+
+    if (!shouldSearchUseRegex && searchText) {
+      var fuzzyMatches = SearchUtils.getFuzzyMatches(searchText, name);
+      pieces = name.split('').map((item, itemIndex) => <span
+          key={itemIndex}
+          style={fuzzyMatches.includes(itemIndex) ? styles.tagNameHighlight : null}
+        >
+          {item}
+        </span>
+      );
+
+      name = <span>{pieces}</span>;
+    }
+
+    if (shouldSearchUseRegex && searchRegExp) {
       var unmatched = name.split(searchRegExp);
       var matched = name.match(searchRegExp);
-      var pieces = [
+      pieces = [
         <span key={0}>{unmatched.shift()}</span>,
       ];
       while (unmatched.length > 0) {
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -231,9 +231,9 @@ class Store extends EventEmitter {
         this.searchRoots = this.searchRoots
           .filter(item => {
             var node = this.get(item);
-            return (node.get('name') && node.get('name').toLowerCase().indexOf(needle) !== -1) ||
-              (node.get('text') && node.get('text').toLowerCase().indexOf(needle) !== -1) ||
-              (typeof node.get('children') === 'string' && node.get('children').toLowerCase().indexOf(needle) !== -1);
+            return (node.get('name') && SearchUtils.isFuzzyMatch(needle, node.get('name'))) ||
+              (node.get('text') && SearchUtils.isFuzzyMatch(needle, node.get('text'))) ||
+              (typeof node.get('children') === 'string' && SearchUtils.isFuzzyMatch(needle, node.get('children')));
           });
       } else {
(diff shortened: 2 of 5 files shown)


frontend/Node.js
```diff
@@ -267,14 +269,31 @@ class Node extends React.Component {
...
 
     // If the user's filtering then highlight search terms in the tag name.
     // This will serve as a visual reminder that the visible tree is filtered.
-    if (searchRegExp) {
+    var shouldSearchUseRegex = SearchUtils.shouldSearchUseRegex(searchText);
+    var pieces = [];
+
+    if (!shouldSearchUseRegex && searchText) {
+      var fuzzyMatches = SearchUtils.getFuzzyMatches(searchText, name);
+      pieces = name.split('').map((item, itemIndex) => <span
```
olegafx: Is it okay to output `<span>` for every char?

frontend/Store.js
```diff
@@ -231,9 +231,9 @@ class Store extends EventEmitter {
         this.searchRoots = this.searchRoots
           .filter(item => {
             var node = this.get(item);
-            return (node.get('name') && node.get('name').toLowerCase().indexOf(needle) !== -1) ||
-              (node.get('text') && node.get('text').toLowerCase().indexOf(needle) !== -1) ||
-              (typeof node.get('children') === 'string' && node.get('children').toLowerCase().indexOf(needle) !== -1);
+            return (node.get('name') && SearchUtils.isFuzzyMatch(needle, node.get('name'))) ||
```
olegafx: Should I replace it with `nodeMatchesText `?

gaearon:
@bvaughn You might want to look at this!

frontend/Node.js
bvaughn: My gut: This is probably not a good idea, no.

bvaughn:
I have some UX concerns with the highlighting behavior in this branch.

In this case, I would expect all 5 of the "a"s to be highlighted, rather than just the first one:
screen shot 2017-05-03 at 3 51 12 pm

In this case, I would expect "AnimatedComp" to be highlighted, since that's the full fuzzy-search match:
screen shot 2017-05-03 at 3 52 32 pm

Compare to the current highlighting behavior of master:
screen shot 2017-05-03 at 3 53 34 pm
screen shot 2017-05-03 at 3 54 56 pm

olegafx:
@bvaughn my solution was inspired by Sublime Text's file search. Here is how it looks in Sublime:

First case with `a`. I think that it's better to show only first `a` because: 1) highlighting all `a`s is confusing 2) it's better for performance (exit from a search function as soon as possible)
__code_sublime-find-test_animatedcomponent_ _sublime-find-test

I find it very useful to see **why** I got these search results. Highlighting only found symbols really helps me to understand it.
__code_sublime-find-test_animatedcomponent_ _sublime-find-test 2

bvaughn:
The biggest perf-limiting factor at the moment may be that we aren't windowing the nodes we render- but it looks like the new fuzzy search is a bit slower than what's in master. I added some timing logging to the `changeSearch` method. Comparing master to this branch, searching for single letters (so the result set should be the same), average search time went up from ~11.89ms to ~15.04ms for a medium/small test application and from ~17.88ms to ~25.39ms for a larger application.

bvaughn:
Those are reasonable arguments for the UI behavior in this branch. 👍

olegafx:
@bvaughn what do you mean by "windowing the nodes"?

It's bad that we should call our function twice – 1st time for a nodes filtering, 2nd time for a name highlighting.

And about a 1 symbol entered. Maybe we should use `indexOf` for that case?

bvaughn:
> What do you mean by "windowing the nodes"?

Ah, sorry. I meant limiting the number of `Node` elements we render. There's no need to render _all_ of them so long as we render enough to fill the viewable portion of devtools.

> It's bad that we should call our function twice – 1st time for a nodes filtering, 2nd time for a name highlighting.

We call it once to _filter_ nodes (true/false match) and another time to _highlight_ text (splitting it into pieces and mapping them to styles). This seems reasonable to me since those 2 are different operations. Windowing would limit the number of times we made the second call.

> And about a 1 symbol entered. Maybe we should use indexOf for that case?

I wouldn't expect that to make _much_ of a difference vs your current `needleIdx < needle.length` check?

olegafx:
@bvaughn what we're gonna do with this PR?

bvaughn:
Last round of back and forth was me saying,

> Since we're gathering matches both times, we are doing unnecessary work. Maybe we should split that up to avoid building a matches array we aren't using during the filter phase? Or maybe we should attach match metadata to filtered nodes? 

Was kind of waiting to hear back from you and what you thought about that. Also I got distracted by a rather big refactor of the UI/styles for theming support.

olegafx:
@bvaughn I think that 4–8ms diff is acceptable. In my opinion, the search operation is not frequently used. If someone will notice the difference, we will return to perf optimizations.

bvaughn:
I'm still a little unsure how I feel about this change, given the performance impact and the fact that devtools was slow to begin with in this area. I'd like to think on it a bit longer and give others a chance to comment if they have any feelings one way or another.

Didn't want you to think that I'd forgotten about it. 😄

olegafx:
@bvaughn Should I close it?

bvaughn:
No harm in leaving it open. I never decided how I felt about it and I've been pulled to work on other projects for the time being. Sorry to leave you hanging!

bvaughn:

Issue #562, opened 2017-02-21
Show whether React is running as production build


mnquintana:
Depending on the module bundler you're using with React, it can be really difficult to verify that changing your `NODE_ENV`  to enable the production build is indeed working correctly, without resorting to more indirect checks (e.g. comparing the file size of the bundle outputs, or trying to call one of the `PropTypes` functions to see if it throws). It would be super helpful if the devtools could display an indicator showing whether React was running in production or development mode.

gaearon:
Great idea. This should be a part of https://github.com/facebook/react-devtools/issues/490.

gaearon:
I think we can do it now. I’m not sure what’s the best way to detect a DEV version but happy to hear proposals. Would be nice if it worked without changes to React itself.

gaearon:
Fixed in https://github.com/facebook/react-devtools/pull/667.

PR #667, opened 2017-05-02
Detect DEV version of React

gaearon:
Implements https://github.com/facebook/react-devtools/issues/562 as suggested by @mnquintana.

This should detect all supported React versions, with the exception of React 16 (Fiber) for which we’ll need to add explicit support (luckily we can do this easily since it’s not released yet).

It should detect all combinations (no envify + no uglify, envify + no uglify, no envify + uglify). It works by introspecting the `Mount._renderNewRootComponent` source which is present in all React DOM versions, and gives us enough information thanks to a warning (after 15.1) and to the fact that it used to be wrapped into a ReactPerf wrapper in DEV (before 15.1).

What's missing:

* [x] Check for outdated version of React
* [x] Popup for outdated version of React
* [x] A flag on React side to support Fiber
* [x] Fiber support (once that flag on React side is merged)
* [x] Actually test 0.13, 0.14, 15.0, 15.1, 15.2, and 15.4

It looks like this.

screen shot 2017-05-02 at 7 08 27 pm
screen shot 2017-05-02 at 7 08 34 pm


bvaughn:
How's this?

screen shot 2017-05-02 at 11 36 25 am

Will push as soon as I have permission to.

gaearon:
Looks good. Can you push now?

bvaughn:
Thanks! Pushed.

gaearon:
I did some testing, and it looks good. Going to ship.

enapupe:
Awesome!
Looking god except for that :white_check_mark:  icon that is missing.

ss
```
Google Chrome	56.0.2924.76 (Official Build) (64-bit)
Revision	cc6576b53870d39ddafd3574d951ff36d5f7218e-refs/branch-heads/2924@{#860}
OS	Linux 
JavaScript	V8 5.6.326.42
Flash	25.0.0.148 /home/enapupe/.config/google-chrome/PepperFlash/25.0.0.148/libpepflashplayer.so
User Agent	Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/56.0.2924.76 Safari/537.36

Command Line	/opt/google/chrome/google-chrome --flag-switches-begin --flag-switches-end
Executable Path	/opt/google/chrome/google-chrome
```

bvaughn:
@enapupe I believe this indicates that your OS is missing the U+2705 character in its default sans-serif font. (Can you see it [here](http://graphemica.com/%E2%9C%85)?)

enapupe:
Kind of??

symbol

bvaughn:
Ah, rats. I didn't notice that page was using a different font.
What if you change the element's font to `font-family: sans-serif;` in Chrome's inspector?

enapupe:
Yeah I guess not. That on the image is likely to be provided by a webfont

enapupe:
So yeah, my system doesn't have it. Sorry for the  trouble. Thanks for that!

bvaughn:
No problem at all. I'm still kind of learning my way around this project too so I wasn't sure anyway 😁

gaearon:
Maybe we can just replace this with an equivalent royalty free SVG or something.

bvaughn:
I guess so. The icons aren't really necessary to be honest. The text is sufficient. (IMO the popups aren't meant to be pretty, just practical.)

PR #669, opened 2017-05-02
[WIP] Add dark mode

Kerumen:
This is a WIP PR to track the progress and discuss about the implementation as I'm not quite sure.

I've added the toggle and I have the color scheme in mind (I will borrow the Firefox dev edition as I find it less aggressive than the Chrome's dark one).

I have concerns on how to implement that however. I can listen to the toggle in each rendering component and a sort of `darkMode && styles.dark` in each `style` but it seems tedious and redundant.

I would like a more general way. Like a theme? https://github.com/markdalgleish/react-themeable?

In pure CSS I would add the `dark` class on the root element and style all the children with this. This is not possible with CSS in JS.

Any ideas?

### DIFF
--- a/frontend/SettingsPane.js
+++ b/frontend/SettingsPane.js
@@ -9,6 +9,7 @@
  */
 'use strict';
 
+var DarkModeFrontendControl = require('../plugins/DarkMode/DarkModeFrontendControl');
 var TraceUpdatesFrontendControl = require('../plugins/TraceUpdates/TraceUpdatesFrontendControl');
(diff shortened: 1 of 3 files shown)


ansumanshah:
I believe we can go with `styled-components` that has an inbuilt support for theming and comes with all the benefits of css as well.

Kerumen:
@bvaughn got the subject covered in https://github.com/facebook/react-devtools/commit/90896c9d63a0cd686cfe5e39bdd5b1652810d8c4. Well done!

bvaughn:
Thanks!

PR #670, opened 2017-05-02
Added an update nudge to the CLI version of react-devtools

dichuvichkin:
Now there will be a message in the console about a new version


gaearon:
Can you also turn off Electron debug messages somehow?
User shouldn't see this:

screen shot 2017-05-04 at 2 41 57 pm

packages/react-devtools/app.js
```diff
@@ -8,9 +8,23 @@
...
+
+//check an update
+var notifier = updateNotifier({
+  pkg,
+  updateCheckInterval: 0,
+});
+
+if (notifier.update) {
+  console.log('Update available ' + notifier.update.current + ' => ' + notifier.update.latest);
+  console.log('Run npm install -g react-devtools to update');
```
gaearon: Can you please use cyan color for the command? `chalk.cyan`

dichuvichkin:
@gaearon hmmm... I've added a cyan color to the command, but didn't see any debug messages

gaearon:
Thanks again for PR. I went with https://github.com/facebook/react-devtools/pull/695 instead because it had less code and already included the default logic for showing the box. I appreciate your effort though!

Issue #668, opened 2017-05-02
React Native Style Editor UX issues


soul-wish:
Hi, everyone! 
As a follow-up to conversation in #627 

---

1. There are some inconsistencies in style editor UI (some paddings, input's layout etc.). But I've checked other issues/PR's in the repo and saw that all these issues will be fixed in the nearest future;

---

2. It would be really cool to be able to navigate to the style value by pressing tab key on the keyboard. Now there is only one input field for the new style key, and only after blur event second field appears. So right now there is no way to add new styles by keyboard only - you should always use your touchpad or mouse;

![blur](https://cloud.githubusercontent.com/assets/1968098/25543920/0df7332e-2c61-11e7-8aa2-35e22c87a0a4.gif)

---

3. If I got it correctly, `RCTView` is always a child of `View` so it inherits all its styles (correct me if I'm wrong). And I've noticed that you can not simply edit `RCTView` styles because they will be reset automatically from the parent `View`. So, maybe it would be better just to display all these styles and not allow to change them (readonly fields, or even better - some type of list with style's keys and values);

---

4. As for the `View`s styles I've found few strange issues: 
    - I can not delete original styles (they just keep reappearing, but maybe it should work in this way);
![original-styles](https://cloud.githubusercontent.com/assets/1968098/25543945/22defc86-2c61-11e7-960b-590812b134a3.gif)

    - But I can delete the styles that were added by me manually (with a small fix in this PR);
![new-styles](https://cloud.githubusercontent.com/assets/1968098/25543979/3dc8335a-2c61-11e7-9121-a2900782c665.gif)

    ---

    - I can not change the keys of the styles (but I can change their values);
![original-styles-values-keys](https://cloud.githubusercontent.com/assets/1968098/25543987/45660920-2c61-11e7-8127-5efc1a7e0243.gif)

    ---

    - If I try to change the key of the original style it will be added as a new style record (key->value) at the bottom of styles editing form;
![original-styles-edit-key](https://cloud.githubusercontent.com/assets/1968098/25543993/4fd625b6-2c61-11e7-87a4-d7da121bdbc1.gif)

    ---

5. I'm not sure if it is possible to not add new style if the key is incorrect? Right now it works in this way:

![incorrect-keys](https://cloud.githubusercontent.com/assets/1968098/25543998/558a7804-2c61-11e7-96f9-24628fc13445.gif)

---

PS. I apologize in case I miss some important points or concepts (as I said earlier - I have zero experience in RN) ;)

[...]

PR #658, opened 2017-04-30
Fix RN Styles Editor

ansumanshah:
#624 #627 
@gaearon Done some basic UI improvement for now
- [ ] Add autosize input as we can see in chrome style editor

Please put forward any other suggestion for the UI
Will try and work with @soul-wish for the UX improvements he suggested [here](https://github.com/facebook/react-devtools/pull/635)

hksbhnzqy9

PR #671, opened 2017-05-02
Fix Native Styles Editor

ansumanshah:
ReOpen #658
Todo #668

- [x] AutoSize

- [x] Fix Click to edit

- [x] Fix Zero values

- [x] Tab Navigation

- [x] Font Weight Fix

- [x] Box Layout Overlay


gaearon:
Wow, nice work. 
A few comments.

I think this style is a bit confusing:

screen shot 2017-05-04 at 2 35 30 pm

Maybe if we remove the colon in this case it would be better?

Zero values seem to render as missing:

screen shot 2017-05-04 at 2 36 55 pm

I don't like how the font gets super thin when I edit a value. Can we keep its weight?

screen shot 2017-05-04 at 2 37 16 pm

Navigating with tabs no longer works. We should ensure Tab and Shift+Tab cycle focus through the fields, just like they do in Chrome DevTools.

If I get the window really narrow, the box layout bleeds:

screen shot 2017-05-04 at 2 38 36 pm

It is odd how editing the key moves the value to the right:

screen shot 2017-05-04 at 2 39 05 pm

This is especially annoying because if I click on the value while I’m editing the key, it doesn’t enter the edit mode for the value.

Do you think you can address these concerns and make it more similar to how Chrome handles it?

ansumanshah:
Cool on it.
@gaearon have we finalised on the styling library that we will be using?

gaearon:
I'd like to keep the existing style system in this PR. We can change how we do styles but it's not related to this PR.

ansumanshah:
fzdaozah4a

Just fixed the UI bit.. have not worked on delete and rename of styles, I think we should deal with them in a separate issue.

gaearon:
This is a great start, thanks. Can you look at other issues with style editor next?

For example, it's confusing that RTCView styles are not editable. They should either appear as readonly, or we should completely hide them, and only show styles on View/Text instead.

ansumanshah:
Awesome, my first PR got merged yay!!!
@gaearon I think we should completely hide them from RCTView
Looking into more issues will get back.

Issue #672, opened 2017-05-02
Differentiate between lack of envification and lack of minification


gaearon:
Currently we show the DEV icon and popup both if the code isn't envified, and if it's not uglified.
https://github.com/facebook/react-devtools/pull/667

I believe we should still highlight both issues, but we should probably change the popup text depending on which issue we hit.

Envifying is important to switch React to a faster mode. Uglifying is important both for shipping less bytes and for cutting out the DEV branches that can cause deopts.

There are comments in https://github.com/facebook/react-devtools/pull/667 to guide you through the permutations. We should probably split `'development'` result into `'no-envify'` and `'no-uglify'` (in that order), and show separate popups with slightly different messages.

PR #674, opened 2017-05-03
Add detection for unminified builds

gaearon:
Fixes https://github.com/facebook/react-devtools/issues/672.
This complements https://github.com/facebook/react-devtools/pull/667 with the ability to differentiate development and unminified builds.

When we're sure it's a development build, we show this:

screen shot 2017-05-03 at 12 23 21

When we're less sure, we show this:

screen shot 2017-05-03 at 12 23 10

Both are bad (unminified build is very large) so I kept the red icon.

I also added minification instructions to the React docs in https://github.com/facebook/react/pull/9592.


gaearon:
Seems to work.

Issue #676, opened 2017-05-03
Proposal: Dim Components with Parens in Name


gaearon:
Inspired by https://github.com/facebook/react-devtools/pull/503, I want to propose a super simple heuristic for making HOCs less prominent.

How about we just dim the components in the tree that have `(` or `)` in their names? These are always HOCs, as people commonly use `X(Y)` convention like `Connect(App)`, `Relay(FeedItem)`, or `withStyles(Button)`.  You can’t accidentally put parens into the name so it’s always intentional. HOCs that don’t adhere to this convention could easily adapt to it.

We wouldn’t completely *hide* the HOCs (at least, I’d like to start without it) because it obscures the data flow. But I think even making them dim in the tree would make a visual difference. Does anyone want to prototype this?

I don’t know if dimming will be helpful in practice or not but I’d like to experiment with it. If not, then we could add a “Hide Wrappers” checkbox that also uses `(` and `)` as a heuristic.

mattvagni:
Can I take this on? I think this could feel really nice and it seems like a good first bug to get to know how the dev tools are built.

neeharv:
I'd like to prototype something out, if no one else has started work on this. 

I like the idea of dimming it out, but I'm worried for people new to react dev tools it might look "disabled". Let me try a few styles out and send screenshots.

gaearon:
Seems like @mattvagni got there first!

gaearon:
(But seriously, feel free to prototype and collaborate. I’m not sure if this is viable, but it may work if someone figures out a good balance.)

mattvagni:
Initial impressions is that it works pretty well just dimmed out.

<img width="616" alt="screen shot 2017-05-03 at 5 44 29 pm" src="https://cloud.githubusercontent.com/assets/1372762/25671505/32844f9a-3028-11e7-9fad-7a36d799cb69.png">

However it would need some more consideration when searching:
<img width="1440" alt="screen shot 2017-05-03 at 5 48 08 pm" src="https://cloud.githubusercontent.com/assets/1372762/25671657/b98c96aa-3028-11e7-8b5f-5f46c3598028.png">

NOTE: Currently just styling the `HigherOrderComponent` vs. doing it properly and only matching components with `()` since I haven't yet found out how to change the pretend application.

arturmuller:
I would love a "Hide Wrappers" checkbox!

iamstarkov:
related #604

bvaughn:
[...]

PR #677, opened 2017-05-03
Dim component tags that are HOCs

mattvagni:
### Description:
In issue: https://github.com/facebook/react-devtools/issues/676

In order to test this behavior I updated the example application to set the display name on it's `HigherOrderComponent` to match the common pattern described in the issue above. 

### Screenshots
By default this PR dims components in the tree view that are considered to be higher order components (have a `()` in their displayName):

screen shot 2017-05-29 at 7 23 44 pm

When a component is selected however it is not dimmed. 

screen shot 2017-05-29 at 7 24 05 pm

Also when a component is selected, the components 'close tag' is **not** dimmed. This makes it easier to determine what component you've selected in cases like this where a few HOCs are nested.

screen shot 2017-05-29 at 7 24 14 pm

It is also not dimmed when shown in the search results:

screen shot 2017-05-29 at 7 24 26 pm


gaearon:
Thanks for PR! Going to review this in a week or so when I get some time to try it on our codebases.

mattvagni:
Thanks @gaearon. Looking forward to having my first 'real' open source contribution in the wild ;)

mattvagni:
@gaearon Would you still like to go ahead with this? In which case I'll try and resolve the merge conflicts locally so I can test my changes still work.

mattvagni:
@gaearon Updated. 

Since I first made this PR support for themes seems to have been appeared 🙈 so I'm not sure whether HOC's should just be dimmed using opacity or whether they should use a theme variable. 

I have, for now, done the former. If you could have a look and advise that would be great. Thanks

bvaughn:
Hey! This is a pretty cool idea. I like showing HOCs as a different color.

I'm not sure about `opacity: 0.5` though. I think the contrast is too low for some of the themes (eg apathy, flat, firefox default, etc) and as a result, it's not readable:
screen shot 2017-06-06 at 3 34 18 pm
screen shot 2017-06-06 at 3 34 25 pm
screen shot 2017-06-06 at 3 34 29 pm

Maybe we should use (or introduce) a new theme color for it. Some of the themes are kind of low-contrast by default and there's been a little mild criticism for this. I think it's okay but I'm hesitant to reduce contrast further.

Alternately perhaps we could color the triangle/arrow on the left and/or the opening/closing brackets a differently to make HOCs stand out. Thoughts?

gaearon:
>Alternately perhaps we could color the triangle/arrow on the left

We already do this for stateful components, and I don't think it works. From my observation even people who used DevTools for a long time had no idea what a red triangle means (or even notice it changes color).

bvaughn:
Yeah that's fair. Maybe we need some new iconography to differentiate between stateful, HOC, and other. I could take a stab at creating some if there's interest.

mattvagni:
@gaearon @bvaughn Some additional ways to potentially go:
- We reduce/increase the opacity to 0.8 or something more subtle.
- We hide the behavior behind an opt-in checkbox (not sure if it's common enough for this to be a good idea, especially for people just starting to use React it might be confusing. Any solution that makes HOCs seem **too** special is probably unfriendly for the first-time react user)
- We lighten the color and ensure there is still enough contrast visually using something like [chroma-js](https://gka.github.io/chroma.js/)

Thoughts?

bvaughn:
Each has downsides.

> We reduce/increase the opacity to 0.8 or something more subtle.

I tried this FWIW. The closer you get to 1 the less HOCs are visually distinguishable in a meaningful way. The further you get from 1 the more chance they're not high enough contrast for the background.

> We hide the behavior behind an opt-in checkbox (not sure if it's common enough for this to be a good idea, especially for people just starting to use React it might be confusing. Any solution that makes HOCs seem too special is probably unfriendly for the first-time react user)

I think this would feel clunky.

> We lighten the color and ensure there is still enough contrast visually using something like chroma-js

This is basically a variant of opacity, with the same downsides.

mattvagni:
yeah.. good points.

Adding a theme variable would definitely solve the contrast issues but that won't address the 'understanding' issue of "Why is that component a different color?". Then again, not sure this has to be solved as part of this.

Shall I give this a try (incl. adding it to the `Preview` component)?

bvaughn:
Up to you Matt. May be worth trying and seeing. If we can't get the color to work I could also try to play with some icons.

mattvagni:
@bvaughn Hey 👋 I've now implemented the 'dimming' of HOCs 'properly' by adding a new theme variable. 

I'd love some guidance/review on the colors I've chosen for each theme. I chose colors which *I think* work well but it's super subjective. Maybe you can have a look?

As part of this, while going through and picking a new color for each them I found the color picker quite fiddly due to how small it was so I made it slightly larger.

Also, since the label for the new theme color was longer than the other I increased the width of all the `<ColorInput/>` component. 

Let me know what you think.

bvaughn:
Haven't had a chance to look at the colors yet. Will try to look today! Did have a quick thought about this point though:

> Also, since the label for the new theme color was longer than the other I increased the width of all the `ColorInput` components.

I'd prefer we shorter the label (eg "HOCs" or maybe just "Higher Order") rather than increase the width of the field by 2 rems. What are your thoughts?

frontend/Themes/Themes.js
```diff
@@ -224,6 +231,7 @@ const FirefoxFirebug: Theme = {
   special05: '#0000ff',
   special06: '#000062',
   special07: '#0000ff',
+  special08: '#0000ff',
```
bvaughn: Mistake? Same as host component color.

bvaughn:
Just looked through all of the newly-added colors and I'm still torn about readability vs noticeability. The HOCs do look a _little_ different but I don't think it's super noticeable. I'm not sure it's worth the contrast trade-off. ☹️ Guess I need to think about it more.

mattvagni:
@bvaughn I kind of agree about the contrast trade-off maybe not being 'worth it'. 

One idea I had is to make HOCs the same color as normal tags such as <div>s :

**Pro**: it's not a new color, no additional hit on contrast/readability
**Cons**: HOCs are still components, coloring them the same was as normal tags (such as <div/>) could be quite confusing/misleading to the uninitiated.

In the meantime I've changed the input colors back to what they were and renamed the new color to have a shorter name.

In any case... let me know how and if you'd like to proceed with this.

frontend/Themes/Themes.js
mattvagni: No this was intentional, this theme is incredibly color-poor (I assumed for accessibility) so I thought it would make sense to not differentiate HOCs on this one. Kind of weird I know.. but looking at the theme it made sense to me ?

bvaughn:
I'm on the fence still. Hope it's okay if I ponder on it a while longer in the hopes that something new occurs to me, or I feel more strongly one way or another. 😄

mattvagni:
@bvaughn Sorry for the late reply! I agree though, shall I close this for now and we can re-open if we come up with something?

bvaughn:
No worries, of course 😄

Yeah, maybe that's a good short term solution. I'm still on the fence. 😅

PR #678, opened 2017-05-03
Prefix cheap-module-eval-source-map with # to avoid warnings

clarkbw:
Changes the source map plugin from the deprecated format `@` for declaring
the sourceMappingURL to the new format with the `#`

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Deprecated_source_map_pragma

And: https://developers.google.com/web/updates/2013/06/sourceMappingURL-and-sourceURL-syntax-changed

Avoids these warnings in Firefox.  But I guess Chrome doesn't log these in the console anymore.

screen shot 2017-05-03 at 11 51 55 am

Either way, it produces the same source map as before but with the syntax changed.

### DIFF
--- a/shells/chrome/webpack.backend.js
+++ b/shells/chrome/webpack.backend.js
@@ -14,7 +14,7 @@ var webpack = require('webpack');
 var __DEV__ = process.env.NODE_ENV !== 'production';
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks.

PR #685, opened 2017-05-04
Use sender.url instead of sender.tab.url

### LINKED ISSUE #679: sender.tab.url is undefined in Firefox
This is odd.

The [Tab object](https://developer.chrome.com/extensions/tabs#type-Tab) inside the [MessageSender](https://developer.chrome.com/extensions/runtime#type-MessageSender) object returned from the [chrome.runtime.onMessage](https://developer.chrome.com/extensions/runtime#event-onMessage) event shouldn't include a URL unless the extension requested the `tabs` permission.  The MessageSender includes a URL, but that is a different value than the tab URL.

According to the docs for the Tab object URL attribute:

> The URL the tab is displaying. This property is only present if the extension's manifest includes the "tabs" permission. 

So I'm not sure why this code is working in Chrome right now as it is.  It doesn't work in Firefox because the URL attribute is undefined.  I'm not sure if this is a bug in Chrome or an intentional thing because once you have content script permission you'll know the URL through other means.

The issue is that [sender.tab.url](https://github.com/facebook/react-devtools/blob/master/shells/chrome/src/background.js#L79) shouldn't exist and it doesn't exist when run in Firefox.  However it does exist in Chrome.

I've found two fixes so far:

* Update the permissions to include `tabs` https://github.com/facebook/react-devtools/compare/master...clarkbw:request-tabs-permission
  * Downside is that all existing users will have to confirm the update of permissions
  * Upside is that according to the docs this is what is required
* Use the `sender.url` attribute instead https://github.com/facebook/react-devtools/compare/master...clarkbw:sender-url
  * This attribute refers to where the message came from (the content script) instead of the URL that is displayed in the URL bar.
  * Downside is that its possible, though not likely, that an iframe could match this and incorrectly be tagged.  If we were worried about this the `indexOf` call could be improved to guard against this.
  * Upside is that users don't need to grant additional permissions

I filed [bug 1361765](https://bugzilla.mozilla.org/show_bug.cgi?id=1361765) to investigate this on the Firefox side of things.  However it appears, so far, that Firefox is implementing this correctly.

clarkbw:
According to the docs the URL shouldn’t be included in the tab object
unless the react-devtools has the ‘tabs’ permission.  I’m not sure why
this is working as is right now.

Fixes #679

### DIFF
--- a/flow/chrome.js
+++ b/flow/chrome.js
@@ -73,9 +73,9 @@ declare var chrome: {
     },
     onMessage: {
       addListener: (fn: (req: Object, sender: {
+        url: string,
         tab: {
           id: number,
-          url: string,
         },
       }) => void) => void,
     },
--- a/shells/chrome/src/background.js
+++ b/shells/chrome/src/background.js
@@ -76,7 +76,7 @@ chrome.runtime.onMessage.addListener((req, sender) => {
     // display a custom default popup when React is *not* detected.
     // It is specified in the manifest.
     var reactBuildType = req.reactBuildType;
-    if (sender.tab.url.indexOf('facebook.github.io/react') !== -1) {
+    if (sender.url.indexOf('facebook.github.io/react') !== -1) {
       // Cheat: We use the development version on the website because
       // it is better for interactive examples. However we're going
       // to get misguided bug reports if the extension highlights it


gaearon:
Thanks!

Issue #662, opened 2017-05-01
Add an update nudge to the CLI version


gaearon:
The Electron version (`packages/react-devtools`) is often installed globally, so we should add an updating nudge that prints a message when a newer version is available.

`update-notifier` is a good package that already knows how to do this so we should just integrate it there.

tsriram:
I'll take a look at this.

sumnerevans:
@gaearon, @tsriram, Is there a reason this is still open? If yes, is there anything that I can do to help complete this?

gaearon:
Sorry, I think it was done and I forgot to close the issue. Thanks for reminding!

PR #695, opened 2017-05-07
update notifications for electron app

tsriram:
For #662. Simple integration of `update-notifier` without much customization. This is how the notification looks now (locally changed the package version to previous one):

react-devtools-updatae-notifier

Would you want to customize the message? Also, I was wondering if integrating something like `node-notifier` for native OS notification would be any better?


gaearon:
Looks great, thanks!

PR #698, opened 2017-05-08
Base16 themes

### LINKED ISSUE #636: A more maintainable way to styling react-devtools
First of all, it's great to hear that the react team is focusing on improving react devtools. As a contributor to this project, one of the major limitations I faced when trying to improve the look and feel was the fact that most of the styling in devtools is done using inline styles.
The issues are -
1) Hard to maintain consistent styling across different components.
2) Writing css for simple things like hover and focus states requires adding event handlers and that has an adverse effect on performance, esp. when the DOM tree is large.
4) Writing media queries is a pain.
3) Maintenance only gets worse with time as inline styles continue to grow.

It would be less time consuming and easier to contirbute to making devtools look better if it were to have a CSS / SCSS  stylesheets based approach to things. Things like adding a dark theme ( #108 ) would become possible too.

While I know that doing this isn't a small task, it might be worth it?

bvaughn:
Resolves #636

This PR includes:
* Extract colors from static styles into new base16 `theme`.
* Add several base16 themes, including a default that looks like the current Chrome-like devtools colors.
* Add settings icon (gear) to devtools to allow users to select a theme.
* Use-selected theme is saved using `localStorage`. This works in the Chrome extension, Firefox add-on, and Electron stand-alone app without requiring any new permissions.
* Refactors styles to avoid `assign(style1, style2, style3)` in render method. Instead new style-factory-functions were added (eg `tagStyle(theme)`). This cleans up the render method in many places and works better with themes.
* Added a `theme` property to most components `contextTypes` as a way of propagating theme. When theme changes, the top-level application is deep-re-rendered by changing a `key` on the root node (that equals the selected theme's name).

### Theme toggling
Theme toggling

### Selection and hovering
Selection and hovering

### Search results highlighting
Search results highlighting

### ReactNative style editor
(The green strip to the right came from my screen recorder. It's not actually in the UI.)
ReactNative style editor

### Tabs UI
Tabs UI

### Relay UI
Relay UI


bvaughn:
FYI this should be ready to review.

Sorry for the size. I tried to minimize the footprint as much as possible.

gaearon:
I'm getting this when I click on `<Todos>` in `shells/plain`:

screen shot 2017-05-10 at 00 34 27

bvaughn:
Really? Works fine for me. ☹️  Are you doing something more than just clicking on it?

gaearon:
No, just clicking.

Is lack of border intentional? It feels a bit like a bug, but maybe I'm just not used to it.

screen shot 2017-05-10 at 00 36 27

Search color doesn't look very good:

screen shot 2017-05-10 at 00 37 05

gaearon:
Hmm. Maybe the crash only happens when I navigate with arrow keys.

bvaughn:
> Is lack of border intentional? It feels a bit like a bug, but maybe I'm just not used to it.

I removed a few borders initially b'c most of the base16 themes I've seen online are more "flat". I could add a couple specific ones back.

> Search color doesn't look very good:

Base 16 `base02` ~> "Selection Background"

> Maybe the crash only happens when I navigate with arrow keys.

Arrow key navigation works for me too. Hm. Maybe we could screenshare (although I need to step away for a bit.)

gaearon:
Crash:



It fails here:

screen shot 2017-05-10 at 00 40 02

`styles.simple` is passed as a style so React freezes it. We should not mutate it.

bvaughn:
I didn't add that.

https://github.com/facebook/react-devtools/blob/c139c2c60c946be0612b163138c4b446c31ea75d/frontend/DataView/Simple.js#L137-L139

gaearon:
Right, but it used to have `style = assign({}, style, typeStyle);` right before it, thus making it a copy.

bvaughn:
Yeah, I realized that right after writing.

bvaughn:
I'll look over this after running. Need to step away.

gaearon:
I'm not very happy with selection. It doesn't look great on most themes except the default one, and in some cases is not legible at all (GitHub). I feel like we either need to give up there or figure out another solution.

gaearon:
Minor nit: this probably should be sans serif?

screen shot 2017-05-10 at 00 44 45

Doesn't seem to match other UI elements.

bvaughn:
Sorry about the bug you saw in `Simple`. Even doing exactly what you're doing in the GIF, I can't reproduce it. I have, however, fixed what I believe is the cause of the issue.

This component was an oversight anyway. I meant to clean up those newly-added classnames. Sorry for missing them!

> I'm not very happy with selection. It doesn't look great on most themes except the default one

I'm not _thrilled_ with it either. base16 is a little tricky in this way. I'm going to take liberty with `base06` (noted in the spec as "_Light Foreground (Not often used)_") and restore the yellow color. I'll just document it inline on the type. It seems at least as good as what we had before for the few themes I spot-checked. We can always tweak any themes that don't look right. (This is the only use of `base06` in the project anyway.)

screen shot 2017-05-09 at 8 42 31 pm

> Minor nit: this probably should be sans serif?

Yeah good call. I didn't specify a font at all here, just the inherited one. It's unfortunate that we have 5 or 6 fonts scattered all over the place. I should probably clean that up in a follow-up PR as well. For now I'll make sure this font is sans-serif at least. 😄

bvaughn:
FYI I've re-added the borders around the major panes and I've updated the gifs to show the latest style changes.

bvaughn:
Meh, I went ahead and tidied up font families too. Might as well.

gaearon:
This is looking much better 😍 
Going to look at it tomorrow.

gaearon:
Can you verify it works well in Firefox and Electron shells? Just in case.

bvaughn:
Thanks Dan!

bvaughn:
Yup! Did a quick run through Firefox+Chrome+plain shells as well as the Electron standalone app and everything seemed reasonable.

bvaughn:
PS also filed issue #705 for tracking follow-up items since this PR is already huge.

bvaughn:
FYI I did some follow-up work to this PR in #706. Kept it separate as this was already pretty huge to merge.

gaearon:
A minor bug.

Here's master:

screen shot 2017-05-11 at 11 48 24 am

Notice how that separator is darker. I intentionally made it darker (just like Chrome does) to indicate that, unlike other separators, it is resizable.

However, on this branch it is neither darker, nor resizable (in vertical mode). We should at least fix resizing in vertical mode, and preferably keep a way to distinguish it from non-resizable separators on sight.

gaearon:
Also can we make the selected breadcrumb consistent with how selection works in the tree?

screen shot 2017-05-11 at 11 51 27 am

gaearon:
Additional minor nit: selected guideline style should take precedence over hover style. Otherwise it's a bit too dancey when you hover over an already selected node.

bvaughn:
> Vertical separation should work

Gah, this got broken for vertical resize only (which I didn't even notice) b'c of a `position: relative` style 2 levels above it. Nice catch!

> Resizable separator should be distinguishable

Fair enough. It doesn't _look_ as nice (IMO) but it does stand out more to make it the darker color- so I'll change that.

> Breadcrumb should be consistent with main selection

Okay. I can dig that. 😄 

> Additional minor nit: selected guideline style should take precedence over hover style. Otherwise it's a bit too dancey when you hover over an already selected node.

I mimic what Chrome and Firefox do here. I think the current behavior is correct.

bvaughn:
I think all nits have been addressed 😄

bvaughn:
Oh snap! 1 second. I need to re-add `position: relative` tag (for the prefs popup!) Lemme push an alternate fix.

bvaughn:
Ok. Resolved for real this time. 😅

gaearon:
Will review tomorrow!

>I mimic what Chrome and Firefox do here. I think the current behavior is correct.

I don't think Chrome prioritizes hover over select.



It stays blue even when you hover it.

But in the PR, the hover style (light gray) is prioritized over selected style (dark gray) when you hover over a selected item. This makes it more noisy.

I can't speak to Firefox but I personally would prefer we keep aligned with Chrome more closely.

gaearon:
(I would also prefer a Chrome-style solid line with light blue for the selected guideline to a dashed gray line, but I realize there might not be a suitable blue.)

bvaughn:
@gaearon Sure no hurry!

I'm confused by your comment though. Are you sure you'r running the latest devtools? It does what Chrome does which is why I thought you were suggesting the opposite:
devtools-hover-dem

bvaughn:
> (I would also prefer a Chrome-style solid line with light blue for the selected guideline to a dashed gray line, but I realize there might not be a suitable blue.)

How about I tackle this in part 2 PR? Just so I don't have to merge it forward.

line-color-demo

gaearon:
Ok this looks great. LGTM

Issue #638, opened 2017-04-26
Add `==$r` suggestion in inspect tree for currently selected node


radubrehar:
In chrome devtools the currently selected node displays a `== $0` faded text besides it.
React Devtools would need to display it like `== $r` - see image below.

 ![59b3e2d8-29d9-11e7-8f74-c5c414a1e0b3](https://cloud.githubusercontent.com/assets/196656/25426015/7f5f5b64-2a76-11e7-99bb-8fd880937f1b.png)

Also relates to #618

jomahoney:
I'm gonna jump on this if y'all don't mind

skyboyer:
I believe it has been implemented and deployed some time ago.
![2018-11-18 22_43_30-uncontrolled components react](https://user-images.githubusercontent.com/956687/48677828-6efb3480-eb83-11e8-8c92-b2c779ed27bb.png)

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #699, opened 2017-05-08
Add == $r  for selected node index

dmitrika:
According to the issue #638 I added == $r text, like we have $0 in chrome devtools.
I copied colors from devtools original $0 text, but not sure with naming.

Also could'nt find out how I can cover with tests this feature.
Feel free to point out any nits or mistakes that I have done 👍


gaearon:
Can you add a screenshot?

gaearon:
While this is not related I would also appreciate if you could find a way to disable this feature in RN DevTools (`packages/react-devtools`). It's not supported there, and the fact that it appears is confusing.

There is an existing system for checking “capabilities” of different shells so maybe you could use this.

gaearon:
Actually never mind my last comment, it works on RN.

dmitrika:
Here gif for this pr:
node index gif

Yeah, didn't thought about RN, actually. Glad, that it's works the same way.

dmitrika:
I resolved conflits with current master branch. Because of the new color schemes I decided to use tag colors for `== $r`.

Example with Eighties:
![== $r in Eighties]
(https://media.giphy.com/media/l4FGjJNxQB9p9BwqI/giphy.gif)

gaearon:
@bvaughn Mind reviewing?

bvaughn:
> I resolved conflits with current master branch. Because of the new color schemes I decided to use tag colors for `== $r`.

I think this looks like a reasonable color choice 😄 

I think I'm seeing a bug in this though, both in Firefox and Chrome, where the `$r` element isn't always updated:
firefox-bug

bvaughn:
Also the `$r` label shows up for the Electron shell, even though `$r` isn't accessible anywhere (that I know of) in that case. Is it accessible? If not, should we hide the label for that (and the plain shell maybe)?

gaearon:
It's accessible in Electron but you need to follow instructions in its README (enable JS debugging).

bvaughn:
Whoops. RTFM ☺ Sorry.

bvaughn:
Well it looks like this PR is good to merge then, since the bug I mention above wasn't introduced in this revision. I'll create a follow up issue for it.

gaearon:
Thanks!

PR #704, opened 2017-05-10
Editor UX fix

### LINKED ISSUE #668: React Native Style Editor UX issues

ansumanshah:
#668 

- [x] Rename and Set Style bug in backend

### DIFF
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -306,17 +306,20 @@ Panel.childContextTypes = {
 
 var panelRNStyle = (bridge, supportsMeasure) => (node, id) => {
   var props = node.get('props');
-  if (!props || !props.style) {
-    return (
-      <div key="rnstyle" style={styles.container}>
-        <strong>No style</strong>
-      </div>
-    );
+  var readOnly = !node.get('canUpdate');
+  if (readOnly) {
+    return null;
   }
   return (
     <div key="rnstyle" style={styles.container}>
       <strong>React Native Style Editor</strong>
-      <NativeStyler id={id} bridge={bridge} supportsMeasure={supportsMeasure} />
+      {!props || !props.style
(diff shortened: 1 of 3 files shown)


gaearon:
Thanks for PR!
In the end we decided to go with a more comprehensive fix but we appreciate the effort.

Issue #705, opened 2017-05-10
Themes follow up work


bvaughn:
Related to PRs #698 and #718, issue #108.

### Followups from Dan

- [x] Fix live state updates; regressed in 90896c9 (see [here](https://github.com/facebook/react-devtools/pull/718#issuecomment-301326181)) (PR #721)
- [x] Tweak dark theme selection colors (see [here](https://github.com/facebook/react-devtools/pull/718#issuecomment-301324534)) (PR #722)
- [x] Add better padding to "_Unable to find React on page_" interstitial text (see [here](https://github.com/facebook/react-devtools/pull/718#issuecomment-301326276)). (PR #731)
- [x] Dan can't press settings button in Firefox 47.0.1 (PR #719)
- [x] Also sees Chrome team there instead of FF theme (PR #720)
- [x] The HTML headers are jarring. Can we get rid of HTML shells and always let Panel render headers, even if shells declare the text? (PR #731)

### Better default theme behavior

- [x] As of Chrome 54 the property `chrome.devtools.panels.themeName` should contain values "default" or "dark". It would be nice for the default devtools theme to have a dark variation as well that is automatically used if the value is "dark". (If a user selects a theme _other than_ default this behavior should not apply.)
- [x] There is active discussion on how to expose this feature to add-ons in Firefox as well (see [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1349896)). Once this settles down it would be nice for the Firefox add-on to implement the same behavior as Chrome. (#802)

### Remove selected component header in detail pane

- [x] There is no equivalent in Chrome or Firefox devtools and the bright-blue active highlight makes it apparent what's selected anyway. We could gain back a lot of vertical space by removing it (>40px) that would be better used by the props/values themselves. (This item came as feedback from @sebmarkbage and @vjeux.)

bvaughn:
First step, I've created 2 new themes and made a couple of minor modifications that I think will enable us to roughly approximate the themes used by Chrome and Firefox. Screenshots~
### Chrome (light)
<img width="951" alt="screen shot 2017-05-10 at 7 26 24 pm" src="https://cloud.githubusercontent.com/assets/29597/25929685/bf99d39e-35b6-11e7-96e6-4409fd390bab.png">

### Chrome (dark)
[...]

PR #706, opened 2017-05-11
Base16 themes (part 2)

bvaughn:
Builds on PR #698; resolves follow-up items #705. (This PR appears to be larger than it is because it also includes the entirety of #698.)

I think this is a nice state for initial release of themes. Covers the use-case of custom (non-browser-specific) themes as well as the use-case of observing browser devtools settings. Firefox isn't quite there yet (see [1349896](https://bugzilla.mozilla.org/show_bug.cgi?id=1349896)) but I've left an inline comment that should be easy to address as soon as it is.

Included in this PR:
- [x] New themes added for Chrome-dark and Firefox-dark modes. (Firefox theme not yet used, pending in-review [Mozilla issue 1349896](https://bugzilla.mozilla.org/show_bug.cgi?id=1349896).)
- [x] Added new, special 'default' option to drop-down. If selected, devtools will respect Chrome devtools 'default' and 'dark' modes. Other shells (eg Firefox, Electron) will use Chrome default appearance for now.
- [x] Tweaked paddings, font-sizes, and border colors slightly to better match Chrome devtools defaults.
- [x] Removed details pane header text (on the right) based on feedback from @sebmarkbage and @vjeux (and myself).
- [x] Changed the vertical line that connects opening and closing tags in the tree-view to be a solid, light-blue to more closely match Chrome _and_ to let selected state out-rank hover state in terms of coloring. (See gif below.)
- [x] Fixed no props/state style to better match Chrome.

I also did one thing that's somewhat controversial and I'm open to reversing it- but I collapsed the tag-name color and angle brackets to be the same style. This is what Firefox does. It's nice because it allows us to remove a few wrapper `<span>` tags and it makes the Firefox dark theme look better without hurting the Chrome light/dark themes much. That being said, if people prefer the Chrome style- I can revert that part only. 😄 

##### Chrome devtools mode toggling
devtools-theme-toggle-chrome

##### New Firefox (dark) and Chrome (dark) themes
screen shot 2017-05-10 at 9 22 41 pm
screen shot 2017-05-10 at 9 22 57 pm
screen shot 2017-05-10 at 9 35 19 pm

##### New "default" option
screen shot 2017-05-10 at 9 18 03 pm

##### New tag connector line
New tag connector line

##### No props/state panel
screen shot 2017-05-11 at 9 14 09 pm


gaearon:
This looks like a bug:

screen shot 2017-05-11 at 11 46 31 am

I feel like the font got too small. Let's be at least as large as default Chrome size.

screen shot 2017-05-11 at 11 45 20 am

bvaughn:
> Let's change tags at least same size as in Chrome.

Interesting. I actually _did_ change the tags to match the font-size in my Chrome. And since I've been switching things over to rems, Cmd +/- zooms the devtools in and out pretty nicely. But I'll look more at it.

Ditto for the hole. Don't know how I caused that. :)

gaearon:
Hmm. Maybe it's just a personal perception and I'm not sure how to back it up. But tags have gotten less comfortable for me to read.

clarkbw:
> Firefox isn't quite there yet (see 1349896) but I've left an inline comment that should be easy to address as soon as it is.

The current patch is approved, should go through soon.  This would mean Firefox 56 will ship with this.

bvaughn:
I used Chrome devtools to inspect themselves and realized that they are using 2 different base font sizes for monospace and sans-serif fonts. So I've updated our approach to do the same and I think it lines up nicely now. Check it out:

screen shot 2017-05-11 at 2 59 14 pm

That being said, you can also Cmd +/- to increase or decrease the font size.

bvaughn:
> This looks like a bug:
> screen shot 2017-05-11 at 11 46 31 am

Ah! Collapsing a tag with a lot of attributes and then resizing so it's too small to all fit on a single line... Nice catch! Looks like I need to re-add a `white-space: nowrap;` that was accidentally dropped. 😁

bvaughn:
Okay, I believe all points in this PR have been addressed as well. Thanks!

gaearon:
Hmm. I just merged the other PR, and this is now showing up as a merge conflict.

gaearon:
I just merged master into this branch with `ours` strategy so this branch always wins.

gaearon:
Actually no, this messed up other things.

gaearon:
WTF is GitHub is doing 😞 
I force-pushed, and it just closed PR for me with "Reopen" button disabled.

gaearon:
Merged via https://github.com/facebook/react-devtools/pull/713.

bvaughn:
Aw, poo. Git history got messed up.

Thanks for pushing it through though Dan.

Issue #694, opened 2017-05-06
Fix Fiber DEV mode detection


gaearon:
I fixed a false positive in https://github.com/facebook/react-devtools/pull/693 but we should have a better strategy than relying on a particular function. I think we should pass a completely separate function in Fiber which we'll check for crushed-ness, and keep it explicit. This will require a small change in React.

prometheansacrifice:
[The approach](https://github.com/facebook/react/blob/b61f060044780cd3d309fc7644cddf7633212465/src/renderers/dom/ReactDOM.js#L108), using a self declared `testFunc` and checking it name in the runtime, taken in earlier ReactDOM.js seems to be missing in `ReactDOMFiber.js`. Is this the small change in React you are talking about?

If the same approach is taken in Fiber, we again run the same check in devtools right?

gaearon:
I think something similar is reasonable. We keep `testFunc` declared in `ReactDOMFiber` but pass it together with `injectInternals()` call. Then in DevTools we can expect it.

I want to check two things:

1) That `name` is minified
2) That something inside `__DEV__` block in `testFunc` does not exist at all in `toString()`'d code

prometheansacrifice:
Ah! Interesting. And use regex to check if __DEV__ block is empty?

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #709, opened 2017-05-11
Fiber dev mode detection using a test function

prometheansacrifice:
Attempting fix for #694. Yet to test creating bundle using [patched react](https://github.com/facebook/react/pull/9661). Will post the results here.

### DIFF
--- a/backend/installGlobalHook.js
+++ b/backend/installGlobalHook.js
@@ -42,14 +42,26 @@ function installGlobalHook(window: Object) {
 
         // By now we know that it's envified--but what if it's not minified?
         // This can be bad too, as it means DEV code is still there.
+        const testMinification = renderer.testMinification;
 
-        // FIXME: this is fragile!
-        // We should replace this check with check on a specially passed
-        // function. This also doesn't detect lack of dead code elimination
-        // (although this is not a problem since flat bundles).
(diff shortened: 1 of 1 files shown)


backend/installGlobalHook.js
```diff
@@ -42,15 +42,27 @@ function installGlobalHook(window: Object) {
...
 
-        // FIXME: this is fragile!
-        // We should replace this check with check on a specially passed
-        // function. This also doesn't detect lack of dead code elimination
-        // (although this is not a problem since flat bundles).
-        if (findFiberCode.indexOf('getClosestInstanceFromNode') !== -1) {
+
+        // DEV mode check
+        if (testMinification() === 42) {
+          // __DEV__ block not removed. Do we warn in console?
```
prometheansacrifice: @gaearon Need a little help here. If `__DEV__` is not removed do we warn?

prometheansacrifice:
Failing flow check. Is it okay to change the type of `renderer` to `any`?

gaearon:
No, please add the new method to renderer definition in `types.js.`

prometheansacrifice:
Not that it is necessary here, but what about [`getInspectorDataForViewTag`](https://github.com/facebook/react/blob/master/src/renderers/native/ReactNativeFiber.js#L123) and in general additions we make to `renderer` in future?

Issue #702, opened 2017-05-10
Mute useless Electron logs


gaearon:
I occasionally see stuff like this printed by the [standalone devtools](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools):

<img width="882" alt="screen shot 2017-05-10 at 10 51 32 am" src="https://cloud.githubusercontent.com/assets/810438/25893167/b21ea264-356e-11e7-80de-4277b4527e73.png">

We should find a way to mute these logs if possible.

prometheansacrifice:
Should all logs be muted?

gaearon:
Yea. (At least coming from Electron native side.)

prometheansacrifice:
We can simply change `{stdio: 'inherit'}` to ` {stdio: 'ignore'}`. But that would prevent **all** logs from electron.

Else we could listen on like this process.stdout.on('data', () => {}) and turn it off appropritately

gaearon:
I'm cool with `{stdio: 'ignore'}`.

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #710, opened 2017-05-11
Ignores all logs from electron

prometheansacrifice:
Fix for #702

### DIFF
--- a/packages/react-devtools/app.js
+++ b/packages/react-devtools/app.js
@@ -11,8 +11,6 @@
 var app = require('electron').app;  // Module to control application life.
 var BrowserWindow = require('electron').BrowserWindow;  // Module to create native browser window.
 var path = require('path');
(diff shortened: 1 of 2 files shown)


gaearon:
Does this keep the output from `update-notifier`? If not we probably want to move its call into `bin`.

prometheansacrifice:
> move its call into bin
I don't understand.

Also, what's `update-notifier`?

gaearon:
See https://github.com/facebook/react-devtools/pull/695. We added this tool to notify users about updates, and we want to keep *that* output. We could move its call to the `bin.js` script you're editing though (so that it doesn't become ignored).

prometheansacrifice:
Got it. Yep. Makes sense. Added a log statement locally and tested. Seems fine. Pushing it as a separate commit for review.

gaearon:
Thanks!

PR #712, opened 2017-05-12
Enable text selection (and copy/paste) in props/state panel

### LINKED ISSUE #697: allow to copy/paste props/state from dev tools
Would be useful if it were possible to copy/paste props/state key-value. Currently right click is disabled so there is no way to copy it. 

https://twitter.com/HozefaJ/status/860202273753845760

bvaughn:
Resolves #697

### DIFF
--- a/frontend/detail_pane/DetailPane.js
+++ b/frontend/detail_pane/DetailPane.js
@@ -38,11 +38,6 @@ var styles = {
     flex: 1,
     display: 'flex',
     flexDirection: 'column',
-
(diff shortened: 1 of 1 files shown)


bvaughn:
cc @gaearon in case you know why were disabling text-selection in this component

I _think_ it was probably just a copy-pasted style setting that we didn't think about.

bvaughn:
This seems relatively uncontroversial so I'm going to just merge it.

PR #713, opened 2017-05-12
More theming changes from Brian

gaearon:
This is a resubmit of #706 which I somehow messed up.

### DIFF
--- a/frontend/Breadcrumb.js
+++ b/frontend/Breadcrumb.js
@@ -17,6 +17,7 @@ import type {Base16Theme} from './types';
 var Fonts = require('./Themes/Fonts');
 var React = require('react');
 var decorate = require('./decorate');
+var {hexToRgba} = require('./Themes/utils');
 
 class Breadcrumb extends React.Component {
   context: {theme: Base16Theme};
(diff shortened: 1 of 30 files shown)


gaearon:
This was a bit tricky because we messed up Git history a little, but I think I cherry-picked all commits in the other PR that weren't in the first PR. This should work.

spicyj:
image

Not showing the component name in the sidebar makes it much harder to copy "WayfinderKioskStitchedMap" (for example) out of the devtools.

bvaughn:
Hi Ben 😄 

The name was moved from the sidebar b'c room there is precious and I decided to look to Chrome's Elements tab for prior art. I'm not totally opposed to re-adding it. I wonder if what you're describing may be fixable with a context menu option though?

spicyj:
With DOM elements it's less likely that the name is a long thing I want to copy. Also, in Chrome I can double-click a name to get an editable field:

image

bvaughn:
¯\_(ツ)_/¯ I didn't say I added complete parity with Chrome DevTools

If you have a concrete proposal, file an issue. I'll be happy to look at it when I get some cycles for DevTools. 😄  I only mentioned the context menu b'c it doesn't take up precious, limited space and we've been trending features in that direction (eg copy name, copy props, and #865).

bvaughn:
PS Sorry if that sounded snarky; realized after submitting it that it might have. Unintentional!

PR #715, opened 2017-05-12
Use higher contrast for selected node

gaearon:
Currently on dark themes, the selection contrast is not very great:

screen shot 2017-05-12 at 12 58 43 pm

I found that changing this improves it on most themes:

screen shot 2017-05-12 at 12 59 09 pm

The only one suffering from this change is Materia:

screen shot 2017-05-12 at 1 29 34 pm

Can we just fix that specific case (e.g. by using a lighter selection instead of blue)?

### DIFF
--- a/frontend/Breadcrumb.js
+++ b/frontend/Breadcrumb.js
@@ -86,7 +86,7 @@ const containerStyle = (theme: Base16Theme) => ({
 const itemStyle = (isSelected: boolean, isComposite: boolean, theme: Base16Theme) => {
   let color;
   if (isSelected) {
-    color = theme.base04;
+    color = theme.base01;
   } else if (isComposite) {
     color = theme.base0E;
   }
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -486,7 +486,7 @@ const headStyle = ({
 const jsxTagStyle = (inverted: boolean, isCustom: boolean, theme: Base16Theme) => {
   let color;
   if (inverted) {
-    color = 'inherit';
+    color = theme.base01;
   } else if (isCustom) {
     color = theme.base08;
   } else {


gaearon:
Another thing that is slightly annoying is that host components often have too low contrast in dark themes:




I wanted to fix them by using `base04` instead of `base03`:

screen shot 2017-05-12 at 1 36 32 pm

However now they disappear in Chrome:

screen shot 2017-05-12 at 1 36 39 pm

They do because `base04` is lighter than `base03` in Chrome, despite it being a light theme. So we don't follow these suggestions:

>Colors base00 to base07 are typically variations of a shade and run from darkest to lighest. [...] In order to create a dark theme colors base00 to base07 should span from dark to light. For a light theme these colours should span from light to dark.

Is this intentional? It seems like the ``00..07` being `dark => light` for dark themes, and `light => dark` for light themes is important for Base16 to make sense and be cohesive. Do we have some use case that breaks it, and can we fix those individual violations?

gaearon:
It seems to me that we should ensure that:

1) All dark themes have `00..07` as a strict dark => light progression
2) All light themes have `00..07` as a strict light => dark progression

I think we might need to introduce our custom `10` and `11` for selection and search highlighting. Or maybe we could somehow calculate/choose them based on background or accent color (I'm okay with deviating from specific Chrome selection color—my only goal is that selection stays inverted).

bvaughn:
I think a better approach to this would be to tweak/fix specific colors in themes that aren't quite right. Changing which baes16 colors we use could have a much larger impact than we want and further break away from normal/recommended base16 usage. For example, this change uses 01 for the color- but according to base16, 01 is to be used for the default background. I'm concerned that this sort of change would limit our ability to re-use preexisting base16 themes if/when we open things up more for user-submitted themes.

I already tweaked colors like this for several of the themes. Admittedly I didn't really spend any time doing this for some of the dark themes. I could do a quick pass over them.

Issue #684, opened 2017-05-04
[RN] Custom Debugger for Remote JS debugging


ansumanshah:
Can we add an option for custom [debugger](https://facebook.github.io/react-native/docs/debugging.html#debugging-using-a-custom-javascript-debugger) in the standalone electron package?

Just like [`react-native-debugger`](https://github.com/jhen0409/react-native-debugger) as we have to open chrome for JS debugging, devtools for Inpecting React Components and one more tool if we use state management like redux.

gaearon:
Can you explain what would this entail on our side and how it would work?

ansumanshah:
Maybe @jhen0409 will be able to explain in detail on what would this entail.
We could just add a Plugin for this and a flag to turn it on.

I believe this would help us get the debugger for react-native inside the electron app, that will maybe help the debugger to always be in the foreground when debugging js bundle.

RN default Chrome Debugger
<img width="1280" alt="screenshotb" src="https://cloud.githubusercontent.com/assets/3208863/25711498/aea16e02-310c-11e7-87f3-b6d8565f70b3.png">
Debugger inside the electron app
<img width="1023" alt="screenshot" src="https://cloud.githubusercontent.com/assets/3208863/25711514/b581a502-310c-11e7-9fed-9f4cf8c7cb33.png">

gaearon:
Oh my, that would be amazing. How hard is it to set up?

jhen0409:
It‘s not hard, except take code from [`debugger.html`](https://github.com/facebook/react-native/blob/master/local-cli/server/util/debugger.html), we need to know the RN packager port (default is 8081), I think it can be an option for start react-devtools, then we can use [`REACT_DEBUGGER`](https://github.com/facebook/react-native/blob/118e981174ae513426e2f5ddd7ee3c5860519ff7/local-cli/server/middleware/getDevToolsMiddleware.js#L29) env for open react-devtools by RN packager:

```bash
$ REACT_DEBUGGER="react-devtools --enable-rn-debug --rn-packager-port=8082" react-native start
```

We can even set different [`window.__REACT_DEVTOOLS_PORT__ `](https://github.com/facebook/react-native/blob/master/Libraries/Core/Devtools/setupDevtools.js#L33) in debugger worker for multi-packager support. (I'm also doing on [`react-native-debugger#63`](https://github.com/jhen0409/react-native-debugger/pull/63))

bvaughn:
[...]

PR #716, opened 2017-05-12
[Electron] Add setDefaultThemeName method

jhen0409:
This PR make `setDefaultThemeName` method in DevToolsUI for set `themeName` prop for `Panel`.

```js
require('react-devtools-core/standalone')
  .setContentDOMNode(document.getElementById('container'))
  .setDefaultThemeName('ChromeDark')
  .startServer(port);
```

I've a plan for detect Chrome DevTools theme on [`react-native-debugger`](https://github.com/jhen0409/react-native-debugger/commit/f5adc772c2a71a95795e8c56693d7e98cacf226e), I think it also useful for #684.

Also parse `--theme` for `react-devtools` bin, so we can use:

```bash
$ react-devtools --theme ChromeDark
```

### DIFF
--- a/packages/react-devtools-core/src/standalone.js
+++ b/packages/react-devtools-core/src/standalone.js
@@ -192,8 +192,9 @@ var DevtoolsUI = {
       var {store} = panel.getChildContext();
       // Change default themeName if panel mounted
       store.setDefaultThemeName(themeName);
-      // Change theme with user selected theme,
-      // it always wins than default theme
+      // Render devtools with the current theme.
+      // If user has selected a theme then ThemeStore will return it.
+      // If not the new default we set above will be used.
       store.changeTheme(ThemeStore.get());
     }
     return DevtoolsUI;
--- a/packages/react-devtools-core/src/standalone.js
+++ b/packages/react-devtools-core/src/standalone.js
@@ -16,6 +16,7 @@ var path = require('path');
 var installGlobalHook = require('../../../backend/installGlobalHook');
 installGlobalHook(window);
 var Panel = require('../../../frontend/Panel');
+var ThemeStore = require('../../../frontend/Themes/Store');
 var launchEditor = require('./launchEditor');
 var React = require('react');
 var ReactDOM = require('react-dom');
@@ -24,6 +25,7 @@ var node = null;
 var onStatusChange = function noop() {};
 var projectRoots = [];
 var wall = null;
+var panel = null;
 
 var config = {
   reload,
@@ -44,16 +46,18 @@ function reload() {
   ReactDOM.unmountComponentAtNode(node);
   node.innerHTML = '';
   setTimeout(() => {
-    ReactDOM.render(<Panel showHiddenThemes={true} {...config} />, node);
+    panel = ReactDOM.render(<Panel showHiddenThemes={true} {...config} />, node);
   }, 100);
 }
 
 function onDisconnected() {
+  panel = null;
   ReactDOM.unmountComponentAtNode(node);
   node.innerHTML = '<div id="waiting"><h2>Waiting for React to connect…</h2></div>';
 }
 
 function onError(e) {
+  panel = null;
   ReactDOM.unmountComponentAtNode(node);
   var message;
   if (e.code === 'EADDRINUSE') {
@@ -177,6 +181,19 @@ var DevtoolsUI = {
 
   setStatusListener(_listener) {
     onStatusChange = _listener;
+    return DevtoolsUI;
+  },
+
+  setDefaultThemeName(themeName) {
+    config.themeName = themeName;
+    if (panel) {
+      var {store} = panel.getChildContext();
+      // Change default themeName if panel mounted
+      store.setDefaultThemeName(themeName);
+      // Change theme with user selected theme,
+      // it always wins than default theme
+      store.changeTheme(ThemeStore.get());
+    }
     return DevtoolsUI;
   },
 
--- a/packages/react-devtools/app.js
+++ b/packages/react-devtools/app.js
@@ -15,7 +15,9 @@ var updateNotifier = require('update-notifier');
 var pkg = require('./package.json');
 
 var mainWindow = null;
-var argv = process.argv.slice(2);
+var argv = require('minimist')(process.argv.slice(2));
+var projectRoots = argv._;
+var defaultThemeName = argv.theme;
 
 app.on('window-all-closed', function() {
   app.quit();
@@ -34,7 +36,10 @@ app.on('ready', function() {
     // We use this so that RN can keep relative JSX __source filenames
     // but "click to open in editor" still works. js1 passes project roots
     // as the argument to DevTools.
-    'window.devtools.setProjectRoots(' + JSON.stringify(argv) + ')'
+    'window.devtools.setProjectRoots(' + JSON.stringify(projectRoots) + ')'
+  );
+  mainWindow.webContents.executeJavaScript(
+    'window.devtools.setDefaultThemeName(' + JSON.stringify(defaultThemeName) + ')'
   );
 
   // Emitted when the window is closed.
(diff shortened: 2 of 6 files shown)


packages/react-devtools/app.js
```diff
@@ -15,7 +15,10 @@ var updateNotifier = require('update-notifier');
 var pkg = require('./package.json');
 
 var mainWindow = null;
-var argv = process.argv.slice(2);
```
jhen0409: I also fixed this, it called `slice(2)` twice. ([`bin.js`](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools/bin.js#L4))

gaearon: Hmm, I'm pretty sure it worked. Isn't calling `slice(2)` twice fine because `bin.js` spawns a new process (which has its own `argv`)?

jhen0409: Sorry I misunderstood it, you can ignore this comment. 😂

bvaughn:
So the plan is for the Chrome extension to notify the Electron instance of its devtools, if running?

The "hidden" Chrome and Firefox look-alike themes are directly exposed in the prefs panel, which allows users to select them explicitly, or one of the other built-in themes. Unless I'm misreading this PR, wouldn't it have the effect of clobbering a user-selected theme in favor of eg "ChromeDefault" or "ChromeDark"?

jhen0409:
> So the plan is for the Chrome extension to tell the Electron devtools what the current Chrome theme is?

Yes.

> Maybe we could pass-through the Chrome theme only if one of the hidden Chrome these is currently selected.

I think it would be good have an option like `Match browser` with `showHiddenThemes` in UI for Electron, then pass by `themeName` prop for Panel, otherwise the `localStorage.themeName` will be always not null if user selected theme.

jhen0409:
I updated this PR, allows standalone app set `themeName` prop as default theme name:

It will not effect user selected theme (Firefox dark):




Add `Reset` option, let it back to default theme name:

packages/react-devtools-core/src/standalone.js
```diff
@@ -44,16 +46,18 @@ function reload() {
   ReactDOM.unmountComponentAtNode(node);
   node.innerHTML = '';
   setTimeout(() => {
-    ReactDOM.render(<Panel showHiddenThemes={true} {...config} />, node);
+    panel = ReactDOM.render(<Panel showHiddenThemes={true} {...config} />, node);
```
bvaughn: FYI `showHiddenThemes` was moved into the `config` in another PR. (Sorry! I Should have put it there to begin with.)

Should be an easy fix to just drop in:
```jsx
ReactDOM.render(<Panel {...config} />, node);
```

packages/react-devtools-core/src/standalone.js
```diff
@@ -180,6 +184,19 @@ var DevtoolsUI = {
...
   },
 
+  setDefaultThemeName(themeName) {
+    config.themeName = themeName;
+    if (panel) {
+      var {store} = panel.getChildContext();
+      // Change default themeName if panel mounted
+      store.setDefaultThemeName(themeName);
+      // Change theme with user selected theme,
+      // it always wins than default theme
```
bvaughn: This comment could perhaps be clearer.

```js
// Render devtools with the current theme.
// If user has selected a theme then ThemeStore will return it.
// If not the new default we set above will be used.
```

bvaughn:
> I updated this PR, allows standalone app set themeName prop as default theme name

By the way, I like the use of default theme for this. Fits in nicely.

jhen0409:
> I think the "Reset" option is confusing. For example, if I launch with --theme Dracula and click "Reset" from the dropdown- I would expect the "Dracula" theme to be unselected. However it remains selected. This makes sense, given the explanation, but I think it hints at the fact that this should actually be a button rather than a drop-down option. Thoughts?

Agreed, I think this is also a better experience (Instead of click twice). I added `Reset` button on bottom of `PreferencesPanel`. It also in browser side, a easy way for back to `Match browser`.

bvaughn:
I wonder if the "Reset" button should auto-close the panel too? (Not sure, just questioning.)

I think there's a bug when resetting that's new to this branch:
devtools-bug

Compare to master with the default browser-specified theme:
devtools-master

bvaughn:
The problem seems to be that the `Panel` is initially rendering with the CLI-specified default theme as the top level `key`. (So the `Store` theme and the `Panel` theme mis-match. The `Store` restores the persisted theme and only falls back to the injected default but `Panel` just assumes the prop is correct.)

bvaughn:
I think a fix for this might be to make this change to `Panel`:
```diff
diff --git a/frontend/Panel.js b/frontend/Panel.js
index bc42779..b8d1f16 100644
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -224,7 +224,10 @@ class Panel extends React.Component {
       window.addEventListener('keydown', this._keyListener);
 
       this._store.on('connected', () => {
-        this.setState({loading: false});
+        this.setState({
+          loading: false,
+          themeName: this._store.themeName,
+        });
         this.getNewSelection();
       });
       this._store.on('preferencesPanelShown', () => {

```

I also wonder if we could DRY up `Store` theme checking a bit more?
```diff
diff --git a/frontend/Store.js b/frontend/Store.js
index b6302d8..670b625 100644
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -149,10 +149,7 @@ class Store extends EventEmitter {
 
     // Don't restore an invalid themeName.
     // This guards against themes being removed or renamed.
-    const restoredThemeKey = ThemeStore.get();
-    const themeName = restoredThemeKey && Themes.hasOwnProperty(restoredThemeKey)
-      ? restoredThemeKey
-      : this._defaultThemeName;
+    const themeName = this._safeThemeName(ThemeStore.get(), this._defaultThemeName);
 
     this.theme = Themes[themeName];
     this.themeName = themeName;
@@ -345,11 +342,15 @@ class Store extends EventEmitter {
     this.emit('contextMenu');
   }
 
+  _safeThemeName(maybeThemeName: string, safeThemeName: string): string {
+    return maybeThemeName && Themes.hasOwnProperty(maybeThemeName)
+      ? maybeThemeName
+      : safeThemeName;
+  }
+
   changeTheme(themeName: ?string) {
     // Only apply a valid theme.
-    const safeThemeKey = themeName && this.themes.hasOwnProperty(themeName)
-      ? themeName
-      : this._defaultThemeName;
+    const safeThemeKey = this._safeThemeName(themeName, this._defaultThemeName);
 
     this.theme = this.themes[safeThemeKey];
     this.themeName = safeThemeKey;
@@ -361,9 +362,7 @@ class Store extends EventEmitter {
 
   setDefaultThemeName(defaultThemeName: ?string) {
     // Don't accept an invalid themeName as a default.
-    this._defaultThemeName = defaultThemeName && Themes.hasOwnProperty(defaultThemeName)
-      ? defaultThemeName
-      : 'ChromeDefault';
+    this._defaultThemeName = this._safeThemeName(defaultThemeName, 'ChromeDefault');
   }
 
   showPreferencesPanel() {
```

jhen0409:
> I wonder if the "Reset" button should auto-close the panel too? (Not sure, just questioning.)

I think it's a button that back default values, I guess there will be more property in the future, auto-close may not be good to me.

> I think there's a bug when resetting that's new to this branch

I can't reproduce this 🤔, but container `key` is really a problem, I'll apply your fix.

bvaughn:
> I think it's a button that back default values, I guess there will be more property in the future, auto-close may not be good to me.

That's fine. In that case, I suggest we disable the "reset" button when the style is already the default. This way we avoid confusion of the user clicking the button and nothing seems to happen.

bvaughn:
I think this looks great otherwise! I'll go ahead and make my own suggested change regarding the "Reset" button so you don't have to worry about doing another round- then we're good to go!

PR #717, opened 2017-05-12
Fix source path not shown fully if content overflow

jhen0409:
See the screenshot:

It has been scrolled to end, remove `overflow: auto` will fix this.

### DIFF
--- a/frontend/PropState.js
+++ b/frontend/PropState.js
@@ -220,7 +220,6 @@ const emptyStyle = (theme: Base16Theme) => ({
 const sourceStyle = (hasViewElementSource: boolean, theme: Base16Theme) => ({
   padding: '0.25rem 0.5rem',
   color: theme.base05,
-  overflow: 'auto',
   overflowWrap: 'break-word',
   cursor: hasViewElementSource ? 'pointer' : 'default',
 });


gaearon:
@bvaughn Can you review this please?

bvaughn:
Hey @jhen0409. Thanks for contributing a PR! 😄 

I'm afraid I don't understand this fix at first look]. `overflow: auto` just tells the browser to enable scrolling if needed. In this case, I don't see how the default value ("visible") would result in any different behavior if the content extends beyond the bounds of the container it's in.

But you've verified it fixes the issue you were seeing?

jhen0409:
Here is screenshot:



Not sure why it was set `overflow: auto`, the `source` area will enable scrolling when `PropState` content overflow:

bvaughn:
I'm just surprised to see that the `overflow` property affects the _height_ of the component. I thought it was totally about how spillover was handled (eg hidden, visible, or scrollable). In this case, I would have expected the real fix to be setting a `flex` grow property or something.

jhen0409:
Hmmm I think we can set `min-height` for source style, but it will only show one line if `PropState` content overflow.

bvaughn:
Right. I don't think min-height is the right fix b'c of line wrap. In fact, the presence of the `min-height` style causes a bit of funkiness with the padding so I'd suggest we just remove it entirely.

```diff
diff --git a/frontend/PropState.js b/frontend/PropState.js
index 7846c82..3703983 100644
--- a/frontend/PropState.js
+++ b/frontend/PropState.js
@@ -222,10 +222,8 @@ const emptyStyle = (theme: Theme) => ({
 const sourceStyle = (hasViewElementSource: boolean, theme: Theme) => ({
   padding: '0.25rem 0.5rem',
   color: theme.base05,
-  overflow: 'auto',
   overflowWrap: 'break-word',
   cursor: hasViewElementSource ? 'pointer' : 'default',
-  minHeight: Fonts.monospace.sizes.normal,
 });
 
 const sourcePosStyle = (theme: Theme) => ({
```

jhen0409:
OK, I removed it. :)

bvaughn:
Thanks @jhen0409 !

PR #718, opened 2017-05-13
Theming improvements

bvaughn:
Follow-up to PR #715

* Renamed `Base16Theme` type to just `Theme` since we've added additional colors.
* Cleaned up a few `Fonts` import statements to import specific font families instead.
* Re-added unique coloring for &lt; and &gt; characters. (I had removed this previously but I regretted it.)
* Added a couple of helper/util functions to the color module. (Some of these aren't used in the app at this point but were useful to me when iterating and debugging so I thought I'd leave them in since they're super tiny.)
* Added a new top-level parameter, `showHiddenThemes`, that can be used for plain and Electron shells to expose the "hidden" themes (eg Chrome and Firefox lookalikes). I think this improves the experience in those shells while allowing us to keep our smart defaults for Chrome and Firefox.
* Tightened up theme filtering and selection logic a bit in `PreferencesPanel`.
* Added a "preview" option (the little eye icon) to preferences panel to preview a theme's appearance. Admittedly this is mostly useful when iterating on the theme colors but maybe it's worth leaving in for users? I could be convinced to remove it pretty easily if there's any dislike.
* Tweaked prop/val coloring slightly.
* Added a couple of additional `Theme` colors for selected foreground and background states. This enables better contrast _and_ better matching of Chrome and Firefox devtools coloring. I also used Chrome devtools to inspect Chrome devtools and more closely match browser default colors.
* Improved inline documentation for themes for future contributors.
* Added `FirefoxDefault` theme for light-mode.

##### Chrome (default) theme
Chrome (default) theme

##### Chrome (dark) theme
Chrome (dark) theme

##### Firefox (default) theme
Firefox (default) theme

##### Firefox (dark) theme
Firefox (dark) theme

##### Theme preview panel
Theme preview panel


bvaughn:
Okay @gaearon, I think this one's ready to be looked at.

gaearon:
Any objections to making vertical lines closer to Chrome?

Currently the hover vertical line is darker than Chrome, but the selected vertical line is lighter.
I'd like them to be closer to Chrome (make hover gray lighter, but selected blue brighter).

gaearon:
Another nit is that the vertical guideline of parent should not be obscured by child hover.

Chrome:

screen shot 2017-05-14 at 17 35 25

DevTools 2.1.9:

screen shot 2017-05-14 at 17 37 05

This PR:

screen shot 2017-05-14 at 17 35 13

gaearon:
Regarding the colors, I think most themes work fairly well. I would like to consider changing selection color in Dracula:

screen shot 2017-05-14 at 17 40 44

and Materia:

screen shot 2017-05-14 at 17 43 04

To me, it feels like dark themes should have a light selection color. Maybe something like beige could work better for both? This would both solve contrast issues and probably look better.

gaearon:
Host element in the Chrome default theme are a tad *too* light.
I think making them darker just a tiny bit will make them look much better, at least on rMBP.

screen shot 2017-05-14 at 17 46 42

gaearon:
Does this font have true italics? It looks a bit odd but maybe it’s just be not being used to it.

screen shot 2017-05-14 at 17 48 13

gaearon:
Should the closing `/` be part of the angle color?

screen shot 2017-05-14 at 17 49 34

gaearon:
Can we make the outline a bit more subtle in the default them (some alpha, like for the guideline?)

screen shot 2017-05-14 at 17 50 49

For comparison, default browser outline:

screen shot 2017-05-14 at 17 51 38

gaearon:
Maybe we should rename "default" to "Match Chrome theme" (or "Match Firefox theme" in Firefox)?
Otherwise it's not obvious that there's a dark more matching Chrome theme.

Or "Match Chrome theme (currently, Light)".

gaearon:
Minor nit but I would like the default search color to be a bit less acidic/greenish, and closer to browser's yellow.

screen shot 2017-05-14 at 17 58 10

gaearon:
Clicking on settings icon in Firefox doesn't work for me.

gaearon:
Here is something else that broke as part of theming work.
Before, I could see live updates to the state:



This no longer works (neither after first theming PR, nor in this branch):

gaearon:
I see this state flashing while the plugin loads:

screen shot 2017-05-14 at 18 13 19

gaearon:
I'm not sure changing the "Connecting" header to be dark made it better, as now the Electron shell jumps between two different colors before potentially jumping to a third set of colors (dictated by the theme).

Can we use the same color for all central messages?

bvaughn:
> Any objections to making vertical lines closer to Chrome?
> Currently the hover vertical line is darker than Chrome, but the selected vertical line is lighter.
I'd like them to be closer to Chrome (make hover gray lighter, but selected blue brighter).

Sure. I'll tweak it some.

> Another nit is that the vertical guideline of parent should not be obscured by child hover.

I did this intentionally because it looks back when the light gray hover bar overlaps the selected bar. I'll just use alpha instead for the gray bar and make them overlap again.

screen shot 2017-05-14 at 10 27 23 am
screen shot 2017-05-14 at 10 27 30 am

> I would like to consider changing selection color in Dracula...and Materia
> To me, it feels like dark themes should have a light selection color

Tempted to say we can follow up with specific theme-tweaks like this in a smaller PR. Getting the colors exactly right is kind of exhausting and I, admittedly, took a bit of a shortcut with the non-default-browser themes. Sound reasonable?

> Host element in the Chrome default theme are a tad too light.
> I think making them darker just a tiny bit will make them look much better, at least on rMBP.

Ok. I can darken it a little to  #888888 

> Does this font have true italics? It looks a bit odd but maybe it’s just me not being used to it.

Depends on the font family. Fallback is "sans-serif" which does. The others... I dunno. Looks ok to me? I used italics where Chrome does.

> Should the closing / be part of the angle color?

Yeah! Good catch.

> Can we make the outline a bit more subtle in the default theme (maybe add some alpha, like we do for the guideline?)

Sure.

> Maybe we should rename "default" to "Match Chrome theme" (or "Match Firefox theme" in Firefox)?
Otherwise it's not obvious that there's a dark more matching Chrome theme.
Or "Match Chrome theme (currently, Light)".

I like this suggestion. Maybe just "match browser theme" (although we might be slightly misadvertising since we won't match _exactly_).

> Minor nit but I would like the default search color to be a bit less acidic/greenish, and closer to browser's yellow.

Sure. I'll go #FFFF00 #ffff00.

> Clicking on settings icon in Firefox doesn't work for me.

Uh...works for me? Been testing and refreshing with Chrome and Firefox (plus plain and Electron). Unload and reload?

> Here is something else that broke as part of theming work.
Before, I could see live updates to the state:
This no longer works (neither after first theming PR, nor in this branch):

Rats.
If it's not a new issue introduced in this branch, I'd prefer to follow up with new PR.

> I see this state flashing while the plugin loads:
I think it didn't use to be so pressed against the borders.

Hm... I'm not sure how my changes would have impacted the placement of this string. This is something rendered by the wrappers (`panel.html` and `run.js`)- neither of which were modified in my PR. Maybe you just didn't notice before?

> I'm not sure changing the "Connecting" header to be dark made it better...Can we use the same color for all central messages?

This is bit of a pain b'c of how the rendering of these messages is strung across HTML files and React source. I guess I'll hard-code the 2 headers inside of the devtools app to match the ones declared in the HTML that wraps the app and has no way to use the theme.

bvaughn:
I've appended the 2 follow-up items to issue #705. The rest of the change have been made

gaearon:
>I did this intentionally because it looks back when the light gray hover bar overlaps the selected bar. 

That's not the behavior in master. Please check: in master, they overlap one way but not the other. I remember because I spent a few hours figuring out how to do it with CSS. :-)

gaearon:
To clarify, this is how it works in master.

Hover parent's line is below child's blue background:

screen shot 2017-05-14 at 22 03 33

But parent's line is above child's light blue hover background:

screen shot 2017-05-14 at 22 03 38

gaearon:
>Hm... I'm not sure how my changes would have impacted the placement of this string. This is something rendered by the wrappers (panel.html and run.js)- neither of which were modified in my PR. Maybe you just didn't notice before?

Sorry I wasn't clear—I interpreted this PR as a followup on all issues from theming. So I'm including other things that regressed since the initial merge of themes. I could create a separate issues.

gaearon:
>Hm... I'm not sure how my changes would have impacted the placement of this string. This is something rendered by the wrappers (panel.html and run.js)- neither of which were modified in my PR. Maybe you just didn't notice before?

You're right, this happened before.

gaearon:
This is the part that made zIndex work in the past:

screen shot 2017-05-14 at 22 12 40

frontend/Node.js
```diff
@@ -601,11 +618,7 @@ const guidelineStyle = (depth: number, isSelected: boolean, isHovered: boolean,
     bottom: 0,
     willChange: 'opacity',
     left: calcPaddingLeft(depth) - 7,
-    // Bring it in front of the hovered children, but make sure
-    // hovering over parents doesn't draw on top of selected
-    // guideline even when we've selected the closing tag.
-    // When unsure, refer to how Chrome does it (it's subtle!)
-    zIndex: isSelected ? 1 : 0,
+    zIndex: 1,
```
gaearon: It does seem like putting this back fixes z ordering issue.

gaearon:
I just pushed a fix for this (I hope it wasn't intentional). Also tweaked the alpha to match the Chrome gray more closely.

gaearon:
>Uh...works for me? Been testing and refreshing with Chrome and Firefox (plus plain and Electron). Unload and reload?

Did that a few times, also restarted the process. Consistently doesn't work for me for Firefox 47.0.1. (I also see the Chrome theme in it for some reason.)

However I can also reproduce this on master so I'll add to followup instead.

gaearon:
Thanks! I'll add followups to https://github.com/facebook/react-devtools/issues/705.

bvaughn:
> Did that a few times, also restarted the process. Consistently doesn't work for me for Firefox 47.0.1.

Firefox 47? Why such an old version? (That was released almost 1 year ago.) I'm running Firefox 53.0.2.

>  (I also see the Chrome theme in it for some reason.)

Yeah, the default themes are hidden behind the "Match browser" setting except for plain and electron shells. I thought that made more sense. Seemed unlikely to me that someone would go for the Chrome look-and-feel within Firefox.

bvaughn:
> Dan can't press settings button in Firefox 47.0.1. Also sees Chrome team there instead of FF theme.

Sounds like maybe you have the chrome theme name stored in `localStorage` from when it was initially in the list? Have you tried selecting a new theme, then selecting "Match browser" again to make it go away?

I've tested FF several times, including a complete re-install of it (for unrelated reasons). I think it should work...

gaearon:
Whatever I had on my laptop. :-)
If we don't support it we should bump the minimal required version.

bvaughn:
Ah! It seems to be related to the default values for top/left properties with `position:absolute`

Should be fixed in #719

infinnie:
Hello, I have been developing themes (light and dark) that rely on colors not changing upon selection:
image
image

Maybe there could be a Boolean field in the theme definition to enable or disable color inversion.

bvaughn:
The code to style elements and attributes is already super complex. I don't think we should add any more complexity to it. If anything, I would love to see us _remove_ complexity from it. :)

PR #728, opened 2017-05-17
[WIP] launchEditor global hook

Timer:


frontend/Panel.js
```diff
@@ -331,9 +339,7 @@ class Panel extends React.Component {
         extraTabs={extraTabs}
         preferencesPanelShown={this.state.preferencesPanelShown}
         theme={theme}
-        onViewElementSource={
-          this.props.showElementSource ? this.viewElementSource.bind(this) : null
-        }
+        onViewElementSource={this.hasLaunchEditor ? this.viewElementSource.bind(this) : null}
```
Timer: We might not be able to do this anymore, as the global hook could be installed at any time and not cause a re-render (obviously). A lot of children components rely on this, however (to determine if it should add `onClick` and `cursor` changes).

If CRA installs this immediately (before a crash), it might not be a problem.

frontend/Panel.js
```diff
@@ -320,7 +328,7 @@ class Panel extends React.Component {
               key: 'showComponentSource',
               title: 'Show ' + node.get('name') + ' source',
               action: () => this.viewComponentSource(id),
-            }, this.props.showElementSource && node.get('source') && {
+            }, this.hasLaunchEditor && node.get('source') && {
```
Timer: Ditto (see below).

frontend/Panel.js
```diff
@@ -239,6 +241,12 @@ class Panel extends React.Component {
     });
   }
 
+  componentWillUpdate() {
+    this.hasLaunchEditor =
```
Timer: This expression might also need to be in `componentDidMount`, not sure about the lifecycle here.

packages/react-devtools-core/src/standalone.js
```diff
@@ -44,6 +41,12 @@ function reload() {
   ReactDOM.unmountComponentAtNode(node);
   node.innerHTML = '';
   setTimeout(() => {
+    if (typeof window.__REACT_DEVTOOLS_GLOBAL_HOOK__ === 'undefined') {
+      window.__REACT_DEVTOOLS_GLOBAL_HOOK__ = {};
```
gaearon: We should never need to do this.
`installGlobalHook` is guaranteed to set it (which is called at the beginning).

frontend/Panel.js
```diff
@@ -184,8 +182,12 @@ class Panel extends React.Component {
     }
     this._bridge.send('putSelectedInstance', id);
     setTimeout(() => {
-      invariant(this.props.showElementSource, 'cannot view source if props.showElementSource is not supplied');
-      this.props.showElementSource(source);
+      invariant(
+        typeof window.__REACT_DEVTOOLS_GLOBAL_HOOK__ !== 'undefined' &&
```
gaearon: It's always guaranteed to be defined.

frontend/Panel.js
gaearon: Let's assume it is set before Panel is rendered for the first time. I *think* that would be the case in CRA environment. So let's just assume that reading it once at startup is safe enough.

frontend/Panel.js
```diff
@@ -239,6 +241,12 @@ class Panel extends React.Component {
     });
   }
 
+  componentWillUpdate() {
+    this.hasLaunchEditor =
+      typeof window.__REACT_DEVTOOLS_GLOBAL_HOOK__ !== 'undefined' &&
```
gaearon: This one always exists.

packages/react-devtools-core/src/standalone.js
```diff
@@ -44,6 +41,12 @@ function reload() {
...
+    }
+    window.__REACT_DEVTOOLS_GLOBAL_HOOK__.launchEditor = (fileName, lineNumber) => {
```
gaearon: Can we assign it outside once? I'd prefer to do it before declaring the config.

Timer: I wasn't sure if it'd ever get overwritten and we wanted to make sure it was reinstalled; if that's not a concern I'll simply hoist it.

Timer:
Updated PR; I have yet to test it with CRA integration.

gaearon:
Okay, this doesn’t quite work.
The `window` in Chrome extension is not the app `window` so we can’t read a global like this.

Something like this seems to work:

```diff
diff --git a/frontend/Panel.js b/frontend/Panel.js
index 443b62c..f0c5771 100644
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -112,7 +112,7 @@ class Panel extends React.Component {
     }
 
     this.hasLaunchEditor =
-      typeof window.__REACT_DEVTOOLS_GLOBAL_HOOK__.launchEditor === 'function';
+      typeof this.props.launchEditor === 'function';
   }
 
   componentWillUnmount() {
@@ -186,10 +186,10 @@ class Panel extends React.Component {
     this._bridge.send('putSelectedInstance', id);
     setTimeout(() => {
       invariant(
:...skipping...
diff --git a/frontend/Panel.js b/frontend/Panel.js
index 443b62c..f0c5771 100644
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -112,7 +112,7 @@ class Panel extends React.Component {
     }
 
     this.hasLaunchEditor =
-      typeof window.__REACT_DEVTOOLS_GLOBAL_HOOK__.launchEditor === 'function';
+      typeof this.props.launchEditor === 'function';
   }
 
   componentWillUnmount() {
@@ -186,10 +186,10 @@ class Panel extends React.Component {
     this._bridge.send('putSelectedInstance', id);
     setTimeout(() => {
       invariant(
-        typeof window.__REACT_DEVTOOLS_GLOBAL_HOOK__.launchEditor === 'function',
+        typeof this.props.launchEditor === 'function',
         'cannot view source if __REACT_DEVTOOLS_GLOBAL_HOOK__.launchEditor is not supplied'
       );
-      window.__REACT_DEVTOOLS_GLOBAL_HOOK__.launchEditor(source.fileName, source.lineNumber);
+      this.props.launchEditor(source.fileName, source.lineNumber);
     }, 100);
   }
 
diff --git a/shells/chrome/src/panel.js b/shells/chrome/src/panel.js
index b1add10..0ade93a 100644
--- a/shells/chrome/src/panel.js
+++ b/shells/chrome/src/panel.js
@@ -31,6 +31,11 @@ var config: Props = {
     chrome.devtools.inspectedWindow.eval('window.__REACT_DEVTOOLS_GLOBAL_HOOK__.$0 = $0');
     bridge.send('checkSelection');
   },
+  launchEditor(fileName, lineNumber) {
+    chrome.devtools.inspectedWindow.eval('window.__REACT_DEVTOOLS_GLOBAL_HOOK__.launchEditor('
+      + JSON.stringify(fileName) +',' +JSON.stringify(lineNumber) 
+      +')');
+  },
   selectElement(id, bridge) {
     bridge.send('putSelectedNode', id);
     setTimeout(() => {
```

But it doesn't let me determine if it's available synchronously.

Maybe we should just always make it clickable but add some kind of a help tooltip explaining in which cases it works? Even in CRA apps it's not guaranteed to click correctly unless you configure the editor integration (e.g. on WIndows).

gaearon:
On CRA side, I'm testing integration using this:

```js


if (window.__REACT_DEVTOOLS_GLOBAL_HOOK__) {
  window.__REACT_DEVTOOLS_GLOBAL_HOOK__.launchEditor = (sourceFileName, sourceLineNumber) => {
    fetch(
    '/__open-stack-frame-in-editor?fileName=' +
      window.encodeURIComponent(sourceFileName) +
      '&lineNumber=' +
      window.encodeURIComponent(sourceLineNumber || 1)
    )
  };
}
```

Timer:
> The `window` in Chrome extension is not the app window so we can’t read a global like this.

Mm, I should've saw that coming. I believe the tooltip is the best course of action for this.
I'll go ahead and revise this PR.

Timer:
# Proposal

## Devtools

`react-devtools` will allow clicking on source files in chrome and will optimistically expect it to work.

Most users will re-click a second time if it doesn't work after a second, which should be more than enough time on modern hardware for us to determine if the feature is working.

  - If the `eval` throws an exception, this means the global hook does not exist. We **could** rule this case out earlier.
    * In this case, we can automatically open a URL to instructions on how to configure the feature

  - If the `eval` returns a non-truthy value, we will consider that the hook is not properly configured. However, we will not do anything special in this case (other than disabling the cursor change).
    * the application will be responsible for showing the error and instructions to mitigate the problem
    * optionally, we can look for an object with the property `{ error: true, fatal: boolean, message: '' } and show something in the devtools

## Hook spec

The hook will return false if the editor is not properly configured or cannot be deduced ([we do this sync](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-dev-utils/launchEditor.js#L81-L112)). If macOS, maybe return an object like so:
```js
{ error: true, fatal: false, message: 'Please open a supported editor or set the EDITOR variable.' }
```

For more complicated failure cases, e.g. [process spawning](https://github.com/facebookincubator/create-react-app/blob/master/packages/react-dev-utils/launchEditor.js#L184-L194), we can cache this state for a couple retries.

gaearon:
Sounds reasonable.

bvaughn:

PR #731, opened 2017-05-20
Cleaned up devtools loading text [WIP]

bvaughn:
Relates to #705

* Various loading/connecting texts use new, shared `Message` component so the text renders more consistently.
* Moved some of the themes-related helper methods into `Themes/utils` for cleanup
* This PR also merged #655 since there was some overlap and that one was half reviewed.


samit4me:
This looks incredibly similar to the work I have done in #655. I literally just did a rebase yesterday after all the major theming changes lol. Might be worth merging the two, what you think?

bvaughn:
Hi @samit4me!

I agree there's some overlap! Sorry for not seeing your PR before creating my own.

Let me try to merge your branch into mine.

bvaughn:
(FYI) Need a bit to try to resolve the differences between the 2 branches.

bvaughn:
@samit4me Looks good for the most part after merging, though I did notice a flicker when switching back to a React page in Firefox:
glitch

The no-mounted-nodes message temporarily flickers in the lower left panel.

samit4me:
That's awesome @bvaughn, thanks for merging the two 🎉 . When I saw `<LoadingText>` and how similar it was to `<Message>` it was obvious we were both trying to achieve the same thing, which can only be a good sign.

Regarding the flicker, has that got something to do with `_loadingTimeout` in `Panel.js`? I added a small delay to give react a chance to load, so the user does see "loading...", then "react not detected", then the devtools loading.

samit4me:
Ah actually, is it in `shells/chrome/src/panel.js`? At the very end take a look at:
```JS
ReactDOM.render(<Panel alreadyFoundReact={true} themeName={themeName} {...config} />, node);
```

I did change that to the following but it looks like it got reverted after the big rebase I did yesterday:
```JS
ReactDOM.render(<Panel {...config} alreadyFoundReact={true} themeName={themeName} />, node);
```

The props were always overridden with the config. Hopefully, that's the issue?

bvaughn:
Yeah, I noticed the timer. I'm on the fence about it. Need to think on it a bit more. 😄

samit4me:
I was too, but the loading timer should only apply while react is not found. There is a another timer (been there for sometime) that looks for react every 200 ms.

Is there a particular aspect that is concerning or is it only the flicker when navigating back to a page with react? Do you know if it is actually take that long for react to load or do you believe there is an issue in the logic? I will take another look tonight to see if I can replicate.

bvaughn:
> I was too, but the loading timer should only apply while react is not found. There is a another timer (been there for sometime) that looks for react every 200 ms. Is there a particular aspect that is concerning or is it only the flicker when navigating back to a page with react?

Just timers in general are kind of "meh" and adding another one feels unfortunate. I haven't really had time to put any thought into alternate approaches though.

> It might look better if the theme can also be applied to the loading screen. What do you think? Is that possible?

Unfortunately this doesn't look right because of the stand-alone shell's initial screen (which is hard-coded, no access to themes). At one point I had the `Message` component using the theme background and foreground colors but this made the Electron shell look weird so I backed off.

samit4me:
💯 agree, timers are very meh. I believe the real problem here is to detect when the page has finished loading and only then actually look for react. Not sure how possible that is and this was my first commit, so I decided to use the timeout.

I did some testing on Firefox and Chrome tonight with and without the timeout and the results were interesting. On Firefox it had zero impact, it was the same with or without the timeout. On Chrome the timeout did prevent the "React not detected" message from showing, but this doesn't look to be a big problem, so it certainly won't hurt to remove the timeout.

Chrome no `_loadingTimeout`:
chrome-react-no-timeout

Chrome with `_loadingTimeout`:
chrome-react-with-timeout

^ Sorry for the 👁  trauma 😄, but I hope that shows what the timeout does.

---

RE: theming, I noticed that the "Looking for react" and "React not detected" message have no themes applied, which makes sense after your comments above, but I noticed that the "No components are mounted" message has the theme applied, does that look correct?
screen shot 2017-05-22 at 8 53 10 pm

shells/chrome/src/panel.js
```diff
@@ -114,10 +121,4 @@ function reload() {
...
 }
 
-// chrome.devtools.panels added in Chrome 18.
-// chrome.devtools.panels.themeName added in Chrome 54.
-const themeName = (chrome.devtools.panels : any).themeName === 'dark'
-  ? 'ChromeDark'
-  : 'ChromeDefault';
-
-ReactDOM.render(<Panel alreadyFoundReact={true} themeName={themeName} {...config} />, node);
+ReactDOM.render(<Panel alreadyFoundReact={true} {...config} />, node);
```
samit4me: Should the `alreadyFoundReact` prop be added after the `config` prop? Otherwise, `alreadyFoundReact` will always be false as defined on line 30.

shells/firefox/panel/run.js
```diff
@@ -27,7 +28,7 @@ const themeName = 'FirefoxLight';
 
 function reload() {
   ReactDOM.unmountComponentAtNode(node);
-  ReactDOM.render(<Panel alreadyFoundReact={true} themeName={themeName} {...config} />, node);
+  ReactDOM.render(<Panel alreadyFoundReact={true} {...config} />, node);
```
samit4me: Another place where `alreadyFoundReact` is overridden by `config`.

bvaughn: `alreadyFoundReact` isn't declared in `config`

shells/chrome/src/panel.js
bvaughn: Huh. Yeah, nice catch! Although this wasn't changed in this PR. I guess it just...didn't work correctly before, in master.

shells/firefox/panel/run.js
samit4me: Oops, sorry wrong one, my bad

bvaughn:
> On Chrome the timeout did prevent the "React not detected" message from showing, but this doesn't look to be a big problem, so it certainly won't hurt to remove the timeout.

Ah, nice. So you think it's okay to just remove that timeout then?

> RE: theming, I noticed that the "Looking for react" and "React not detected" message have no themes applied, which makes sense after your comments above, but I noticed that the "No components are mounted" message has the theme applied, does that look correct?

Hmm...I _think_ this is okay, since that message is shown within the larger chrome of the devtools.

screen shot 2017-05-22 at 12 59 32 pm

samit4me:
So many themes, did more get added? I'm digging the Gruvbox (light) theme.

> Ah, nice. So you think it's okay to just remove that timeout then?

Yeah certainly, the timeout is adding complexity for very little benefit, so I think removing it would be best.

> Hmm...I think this is okay, since that message is shown within the larger chrome of the devtools.

Makes sense to me, I just pulled the updated branch and it appears to be working well here 👍

bvaughn:
> So many themes, did more get added? I'm digging the Gruvbox (light) theme.

Yeah! 😁 4 more have been added by external contributors today. Exciting!

bvaughn:
PS Would you like to remove the setTimeout call and make a PR into this branch? I'm a but busy to do it at the moment.

frontend/Themes/utils.js
```diff
@@ -23,6 +25,20 @@ function getBrightness(hex: string): number {
...
 }
 
+function getDefaultThemeName(): string {
+  return 'ChromeDefault';
+}
+
+function getSafeThemeName(themeName: ?string, defaultThemeName: ?string): string {
+  if (themeName && Themes.hasOwnProperty(themeName)) {
+    return themeName;
+  } else if (defaultThemeName === 'string') {
```
jhen0409: `typeof defaultThemeName === 'string'`

bvaughn: Hm. Looks like I flubbed up that merge/refactor. Thanks.

samit4me:
> PS Would you like to remove the setTimeout call and make a PR into this branch? I'm a but busy to do it at the moment.

Done, see https://github.com/facebook/react-devtools/pull/737

bvaughn:
Okay, I think this one's ready for a review.

bvaughn:

PR #735, opened 2017-05-22
Add Tomorrow and Tomorrow Night themes

daywaves:
I saw your tweet and decided to add my favorite theme (and its light version!)

Tomorrow

Tomorrow Night

I wasn't sure which colors to use for base0H to base0L, so I chose what looked right to me. Let me know if I should change anything!

### DIFF
--- a/frontend/Themes/Themes.js
+++ b/frontend/Themes/Themes.js
@@ -283,6 +283,56 @@ const Phd: Theme = {
   base0L: '#061229',
 };
 
+const Tomorrow: Theme = {
+  displayName: 'Tomorrow',
+  base00: '#ffffff',
+  base01: '#e0e0e0',
+  base02: '#d6d6d6',
+  base03: '#8e908c',
+  base04: '#969896',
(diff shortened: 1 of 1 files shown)


bvaughn:
This looks great! Thanks for being the first to submit a custom theme. 😁

Regarding the color range you mention, I have a couple of minor nits:
* **Tomorrow**: The white foreground text (`base0L`) with yellow background for search results is a bit low contrast. Maybe we should use a dark foreground color for filter matches (like with the dark theme)?
* **Tomorrow Night**: The selected-and-focused background (`base0H`) looks a bit unfocused to me. (This color is often gray for the other unfocused themes.) I wonder if there's a more saturated color we could use for the focused background and then we could use the current light-gray (`#e0e0e0`) for the unfocused background?

I'd love to know your thoughts on the above. Otherwise I think this looks good to go!

bvaughn:
I'm going to move forward with the suggested tweaks. Let me know if you have any concerns or would otherwise like to tweak them further! 😁

PR #738, opened 2017-05-23
Fix drag position in SplitPane

jhen0409:
I'm experiencing the drag problem for use [`react-devtools-core`](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools-core) and render it on different position:

Use `node.getBoundingClientRect` fixed this.

### DIFF
--- a/frontend/SplitPane.js
+++ b/frontend/SplitPane.js
@@ -59,15 +59,15 @@ class SplitPane extends React.Component {
   }
 
   onMove(x: number, y: number) {
-    var node = ReactDOM.findDOMNode(this);
+    var rect = ReactDOM.findDOMNode(this).getBoundingClientRect();
 
     this.setState(prevState => ({
       width: this.props.isVertical ?
         prevState.width :
-        Math.floor(node.offsetLeft + node.offsetWidth - x),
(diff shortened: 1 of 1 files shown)


bvaughn:
I'm not able to reproduce the problem you're showing in either the Chrome and Firefox extensions nor the Electron and plain shells. Could you give more context on where you're seeing this?

jhen0409:
This will more clearer:



It's not affect browser extensions & Electron shell because it render on (x: 0, y: 0), but affected `react-devtools-core` usage.

bvaughn:
I'm sorry, but I still don't understand. Which target build is it affecting, if not Chrome/Firefox/plain/Electron?

jhen0409:
This is the case for who uses [`react-devtools-core`](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools-core) to building an another Electron app and put `#waiting` tag on the different position. It doesn't affect all targets of this repo.

bvaughn:
Ah, thank you for elaborating. I didn't realize anyone was using the core in this way. 😄

In that case, I think your change looks reasonable. Thank you for explaining more about it.

PR #729, opened 2017-05-18
Devtools inspect-element button

bvaughn:
Resolves #701

Works for Chrome and Firefox (and plain shell). Disabled for standalone shell.

PR #740, opened 2017-05-24
Show highlighter when inspect is enabled

joecritch:
(Kind of a follow-up to #729)

- Added highlighting for the Inspect button when hovering
- Disables the Inspect button once an element has been clicked
- Hide the highlight when the window is resized (otherwise it is incorrectly positioned)

inspect-button-hover

cc @bvaughn

### DIFF
--- a/agent/Agent.js
+++ b/agent/Agent.js
@@ -126,6 +126,8 @@ class Agent extends EventEmitter {
       this._updateScroll = this._updateScroll.bind(this);
       window.addEventListener('scroll', this._onScroll.bind(this), true);
       window.addEventListener('click', this._onClick.bind(this), true);
+      window.addEventListener('mouseover', this._onMouseOver.bind(this), true);
+      window.addEventListener('resize', this._onResize.bind(this), true);
     }
   }
 
@@ -213,6 +215,7 @@ class Agent extends EventEmitter {
       bridge.forget(id);
     });
     this.on('setSelection', data => bridge.send('select', data));
+    this.on('setInspectEnabled', data => bridge.send('setInspectEnabled', data));
   }
 
   scrollToNode(id: ElementID): void {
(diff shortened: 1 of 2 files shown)


agent/Agent.js
```diff
@@ -213,6 +218,7 @@ class Agent extends EventEmitter {
       bridge.forget(id);
     });
     this.on('setSelection', data => bridge.send('select', data));
+    this.on('setInspectEnabled', data => bridge.send('setInspectEnabledFromBackend', data));
```
joecritch: I'd be keen to get thoughts on the naming of the `setInspectEnabledFromBackend` architecture and naming.

bvaughn:
> Disables the Inspect button once an element has been clicked

I thought about doing this but decided against it because, as I was testing the feature, I realized that I frequently either mis-clicked or wanted to click more than one item when browsing around. Also since there is no shortcut key that will re-activate the tool once the page has been clicked, I thought it maybe better to leave the tool active.

That being said, the hover change drastically decreases the likelihood of me clicking on the wrong thing. Overall this change feels really nice. 😁

agent/Agent.js
bvaughn: Why not just `setInspectEnabled` ?

agent/Agent.js
```diff
@@ -160,6 +162,9 @@ class Agent extends EventEmitter {
     bridge.on('selected', id => this.emit('selected', id));
     bridge.on('setInspectEnabled', enabled => {
       this._inspectEnabled = enabled;
+      if (enabled) {
+        this.emit('stopInspecting');
```
bvaughn: It's not obvious to me why this is necessary.

joecritch: You're right, it wasn't necessary. I think it was zombie code from when I was debugging. Oops!

agent/Agent.js
joecritch: Oh, I thought this would cause a circular reference. Works fine though :)

joecritch:
@bvaughn Thanks man, I've made those changes. WRT disabling the Set Inspect button, do you think I should revert that then?

bvaughn:
> WRT disabling the Set Inspect button, do you think I should revert that then?

No, I think this change feels really nice over-all.

bvaughn:
I have a follow-up change I'd like to make to this that maybe I'll tag you on for review? 😄

joecritch:
@bvaughn 🙌 I'll leave this with you then. Yeah cool, feel free to tag me. Also — I was thinking it would be nice if the inspect button activated its composite component, rather than the DOM component?

bvaughn:
> Also — I was thinking it would be nice if the inspect button activated its composite component, rather than the DOM component?

To me it seems equally plausible that you'd want to select the host component vs the composite component. 😄

PR #744, opened 2017-05-26
Organized theme keys, removed unused keys

bvaughn:
After creating a few themes, I've noticed the previous `base00...base0L` naming to be a bit confusing. In part this is because the order of keys is so arbitrary after `base07`. This is because of how we evolved kind of organically from the original base16 color set.

I've addressed that by grouping keys into 3 categories (base, special, and state) and cleaning up the docs around color usage:
screen shot 2017-05-26 at 10 05 43 am

I also changed the following:
* I noticed that 2 of the previous `base00...base07` keys weren't being used so I deleted them.
* I created a new field, `special07` for host components, and copied the value from `base04` into it. It seemed like a unique enough use-case that it didn't belong in the base colors.

I've migrated all themes and their references to use the newly named keys via find-and-replace.

This is an initial step towards a simpler theme-creation UI, which I've begun playing with in another branch, eg:
screen shot 2017-05-26 at 9 59 36 am


bvaughn:
cc @alexkrolick, @LanceJernigan, @daywaves

PR #748, opened 2017-05-29
[React Native] Fix instance measure of rnstyle for Android

jhen0409:
It looks like measure some instance will got `undefined` in Android (See [this comment](https://github.com/jhen0409/react-native-debugger/issues/84#issuecomment-304611817)), we can avoid it use `undefined` values as `measuredLayout` send to react-devtools.

The original errors will causing react-devtools not to work properly on Android.

### DIFF
--- a/plugins/ReactNativeStyle/setupBackend.js
+++ b/plugins/ReactNativeStyle/setupBackend.js
@@ -65,7 +65,8 @@ function measureStyle(agent, bridge, resolveRNStyle, id) {
   }
 
   instance.measure((x, y, width, height, left, top) => {
-    // It may got undefined for measure some instance in Android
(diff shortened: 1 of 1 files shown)


gaearon:
Doesn't this mean React Native itself has a bug in `measure()` implementation?

jhen0409:
> Doesn't this mean React Native itself has a bug in measure() implementation?

It should does, it seems this issue is only for some top-level `View`. I just saw the built-in inspector of RN, it have the same issue (It should get the correct width / height in iOS):



The built-in inspector doesn't throw any errors, so I think we can avoid the error first in here. I might to tracking code of RN for Android, try to fix it or just file an issue.

gaearon:
Thanks. I’ll merge this for now but please raise an issue in RN.

PR #750, opened 2017-05-29
Clear stored themeName if select defaultThemeName

jhen0409:
Currently if we select the `defaultThemeName` on PreferencesPanel, the `Reset` button will be disabled but `localStorage.themeName` still not be `''`, so we should clear it if user select `defaultThemeName`.

cc @bvaughn

### DIFF
--- a/frontend/PreferencesPanel.js
+++ b/frontend/PreferencesPanel.js
@@ -127,13 +127,18 @@ class PreferencesPanel extends React.Component {
   }
 
   _changeTheme = (event) => {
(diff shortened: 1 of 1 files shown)


bvaughn:
@jhen0409 What are your thoughts on #751? I wasn't super thrilled with the "Reset" button functionality so I got rid of it in that PR.

jhen0409:
That will be better, let me close this. :)

bvaughn:
❤️ Thanks!

PR #751, opened 2017-05-29
Theme editor GUI

bvaughn:
After making a walk-through video of how to create a custom devtools theme, I felt it was currently too complicated. So this is my attempt to clean things up a bit in the UI and the code.

This is a lot of code. For what it's worth, I tried to keep each commit atomic and meaningful to simplify the review.

One note: I'm not sure why I had to tag some instance methods with `$FlowFixMe` for `ColorInput` and `ColorPicker`.

### Code changes
- [x] Refactored theme-specific `Store` code into `Themes/Store` object for easier maintainability.
- [x] Tidied up `localStorage` logic a bit into helper methods used by `Themes/Store`.
- [x] Added support for "Custom" theme, cloned from `ChromeDefault` when launched the first time.
- [x] Add error handling with default values for invalid or outdated themes. New `Serializer` white-lists theme keys and also sets default values for any missing keys.
- [x] Added new `SvgIcon` component to tidy up redundant SVG tags and styles a bit.

### User-facing change
- [x] Default theme now displays as the browser name (eg "Chrome") instead of "Match browser". I think this is more intuitive _and_ works better with the new "Custom" option.
- [x] Selecting "Custom" from themes menu shows pencil button which opens the new theme editor (see below video). Theme editor has color picker and realtime preview to make theme customization something that can be done easily without running devtools locally.
- [x] Added option to import/export theme via a text input in the theme editor. This idea was inspired by Slack's sidebar themes and seemed like a reasonable first pass at a sharing UI, given how simple it was to implement.

### Customizing a theme
theme-editor-final-customizing

### Importing / exporting between browsers
theme-editor-final-sharing

### Reset unsaved changes
theme-editor-final-reset


jhen0409:
It seems changed the `Reset` button to `Chrome` option for browser, what is the solution for Electron shell? Also, currently [this line](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools-core/src/standalone.js#L194) is broken on this branch.

bvaughn:
> It seems changed the Reset button to Chrome option for browser, what is the solution for Electron shell?

The plain and electron shells don't use this option b'c they are configured with `showHiddenThemes={true}`. (There is no "default" for them.) They would just look like this:

screen shot 2017-05-29 at 9 15 51 am

If we wanted them to use the same functionality we _could_ pass a `browserName` param in. I'm not sure if it makes as much sense to match the browser in that case though, particularly not if it introduces any significant complexity.

> Also, currently this line is broken on this branch.

Thanks for pointing that out. I've added a "test Electron shell" todo to the WIP section. 😄

bvaughn:
Okay, I think this PR is ready to rock.

jhen0409:
> If we wanted them to use the same functionality we could pass a browserName param in. I'm not sure if it makes as much sense to match the browser in that case though, particularly not if it introduces any significant complexity.

Could we add `browserName` arg into `setDefaultThemeName` or just add `setBrowserName` for [`react-devtools-core/src/standalone`](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools-core/src/standalone.js)?

In case of [`react-native-debugger`](https://github.com/jhen0409/react-native-debugger) I think it is well suited to have an option for match devtools. Not sure if there is a suitable name for [`react-devtools`](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools) package.

bvaughn:
Hm. Sure I guess we can mirror the functionality. This PR is already kind of huge so I might prefer for one of us to do that in a follow up.

bvaughn:
CI errors seem unrelated to my recent commits.

PR #755, opened 2017-05-30
Fiber devtools fixes

bvaughn:
Builds on top of Dan's work in PR #528, with Dom's changes from master merged in. Initial smoke tests of Stack + Fiber seem promising.

### DIFF
--- a/backend/getDataFiber.js
+++ b/backend/getDataFiber.js
@@ -98,13 +98,8 @@ function getDataFiber(fiber: Object, getOpaqueNode: (fiber: Object) => Object): 
       if (typeof fiber.stateNode.setNativeProps === 'function') {
         // For editing styles in RN
         updater = {
-          setState() {},
-          forceUpdate() {},
-          setInState() {},
-          setInContext() {},
-          setInProps(path: Array<string | number>, value: any) {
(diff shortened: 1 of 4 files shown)


backend/getDataFiber.js
```diff
@@ -134,7 +146,14 @@ function getDataFiber(fiber: Object, getOpaqueNode: (fiber: Object) => Object):
 }
 
 function setInProps(fiber, path: Array<string | number>, value: any) {
-  fiber.pendingProps = copyWithSet(fiber.memoizedProps, path, value);
+  const inst = fiber.stateNode;
+  fiber.pendingProps = copyWithSet(inst.props, path, value);
```
trueadm: @sebmarkbage @acdlite is this logic still applicable given the recent changes to how fiber will work with async?

trueadm:
I've tested this locally and it works for me with both Stack and Fiber. I've made a comment for @acdlite in regards to his recent PR changes (https://github.com/facebook/react/pull/9695) that may/may not affect this.

backend/getDataFiber.js
acdlite: In https://github.com/facebook/react/pull/9695, we never read `pendingProps` from the current tree, so I think this is fine.

Issue #759, opened 2017-05-31
Fix RN style editor


gaearon:
Let's just fix all of these. It can't be too hard.

#### Done (needing review)
- [x] #761: RN style editor doesn't work with themes
- [x] #762: Editing RTCView properties produces redbox on Fiber
- [x] #762: Editing RTCView properties doesn't update DevTools UI (both Stack and Fiber)
- [x] #762: Remove style by deleting property name
- [x] #765: It's too hard to click into the value field of a newly created style before any text is entered.
- [x] https://github.com/facebook/react-devtools/pull/770 and https://github.com/facebook/react-native/pull/14290 Sometimes, the Inspect tool fails to select certain UI bits (eg Catalyst app main "Search" input). Mentioned [here](https://github.com/facebook/react-devtools/pull/762#issuecomment-305310100).

gaearon:
https://github.com/facebook/react-devtools/pull/762

bvaughn:
#761

PR #762, opened 2017-05-31
Fix RN style editor again (both Stack and Fiber)

gaearon:
This should fix two first items off https://github.com/facebook/react-devtools/issues/759.
Please test it.

Editing both RCTViews and Views should work now. Style changes should reflect in UI. The last one edited wins.

### DIFF
--- a/backend/getDataFiber.js
+++ b/backend/getDataFiber.js
@@ -80,11 +80,12 @@ function getDataFiber(fiber: Object, getOpaqueNode: (fiber: Object) => Object): 
       break;
     case HostComponent:
       nodeType = 'Native';
-      name = typeof fiber.type === 'string' ?
-        fiber.type :
-        // Necessary for React Native Fiber if host types are not strings.
-        // https://github.com/facebook/react/pull/9013
-        getDisplayName(fiber.type);
+      name = fiber.type;
+
+      // TODO: we plan to remove this prefix anyway.
+      // We can cut this special case out when it's gone.
+      name = name.replace('topsecret-', '');
+
       publicInstance = fiber.stateNode;
       props = fiber.memoizedProps;
       if (
--- a/backend/getDataFiber.js
+++ b/backend/getDataFiber.js
@@ -82,7 +82,7 @@ function getDataFiber(fiber: Object, getOpaqueNode: (fiber: Object) => Object): 
       nodeType = 'Native';
       name = fiber.type;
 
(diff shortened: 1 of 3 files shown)


bvaughn:
Editing works for RCT views now! 😁 

Added props are "lost" (in the UI) when editing RCT views though. The same is true for stack so maybe this was always the case?

For fiber, inspecting things in the Simulator doesn't update the active/selected node in devtools. (Although selecting a new node in devtools does update the selected thing in the simulator.) In Stack it works in both directions.

backend/getDataFiber.js
```diff
@@ -80,11 +80,12 @@ function getDataFiber(fiber: Object, getOpaqueNode: (fiber: Object) => Object):
...
     case HostComponent:
       nodeType = 'Native';
-      name = typeof fiber.type === 'string' ?
-        fiber.type :
-        // Necessary for React Native Fiber if host types are not strings.
-        // https://github.com/facebook/react/pull/9013
-        getDisplayName(fiber.type);
+      name = fiber.type;
+
+      // TODO: we plan to remove this prefix anyway.
```
bvaughn: Feel free to tag me by name in this TODO if you'd like. It's on my TODO list to remove the prefix in RNF anyway.

bvaughn:
Actually I think new styles aren't "lost" for fiber on RCT views after all. Maybe I only saw that earlier b'c I had a stale backend.

Regarding the Inspector clicks not updating the devtools selected node, this seems to work at first but then gets broken after I click on the "search" input (in Catalyst). At this time the (frontend) console also shows an error:
> Store.js: Uncaught TypeError: Cannot read property 'get' of undefined

Which maps to:
```js
var nodeType = node.get('nodeType');
```

At this point, devtools seems to get stuck in a state of only showing "**React Native Style Editor** _loading_" for me (for fiber) regardless of which node I click on.

bvaughn:
I think I fixed the style deletion issue via c0766c1.

bvaughn:
Okay. For the most part, this PR looks great.

There's the sometimes-issue I mentioned above with `skipWrapper`:
> Store.js: Uncaught TypeError: Cannot read property 'get' of undefined

I'm not sure what causes this. I go a long time without seeing it and then I see it for a bit. Confusing.

I added a null-check in 2f976f4 for this, so even if the inspector can't select the item- at least devtools doesn't error.

Issue #754, opened 2017-05-30
Component name gone from the right panel, can't copy the name


kortv:
![image](https://cloud.githubusercontent.com/assets/10098905/26604925/a116d488-4594-11e7-9fb6-a0cbd06ee7e2.png)

glennreyes:
Not sure if that has been removed intentionally (cc @bvaughn @gaearon @jaredly). This is how it looked like before and what @kortv means iirc:
![9ffeac6c-463c-11e7-8f08-ea594ad9228d](https://cloud.githubusercontent.com/assets/5080854/26646693/d863a5b2-463c-11e7-8c84-760db21807e6.png)

gaearon:
This was intentional, to clean up some space. Why do you copy the name?

bvaughn:
Yup. Was intentional b'c vertical space is limited and we thought the props were more important than the name (given that the name is also shown in the left panel).

kortv:
@gaearon I copy the name to fast finding it in the IDE.
@bvaughn I see the name in left panel but it's not copyable
Maybe it's possible to add copy name through right click menu at the bottom?
![image](https://cloud.githubusercontent.com/assets/10098905/26653132/dacc0380-465a-11e7-87c6-481ab07995df.png)

Or add option to show the name at old place?

bvaughn:
> Or add option to show the name at old place?

Subjective opinion here, but I don't think this is worth the added complexity it would introduce to the codebase.

> Maybe it's possible to add copy name through right click menu at the bottom?

Adding an option to the context menu to copy some info from the node, maybe including name, seems potentially worth exploring.

kortv:
Maybe to make the name selectable in the selected component tab, at bottom, will be not so complex?

kortv:
Cursor at selected tab is changing from "pointer" to "default", even now. So maybe changing it to the text instead of default, will not make UX so bad?
Or even, changing the cursor to default, as it now, but, only at selected tab, on double click, make a select the name text, or enable selection.

bvaughn:
Debatably the current behavior should be changed to always preserve cursor as pointer when over a tab, even if current.

I feel pretty strongly that we should not make tab text selectable.

glennreyes:
What if we add a `Copy element name` item to the `ContextMenu`? Similar to the Chrome devtools. Unlike the `Copy element`, this would only copy the name of the selected element. This way you have an easy alternative to copy complex component names.

kortv:
[...]

PR #768, opened 2017-06-01
Add 'Copy element name' to context menu

LanceJernigan:
This feature was mentioned in issue #754 and should add a 'Copy element name' to the context menu.

### DIFF
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -18,7 +18,7 @@ var performanceNow = require('fbjs/lib/performanceNow');
 
 // Use the polyfill if the function is not native implementation
 function getWindowFunction(name, polyfill): Function {
-  if (String(window[name]).indexOf(`[native code]`) === -1) {
+  if (String(window[name]).indexOf('[native code]') === -1) {
     return polyfill;
   }
   return window[name];
--- a/frontend/Container.js
+++ b/frontend/Container.js
@@ -133,17 +133,19 @@ var DEFAULT_MENU_ITEMS = {
         title: 'Show all ' + node.get('name'),
         action: () => store.changeSearch(node.get('name')),
       });
-      items.push({
-        key: 'copyNodeName',
-        title: 'Copy element name',
-        action: () => store.copyNodeName(node.get('name')),
-      });
     }
     if (store.capabilities.scroll) {
       items.push({
         key: 'scrollToNode',
         title: 'Scroll to node',
         action: () => store.scrollToNode(id),
+      });
+    }
+    if (node.get('nodeType') === 'Composite' && node.get('name')) {
+      items.push({
+        key: 'copyNodeName',
+        title: 'Copy element name',
+        action: () => store.copyNodeName(node.get('name')),
       });
     }
     return items;
--- a/frontend/Container.js
+++ b/frontend/Container.js
@@ -133,6 +133,11 @@ var DEFAULT_MENU_ITEMS = {
         title: 'Show all ' + node.get('name'),
         action: () => store.changeSearch(node.get('name')),
       });
+      items.push({
+        key: 'copyNodeName',
+        title: 'Copy element name',
+        action: () => store.copyNodeName(node.get('name')),
+      });
     }
     if (store.capabilities.scroll) {
       items.push({
--- a/frontend/ContextMenu.js
+++ b/frontend/ContextMenu.js
@@ -11,12 +11,12 @@
 'use strict';
 
 var React = require('react');
-var { sansSerif } = require('./Themes/Fonts');
+var {sansSerif} = require('./Themes/Fonts');
 var HighlightHover = require('./HighlightHover');
 
 var decorate = require('./decorate');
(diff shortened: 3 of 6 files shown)


agent/Bridge.js
```diff
@@ -18,7 +18,7 @@ var performanceNow = require('fbjs/lib/performanceNow');
 
 // Use the polyfill if the function is not native implementation
 function getWindowFunction(name, polyfill): Function {
-  if (String(window[name]).indexOf(`[native code]`) === -1) {
+  if (String(window[name]).indexOf('[native code]') === -1) {
```
bvaughn: This was fixed in master yesterday. Maybe need to rebase?

frontend/Container.js
```diff
@@ -133,6 +133,11 @@ var DEFAULT_MENU_ITEMS = {
         title: 'Show all ' + node.get('name'),
         action: () => store.changeSearch(node.get('name')),
       });
+      items.push({
+        key: 'copyNodeName',
+        title: 'Copy element name',
+        action: () => store.copyNodeName(node.get('name')),
+      });
```
bvaughn: nit: I think this option should be at the bottom (after "Scroll to node")

I was also wondering if we should only show it for composite component types (eg `node.get('nodeType') === 'Composite'`)? It seems a bit unlikely that someone would want to copy "div" or "span" although I guess it doesn't hurt to have the menu option there either, if it makes things simpler.

frontend/ContextMenu.js
```diff
@@ -11,12 +11,12 @@
...
 
 var React = require('react');
-var {sansSerif} = require('./Themes/Fonts');
+var { sansSerif } = require('./Themes/Fonts');
 var HighlightHover = require('./HighlightHover');
 
 var decorate = require('./decorate');
 
-import type {Theme} from './types';
+import type {Theme } from './types';
```
bvaughn: This spacing is weird ^

I think we generally don't have spaces between the `{}`s and their inner words FWIW>

frontend/ContextMenu.js
```diff
@@ -58,16 +63,39 @@ class ContextMenu extends React.Component {
     this.props.hideContextMenu();
   }
 
+  getHeight = element => {
```
bvaughn: The changes to this file seem unrelated to the description of the PR. And I think they're actually included in another PR. Mind if we keep this PR focused on just the copy-name issue?

bvaughn: I guess it's okay. We can just combine that PR and this PR if it's easier for you.

frontend/ContextMenu.js
```diff
@@ -58,16 +63,39 @@ class ContextMenu extends React.Component {
...
+      pos.y -= elementHeight;
+    }
 
     if (!open) {
       return <div style={styles.hidden} />;
     }
 
     return (
-      <div style={styles.backdrop} onClick={this.handleBackdropClick}>
+      <div style={styles.backdrop} onClick={this.handleBackdropClick} ref={this.getHeight}>
```
bvaughn: nit: Not a fan of the name of this method. It isn't a getter; it updates state. It's also a bit weird to have a ref-setter named "get" something. My preference would be to call it something like `_setRef`

frontend/ContextMenu.js
LanceJernigan: I'm honestly not sure how those even got added to this branch.  I didn't realize they were there until I made the pull request.

frontend/ContextMenu.js
LanceJernigan: The spacing should be fixed, I believe that was due to vscode's selection formatting.

frontend/ContextMenu.js
bvaughn: No worries. I assume you just branched from your other branch instead of master, maybe accidentally it sounds like 😁

bvaughn:
I noticed a jumping issue that happens if you right click a node before first focusing on it (via left click):
context-menu-jump-alignment

This also happens in master though, and your alignment fix definitely improves the bottom overflow issue in master. So I think we can just tackle that in a follow up. (I mention it here just because I noticed it. Not because I expect you to fix it or anything. 😄 )

LanceJernigan:
I noticed that, too.  I just assumed it was to scroll the element into view a little better instead of having it so close to the bottom of the container.  I wasn't sure if it was a bug or a feature.

bvaughn:
Yeah, the scroll-into-view bit is a feature. It's just unfortunate how the jump can cause the node and context menu to be misaligned slightly. 😁  But not a big deal. This PR is a def improvement.

LanceJernigan:
Gotcha.  If I have time, I'll try to look into a way to keep the two lined up a little better and submit a separate PR for it.

bvaughn:
That would be awesome but only if you're interested and have the time. 🙇 
I'm almost done testing this. I think it looks good.

bvaughn:
Thanks for this contribution!

PR #770, opened 2017-06-01
Add an argument to selectFromDOMNode() to skip leafs

gaearon:
This reverts a bandaid fix that attempted to work around a crash, and instead fixes the root issue (with another PR to RN).

The issue is that passing Fiber from outside to `selectFromReactInstance()` didn't work because it could potentially be an alternate, and our [deduplication logic](https://github.com/facebook/react-devtools/blob/575e416f45aa9d2898f7493b0266e306f9d0646e/backend/attachRendererFiber.js#L27-L38) wouldn't work for a [direct call like this](https://github.com/facebook/react-native/blob/a93b7a2da0849f15708afef1ff99022d8cc55b14/Libraries/Inspector/Inspector.js#L179).

Instead I'm adding support to `selectFromDOMNode()` to skip a few leafs (which was original motivation of [exposing](https://github.com/facebook/react/blob/9dcc60af3394f5e5f92a7cc9d54a68ce48795681/src/renderers/native/ReactNativeFiberInspector.js#L107) the `instance`). This lets us change RN code to just call:

```js
    if (this.state.devtoolsAgent) {
      // Skip host leafs
      const offsetFromLeaf = hierarchy.length - 1 - selection;
      this.state.devtoolsAgent.selectFromDOMNode(touchedViewTag, true, offsetFromLeaf);
    }
```

This doesn't pass a fiber, so we don't get the "wrong fiber" issue.

### DIFF
--- a/agent/Agent.js
+++ b/agent/Agent.js
@@ -272,14 +272,18 @@ class Agent extends EventEmitter {
     return null;
   }
 
-  selectFromDOMNode(node: Object, quiet?: boolean) {
+  selectFromDOMNode(node: Object, quiet?: boolean, offsetFromLeaf: ?number = 0) {
     var id = this.getIDForNode(node);
     if (!id) {
       return;
     }
-    this.emit('setSelection', {id, quiet});
+    this.emit('setSelection', {id, quiet, offsetFromLeaf});
   }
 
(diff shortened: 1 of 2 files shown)


gaearon:
RN counterpart: https://github.com/facebook/react-native/pull/14290.

PR #771, opened 2017-06-01
Added Firebug Theme

LanceJernigan:
@bvaughn this is about as close as I can get the Firebug theme due to theming differences.  Let me know if I need to tweak anything to try to get it closer.

### DIFF
--- a/frontend/Themes/Themes.js
+++ b/frontend/Themes/Themes.js
@@ -225,8 +225,9 @@ const FirefoxLight: Theme = {
   state05: '#585959',
 };
 
-const Firebug: Theme = {
-  displayName: "Firebug",
+const FirefoxFirebug: Theme = {
+  displayName: "Firefox (Firebug)",
+  hidden: true,
   base00: "#ffffff",
   base01: "#f5f5f5",
   base02: "#dde1e4",
@@ -244,7 +245,7 @@ const Firebug: Theme = {
   state00: "#3399ff",
   state01: "#85c0fb",
   state02: "#ffffff",
-  state03: "#e0e0e0",
+  state03: "#e6e6e6",
   state04: "#3399ff",
   state05: "#ffffff",
 }
--- a/frontend/Themes/Themes.js
+++ b/frontend/Themes/Themes.js
@@ -225,6 +225,30 @@ const FirefoxLight: Theme = {
   state05: '#585959',
 };
 
+const Firebug: Theme = {
+  displayName: "Firebug",
+  base00: "#ffffff",
+  base01: "#f5f5f5",
+  base02: "#dde1e4",
+  base03: "#c1c1c1",
+  base04: "#888888",
+  base05: "#2a2a2a",
+  special00: "#0000ff",
+  special01: "#ff0000",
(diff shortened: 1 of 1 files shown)


frontend/Themes/Themes.js
```diff
@@ -450,6 +474,7 @@ module.exports = {
   Eighties,
   FirefoxDark,
   FirefoxLight,
+  Firebug,
```
bvaughn: nit: Let's name this `FirefoxFirebug` to keep it better grouped with the other Firefox themes?

frontend/Themes/Themes.js
```diff
@@ -225,6 +225,30 @@ const FirefoxLight: Theme = {
   state05: '#585959',
 };
 
+const Firebug: Theme = {
+  displayName: "Firebug",
```
bvaughn: I think this should be `hidden: true` like Firefox light and dark themes.

frontend/Themes/Themes.js
```diff
@@ -225,6 +225,30 @@ const FirefoxLight: Theme = {
...
+  special02: "#ff0000",
+  special03: "#292e33",
+  special04: "#ff0000",
+  special05: "#0000ff",
+  special06: "#000062",
+  special07: "#0000ff",
+  state00: "#3399ff",
+  state01: "#85c0fb",
+  state02: "#ffffff",
+  state03: "#e0e0e0",
```
bvaughn: nit: I think this is actually `#e6e6e6` in Firebug

frontend/Themes/Themes.js
LanceJernigan: Should the `displayName` be `Firefox Firebug` or something different?

bvaughn: Hm. Probably. Is that too weird?

LanceJernigan: That's what I was questioning, too.  I think, whichever decision we go with, the displayName and the key should reflect each other.  If we use `FirefoxFirebug`, I think the displayName should be either `Firefox Firebug` or `Firefox (Firebug)`.

frontend/Themes/Themes.js
```diff
@@ -225,6 +225,31 @@ const FirefoxLight: Theme = {
...
+  special03: "#292e33",
+  special04: "#ff0000",
+  special05: "#0000ff",
+  special06: "#000062",
+  special07: "#0000ff",
+  state00: "#3399ff",
+  state01: "#85c0fb",
+  state02: "#ffffff",
+  state03: "#e6e6e6",
+  state04: "#3399ff",
```
bvaughn: Could we match the yellow search color? (`#ffee99`)

screen shot 2017-06-01 at 11 07 01 am

LanceJernigan: I changed the background to yellow and the foreground to black since the white wasn't enough of a contrast.  Any suggestions on the foreground or does black work?

bvaughn: Black works fine! Thanks!

I'm going to patch up the linting issues and merge. Thanks so much for the PR and for being so responsive!

PR #778, opened 2017-06-01
[Electron] Add setBrowserName method

jhen0409:
Related to https://github.com/facebook/react-devtools/pull/751#issuecomment-304695648, this make the Electron shell can custom the name for top option of theme dropdown, mirror the functionality for clear `localStorage.themeName`.

### DIFF
--- a/frontend/PreferencesPanel.js
+++ b/frontend/PreferencesPanel.js
@@ -92,7 +92,7 @@ class PreferencesPanel extends React.Component {
             >
               {browserName && (<option value="">{browserName}</option>)}
               {hasCustomTheme && (<option value={CUSTOM_THEME_NAME}>Custom</option>)}
-              {(!showHiddenThemes || hasCustomTheme) && <option disabled="disabled">---</option>}
+              {(browserName || hasCustomTheme) && <option disabled="disabled">---</option>}
               {themeNames.map(key => (
                 <option key={key} value={key}>{themes[key].displayName}</option>
               ))}
--- a/frontend/PreferencesPanel.js
+++ b/frontend/PreferencesPanel.js
@@ -90,7 +90,7 @@ class PreferencesPanel extends React.Component {
               ref={this._setSelectRef}
               value={themeName}
             >
-              {!showHiddenThemes && (<option value="">{browserName}</option>)}
+              {browserName && (<option value="">{browserName}</option>)}
               {hasCustomTheme && (<option value={CUSTOM_THEME_NAME}>Custom</option>)}
               {(!showHiddenThemes || hasCustomTheme) && <option disabled="disabled">---</option>}
               {themeNames.map(key => (
--- a/packages/react-devtools-core/src/standalone.js
+++ b/packages/react-devtools-core/src/standalone.js
@@ -200,6 +200,11 @@ var DevtoolsUI = {
     return DevtoolsUI;
   },
 
+  setBrowserName(name) {
+    config.browserName = name;
+    return DevtoolsUI;
+  },
+
   startServer,
 };
 


frontend/PreferencesPanel.js
```diff
@@ -90,7 +90,7 @@ class PreferencesPanel extends React.Component {
               ref={this._setSelectRef}
               value={themeName}
             >
-              {!showHiddenThemes && (<option value="">{browserName}</option>)}
+              {browserName && (<option value="">{browserName}</option>)}
```
bvaughn: This change would make the UX a bit weird.

The themes menu would have both "_Chrome_" and "_Chrome (default)_" and "_Chrome (dark)_" options, and selecting "_Chrome_" would actually select one of the other 2. That's confusing.

I wonder if `setBrowserName` should also toggle `showHiddenThemes` to false? That would also be a bit weird I guess, if the user had pre-selected one of the "hidden" themes (eg maybe the user _prefers_ to run with a Firefox theme even though "Chrome" has been injected as the `browserName`.)

To be honest I'm still a little unclear on the benefit of this feature (matching theme between standalone devtools and a browser).

jhen0409: > To be honest I'm still a little unclear on the benefit of this feature (matching theme between standalone devtools and a browser). Am I misunderstanding it?

The context is I made the `match devtools` feature for [`react-native-debugger`](https://github.com/jhen0409/react-native-debugger), see the screenshot:



The problem is the `localStorage.themeName` must be empty if we want to tweak theme automatically , but it will never tweak if user selected a theme, so we need a way to back.

So it will not called as `Chrome` in this case, or some case for whos use the standalone devtools with the added method of this PR.

bvaughn: Thanks for providing additional context and a demo gif.

I'm still not sure I understand why someone would set a theme in the way shown in your gif above vs using the built-in theme setting. It seems...unnecessarily indirect. Which is fine, except that it adds complexity to code that's already a little more complex than I'd prefer.

I find myself questioning whether the added complexity is justified.

jhen0409: > I'm still not sure I understand why someone would set a theme in the way shown in your gif above vs using the built-in theme setting. It seems...unnecessarily indirect.

Because the main usage of react-native-debugger is devtools, it should have an option similar to the browser extension have `match browser`.

It may just be a unilateral request, not sure if people need to go back to default theme for `react-devtools` package. (It had a `Reset` button but removed)

I reverted https://github.com/facebook/react-devtools/pull/778/commits/8f2c932528da76bf15954e5be6415a1cbab3fce8 because It may not be necessary.

packages/react-devtools-core/src/standalone.js
```diff
@@ -200,6 +200,14 @@ var DevtoolsUI = {
     return DevtoolsUI;
   },
 
+  setBrowserName(name) {
+    config.browserName = name;
+    if (panel) {
+      panel.props.browserName = name;
```
gaearon: We shouldn't mutate props like this. Do you have an alternative solution that re-renders panel with the right props?

bvaughn: One way to do this might be to move `browserName` to the `Store` and then use a similar approach as with `setDefaultThemeName`:
```diff
diff --git a/frontend/Panel.js b/frontend/Panel.js
index 34ed519..e164619 100644
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -100,7 +100,7 @@ class Panel extends React.Component {
 
   getChildContext(): Object {
     return {
-      browserName: this.props.browserName || '',
+      browserName: this._store && this._store.browserName || '',
       defaultThemeName: this._themeStore && this._themeStore.defaultThemeName || '',
       showHiddenThemes: !!this.props.showHiddenThemes,
       showInspectButton: this.props.showInspectButton !== false,
@@ -221,7 +221,11 @@ class Panel extends React.Component {
       this._bridge = new Bridge(wall);
 
       this._themeStore = new ThemeStore(this.state.themeName);
-      this._store = new Store(this._bridge, this._themeStore);
+      this._store = new Store(
+        this._bridge,
+        this._themeStore,
+        this.props.browserName,
+      );
       
       var refresh = () => this.forceUpdate();
       this.plugins = [
diff --git a/frontend/PreferencesPanel.js b/frontend/PreferencesPanel.js
index d5473bd..8fc5d2a 100644
--- a/frontend/PreferencesPanel.js
+++ b/frontend/PreferencesPanel.js
@@ -24,13 +24,13 @@ class PreferencesPanel extends React.Component {
   _selectRef: any;
 
   context: {
-    browserName: string,
     showHiddenThemes: boolean,
     theme: Theme,
     themeName: string,
     themes: { [key: string]: Theme },
   };
   props: {
+    browserName: string,
     changeTheme: (themeName: string) => void,
     hasCustomTheme: boolean,
     hide: () => void,
@@ -61,8 +61,8 @@ class PreferencesPanel extends React.Component {
   }
 
   render() {
-    const {browserName, showHiddenThemes, theme, themeName, themes} = this.context;
-    const {hasCustomTheme, hide, open} = this.props;
+    const {showHiddenThemes, theme, themeName, themes} = this.context;
+    const {browserName, hasCustomTheme, hide, open} = this.props;
     const {editMode} = this.state;
 
     if (!open) {
@@ -161,13 +161,13 @@ class PreferencesPanel extends React.Component {
 }
 
 PreferencesPanel.contextTypes = {
-  browserName: React.PropTypes.string.isRequired,
   showHiddenThemes: React.PropTypes.bool.isRequired,
   theme: React.PropTypes.object.isRequired,
   themeName: React.PropTypes.string.isRequired,
   themes: React.PropTypes.object.isRequired,
 };
 PreferencesPanel.propTypes = {
+  browserName: React.PropTypes.string.isRequired,
   changeTheme: React.PropTypes.func,
   hide: React.PropTypes.func,
   open: React.PropTypes.bool,
@@ -184,10 +184,11 @@ const blockClick = event => event.stopPropagation();
 
 const WrappedPreferencesPanel = decorate({
   listeners() {
-    return ['preferencesPanelShown'];
+    return ['browserName', 'preferencesPanelShown'];
   },
   props(store, props) {
     return {
+      browserName: store.browserName || '',
       changeTheme: themeName => store.changeTheme(themeName),
       hasCustomTheme: !!store.themeStore.customTheme,
       hide: () => store.hidePreferencesPanel(),
diff --git a/frontend/Store.js b/frontend/Store.js
index 67b4465..6c9ff70 100644
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -89,6 +89,7 @@ class Store extends EventEmitter {
   _eventTimer: ?number;
 
   // Public state
+  browserName: ?string;
   isInspectEnabled: boolean;
   traceupdatesState: ?ControlState;
   colorizerState: ?ControlState;
@@ -112,7 +113,11 @@ class Store extends EventEmitter {
     rnStyleMeasure?: boolean,
   };
 
-  constructor(bridge: Bridge, themeStore: ThemeStore) {
+  constructor(
+    bridge: Bridge,
+    themeStore: ThemeStore,
+    browserName: ?string,
+  ) {
     super();
 
     this._nodes = new Map();
@@ -121,6 +126,7 @@ class Store extends EventEmitter {
     this._bridge = bridge;
 
     // Public state
+    this.browserName = browserName;
     this.isInspectEnabled = false;
     this.roots = new List();
     this.contextMenu = null;
@@ -345,6 +351,11 @@ class Store extends EventEmitter {
     this.emit('theme');
   }
 
+  setBrowserName(browserName: ?string) {
+    this.browserName = browserName;
+    this.emit('browserName');
+  }
+
   setDefaultThemeName(defaultThemeName: ?string) {
     this.themeStore.setDefaultThemeName(defaultThemeName);
   }
diff --git a/packages/react-devtools-core/src/standalone.js b/packages/react-devtools-core/src/standalone.js
index 8b28055..231a109 100644
--- a/packages/react-devtools-core/src/standalone.js
+++ b/packages/react-devtools-core/src/standalone.js
@@ -199,10 +199,12 @@ var DevtoolsUI = {
     return DevtoolsUI;
   },
 
-  setBrowserName(name) {
-    config.browserName = name;
+  setBrowserName(browserName) {
+    config.browserName = browserName;
     if (panel) {
-      panel.props.browserName = name;
+      var {store} = panel.getChildContext();
+      // Change browserName if panel already mounted
+      store.setBrowserName(browserName);
     }
     return DevtoolsUI;
   },
```

frontend/PreferencesPanel.js
bvaughn: > It may just be a unilateral request, not sure if people need to go back to default theme for `react-devtools` package. (It had a `Reset` button but removed)

Yeah, I thought replacing the "Reset" button with the name of the browser, particularly given the addition of the new "Custom" option, made more sense in terms of UX. (Subjective call.)

frontend/PreferencesPanel.js
```diff
@@ -90,9 +90,10 @@ class PreferencesPanel extends React.Component {
...
               {hasCustomTheme && (<option value={CUSTOM_THEME_NAME}>Custom</option>)}
-              {(!showHiddenThemes || hasCustomTheme) && <option disabled="disabled">---</option>}
+              {(browserName || !showHiddenThemes || hasCustomTheme) &&
```
bvaughn: Given that the above conditional changed from `!showHiddenThemes` to `browserName` - I think this check should change to just `(browserName || hasCustomTheme)` ?

packages/react-devtools-core/src/standalone.js
jhen0409: If we do [`reload()`](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools-core/src/standalone.js#L46) it will lose current session of RN (Shown `Waiting for roots to load...`), maybe just don't update it? I think call the method before `startServer` is enough in this case.

bvaughn: > I think call the method before startServer is enough in this case.

If so, that sounds good to me as it reduces churn a bit.

frontend/PreferencesPanel.js
jhen0409: > I think this check should change to just (browserName || hasCustomTheme) ?

You're right, I'll change it.

PR #780, opened 2017-06-01
Add `innerRef` prop for Input & fix all ref issues

jhen0409:
It looks #761 breaks all input refs because it just get the wrapper. 

cc @bvaughn


bvaughn:
D'oh! Nice catch @jhen0409! Thanks! 🙇

Issue #612, opened 2017-04-20
"Down" button doesn't always work in Fiber mode


gaearon:
I can't navigate past some `<div>`s with the down button if the project uses Fiber. I think this is related to the hacky way I implemented fragment support (which are used in Fiber in more cases).

gaearon:
To reproduce this you'd need to use React DevTools in a project using React 16 alphas.

Tushkiz:
Hey @gaearon,

I tried to reproduce this issue, it is working for me.
I am using 16.0.0-alpha.3 and created an app with 'create-react-app'

gaearon:
I'll see if I can get a full example reproducing this. I think it happened for me when going "down" into an array.

Tushkiz:
Ohh, with an array I see that while going up, it jumps to the first element from that array. I think these are related.

![going_up](https://cloud.githubusercontent.com/assets/1854763/25427310/c785987c-2a8f-11e7-8235-96b6f3fd1b2d.PNG)

Example:
```
class App extends Component {
  render() {
    return (
      <div className="App">
        <div className="App-header">
          <h2>Welcome to React</h2>
        </div>
        {
          ["foo", "bar", "baz", "fiber"].map((str) => <div key={str}>{str}</div>)
        }
      </div>
    );
  }
}
```

But if I wrap those array elements, it works.

```
class App extends Component {
  render() {
    return (
      <div className="App">
        <div>
          <h2>Welcome to React</h2>
        </div>
        <div>
          {
            ["foo", "bar", "baz", "fiber"].map((str) => <div key={str}>{str}</div>)
          }
        </div>
      </div>
    );
  }
}
```

trueadm:
@Tushkiz can you try out my PR (https://github.com/facebook/react-devtools/pull/781) and tell me if this fixes this issue for you? Thanks!

Tushkiz:
@trueadm Yup! trying it out.

Tushkiz:
@trueadm great, this fixes my use case 🎉

PR #781, opened 2017-06-02
Fix keyboard nav

trueadm:
This should fix the issues in https://github.com/facebook/react-devtools/issues/612

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -495,7 +495,7 @@ class Store extends EventEmitter {
   skipWrapper(id: ElementID, up?: boolean, end?: boolean): ?ElementID {
     if (!id) {
       return undefined;
-    }    
+    }
     while (true) {
       var node = this.get(id);
       var nodeType = node.get('nodeType');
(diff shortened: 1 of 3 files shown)


gaearon:
Seems like there's still a bug:



Reproduces consistently for me on first Workplace UFIAsyncWrapper.

bvaughn:
On facebook.com, clicking on any of the Like/Comment nodes:
screen shot 2017-06-02 at 9 06 29 am

Then up/down arrowing:
skip-null

Temporarily focuses on a "null" node (and hides the focused/selection color) for one up/down arrow press.

gaearon:
Is this a different issue? Seems like the same div.

bvaughn:
Ah! I think you're right. Sorry. I had my comment open/drafting for a while before I submitted it. Hadn't noticed you'd tagged the same thing above. Sorry.

gaearon:
Haha it’s cool 😄 I was wondering if I’m missing an additional detail.

bvaughn:
Filed #786 for the follow-up work. The rest of this looks solid. Let's go!

PR #787, opened 2017-06-04
Button cursors and hover-styles

bvaughn:
@gaearon pointed out that several of the devtools buttons used the `pointer` cursor style when they should probably be using the `default` style instead. My original reasons for this were that they don't look like normal buttons or otherwise provide visual cues that they are clickable (eg Chrome and Firefox icons change color on hover).

This PR removes the `pointer` cursor in favor of adding hover colors via a new `Hoverable` HOC. To make this work I also had to add a new theme color, `state06`. I backfilled this color in all existing themes so they would work automatically.

### Settings and Inspect icons
hover-color

### Clear search icon
clear-search-hover

### Edit theme icon
pencil-hover

### DIFF
--- a/frontend/Themes/Editor/ColorInput.js
+++ b/frontend/Themes/Editor/ColorInput.js
@@ -180,7 +180,6 @@ const colorChipStyle = (theme: Theme, color: string = '', showBorder: boolean = 
   borderRadius: '2px',
   backgroundColor: color,
   boxSizing: 'border-box',
-  cursor: 'pointer',
   border: showBorder ? `1px solid ${theme.base03}` : 'none',
 });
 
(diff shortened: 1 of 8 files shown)


frontend/Themes/Editor/ColorInput.js
```diff
@@ -77,6 +77,8 @@ class ColorInput extends React.Component {
 
     const backgroundIsBright = isBright(theme.base00);
     const chipIsBright = isBright(color);
+    const showColorChipBorder = backgroundIsBright === chipIsBright &&
+      getBrightness(color) > getBrightness(theme.base03);
```
bvaughn: This change wasn't really related to the overall PR but I noticed it while testing the hover colors and thought I'd fix it while I was in the file.

bvaughn:
Wille up and realized I hadn't verified that custom theme handled this change well. So I added a to-do for myself to hit in the morning. Should be super simple.

gaearon:
This is really cool, thank you!

bvaughn:
Thanks @gaearon. This is ready for review now.

PR #788, opened 2017-06-05
Further keyboard nav fixes

trueadm:
This fixes the last remaining issue reported in https://github.com/facebook/react-devtools/pull/781

### DIFF
--- a/frontend/keyboardNav.js
+++ b/frontend/keyboardNav.js
@@ -112,7 +112,7 @@ function getNewSelection(dest: Dest, store: Store): ?ElementID {
     return undefined;
   }
   var node = store.get(id);
-  var pid = store.skipWrapper(store.getParent(id), true);
+  var pid = store.getParent(id);
 
   if (pid) {
     var lastId = id;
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks! While we're at it, do you mind also fixing the indentation in this file?

screen shot 2017-06-05 at 3 15 11 pm

trueadm:
@gaearon fixed :)

frontend/keyboardNav.js
```diff
@@ -104,15 +104,15 @@ function getRootSelection(dest, store, id) {
...
-	}
+}
 
 function getNewSelection(dest: Dest, store: Store): ?ElementID {
   var id = store.selected;
   if (!id) {
     return undefined;
   }
   var node = store.get(id);
-  var pid = store.skipWrapper(store.getParent(id), true);
```
gaearon: Can you help me understand why this call is no longer necessary? What was it doing?

trueadm: It was actually what I was doing before, but failed to copy it over to my PR (it was working on the example I showed you locally on my Mac) in the compiled bundle. N00b error.

We don't want to skip the wrapper in case we're needing to traverse an actual wrapper. This breaks it.

gaearon:
This seems to work great. Thank you!

PR #789, opened 2017-06-05
Theme Editor: Use percent instead of viewport for max size

jhen0409:
Currently the theme editor size is not depends on Panel when use `react-devtools-core` and put it to the different position, see the following screenshot:

Use percent for max size will fix this.

### DIFF
--- a/frontend/Themes/Editor/Editor.js
+++ b/frontend/Themes/Editor/Editor.js
@@ -193,8 +193,8 @@ const WrappedEditor = decorate({
 const editorStyle = (theme: Theme) => ({
   display: 'flex',
   flexDirection: 'column',
-  maxWidth: 'calc(100vw - 2rem)',
-  maxHeight: 'calc(100vh - 2rem)',
+  maxWidth: 'calc(100% - 2rem)',
+  maxHeight: 'calc(100% - 2rem)',
   boxSizing: 'border-box',
   zIndex: 1,
   padding: '0.5rem',


bvaughn:
I had to use `vh` and `vw` units for Firefox. Percentage based size caused vertical alignment problems for that browser.

You can see this problem with your diff/branch by running in Firefox:
screen shot 2017-06-05 at 9 52 59 am

gaearon:
cc @clarkbw regarding the FF issue

bvaughn:
> Currently the theme editor position is not depends on Panel when use react-devtools-core and put it to the different position, see the following screenshot:

I also don't really understand the description of this PR. Is the problem lack of view unit support? Those have been around for a while (IE9+) right?

jhen0409:
> I also don't really understand the description of this PR. Is the problem lack of view unit support? Those have been around for a while (IE9+) right?

The problem is `vw` / `vh`  is use the viewport (window content) size, not parent view size, so we can see Its size is more than the react-devtools panel, this is a little problem for use the package. (The above should be mean `the theme editor size` I edited it, sorry for the typo)

bvaughn:
Ah, I understand. So you want to use the devtools/editor within a subset of the window/viewport.

I guess wrapping an `<iframe>` around devtools would be too much of a hack? 😅 

To be honest I'm surprised there aren't more issues like this that crop up, given that you're running the UI within another UI.

jhen0409:
Maybe we can just use `fixed` instead of `absolute` for position of the `PreferencesPanel` content? Make it just like a modal on the window for me is reasonable.

bvaughn:
Check out PR #799

bvaughn:
I think #799 is a more robust change and has the added benefit of not causing problems with Firefox. I'm going to close this PR in favor of that one. Let's continue chatting over there, @jhen0409 😄

Issue #783, opened 2017-06-02
Right and left arrows behave unexpectedly when tail tag is focused


gaearon:
The right and left arrow keys behave unexpectedly when a tail tag of a node with children is selected.

Note that Firefox doesn't really let you select a tail tag. (Keyboard nav skips it. Mouse clicks redirect to the head tag.) So we'll compare to Chrome's behavior.

#### Chrome
Chrome jumps to the head tag.
![chrome-left](https://cloud.githubusercontent.com/assets/29597/26733766/b510a378-4770-11e7-9c0d-ad5f38dd5b1c.gif)

#### DevTools
DevTools collapses the node. I don't feel strongly about this but it might be nice to match Chrome's behavior.
![devtools-left](https://cloud.githubusercontent.com/assets/29597/26733909/2730f84a-4771-11e7-9981-cab2bad6f040.gif)

#### Chrome
Chrome does nothing in this case.

#### DevTools
DevTools steps into the tree (into children, moving the cursor up). This feels unnatural.
![devtools-right](https://cloud.githubusercontent.com/assets/29597/26733923/36ff1d42-4771-11e7-94be-6b11dcc15aaa.gif)

jobala:
@gaearon,

I can pick this up, mainly because I have free time on my hands.
And willing to learn.

How do I get started.

gaearon:
Hey, thanks for offering to help! The related logic is in `frontend/keyboardNav.js`. You can follow "Contributing" instructions in README to set up the enrichment.

bvaughn:
Sweet! Feel free to ping me on Twitter if you hit a snag @jobala 😄

jobala:
@bvaughn I will.

jobala:
@bvaughn,

I git cloned, the repo to my local machine.
I will need to be able to see the effect of the changes I am making.

How do I set it up, ie like a chrome extension running from my machine?
Or better yet, you could point me to a resource that has this information.

Could be a dumb question, but yeah 😅

jobala:
Otherwise,

I have done some digging and think this could be a possible solution.

I followed a rabbit hole from keyboardNav.js to dirToDest.js

for the right arrow `(dirToDest.js line 39-50)`
```
...
if (hasChildren) {
  if (bottom) {
     // This is the cause for our first bug, if we are at the bottom of a tree,
     // return the last child of the tree. A solution would be to return...uhm...
     // not really sure what to return perhaps an empty string or just return null
     return 'lastChild'
[...]

PR #792, opened 2017-06-05
Fix left right arrow behavior

jobala:
#### What does this PR do
* When the left arrow is clicked on the closing tag, the cursor jumps to the head tag.
* When the right arrow is clicked on the closing tag, it does nothing instead of the cursor
focusing on the last child.

Fixes [783](https://github.com/facebook/react-devtools/issues/783)

### DIFF
--- a/frontend/dirToDest.js
+++ b/frontend/dirToDest.js
@@ -42,7 +42,7 @@ module.exports = function(dir: Dir, bottom: boolean, collapsed: boolean, hasChil
     }
     if (hasChildren) {
       if (bottom) {
-        return 'lastChild';
+        return null;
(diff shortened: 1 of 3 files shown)


gaearon:
I think it currently breaks collapsing. I can't seem to press "left" on the opening tag to collapse it.

trueadm:
@jobala I just tried your PR out locally and I can't close tags by pressing left anymore on the opening tag (I can on master). Does this happen for you locally too?

gaearon:
I fixed that. Thank you so much for the PR!

bvaughn:
Nice work @jobala !

PR #799, opened 2017-06-06
Preferences panel position is fixed now instead of absolute

bvaughn:
This is to help applications that include devtools core inside of a larger UI.

cc @jhen0409

This improvement can also be seen in the plain shell.

#### Before
screen shot 2017-06-06 at 9 35 05 am

#### After
screen shot 2017-06-06 at 9 34 52 am


jhen0409:
Thank you! 😀

bvaughn:
No problem! Thanks for working with me on it! 😁

PR #632, opened 2017-04-25
npm script for bundling chrome extension (WIP)

Tushkiz:

PR #802, opened 2017-06-06
Firefox WebExtension (and themeName support)

### LINKED ISSUE #705: Themes follow up work

bvaughn:
Resolves #705 and #524

**Note** that this change will mean dropping support for Firefox <= 53 with future releases of the extension. I think this is the right decision to make going forward. **Firefox 54 has been released.** ~~Firefox 54 is [targeted for release on 2017-06-13](https://wiki.mozilla.org/RapidRelease/Calendar). I think it's worth _reviewing_ this diff now, but maybe we should hold off on merging it until the 13th.~~

# tl;dr
* Delete old Firefox add-ons directory
* Move Chrome extension to new `shells/webextension` directory
* Replace old build script with new Node script
  * Can generate both Firefox and Chrome web extensions
  * Runs via `npm run build:extension`
* Add support for Firefox theme mirroring

# Firefox themeName (#705)
The default Firefox theme will work as before for Firefox 54 (and earlier). As of the upcoming Firefox 55, DevTools will mimic Firefox's active theme. (This can currently be tested against the Firefox Nightly release.)

Firefox theme switching

#### References
* [Firefox issue 1349896](https://bugzilla.mozilla.org/show_bug.cgi?id=1349896)
* [MDN add-ons doc](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/devtools.panels/themeName)

# Change to WebExtensions (#524)
Firefox recently added a new API for `browser.devtools.panels.themeName`. After some testing it became apparent that this API is only available for WebExtensions and not add-ons (see [here](https://github.com/facebook/react-devtools/pull/802#discussion_r120537356)). Since Firefox will be dropping add-on support in the future anyway ([by the end of the year](https://blog.mozilla.org/addons/2016/11/23/add-ons-in-2017/)) it seems the way forward is to switch over to using the WebExtensions build for Firefox as well.

To do this, we could fork the Chrome extension. However there's no differences other than the injected theme name that I'm aware of, so I don't think that's necessary (cc @clarkbw for confirmation). Instead I opted for updating the [Firefox shell instructions](https://github.com/facebook/react-devtools/blob/b64fd01db456f2966e6fa0f2d89a9a278d159400/shells/firefox/README.md) to reference the Chrome shell instead. (WebExtensions support was [first previewed in Firefox 42](https://blog.mozilla.org/addons/2015/08/21/the-future-of-developing-firefox-add-ons/), in 2015, so hopefully this won't cause any backwards compatibility issues either?)

Another nice thing about switching over Firefox to use the WebExtension is that the DevTools dev-mode icon will work:

screen shot 2017-06-07 at 9 13 33 am
screen shot 2017-06-07 at 9 13 40 am

Here's the output of the new build script. (Isn't it pretty? 😅)
build-updated-3


clarkbw:
Looks good so far!

shells/firefox/panel/run.js
```diff
@@ -21,10 +23,21 @@ var node = document.getElementById('container');
...
-// TODO (bvaughn) Read default :themeName and switch between 'FirefoxLight' and 'FirefoxDark'
-// Refer to 'shells/chrome/src/panel.js' for an example
 const browserName = 'Firefox';
-const themeName = 'FirefoxLight';
+let themeName = 'FirefoxLight';
+
+// If available, mirror the current Firefox devtools theme:
+// https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/devtools.panels/themeName
+if (typeof devtools !== 'undefined' && devtools.panels) {
+  switch (devtools.panels.themeName) {
```
clarkbw: Should this be browser.devtools.panels.themeName?

bvaughn: I tried `devtools.panels`, `browser.devtools.panels`, and `chrome.devtools.panels` but all were undefined. Based on the docs (which mention all of the above) I took a guess with `devtools.panels` 😅

clarkbw: I'm now wondering if this is because this code is an add-on sdk environment and that API is part of the WebExtensions.  `browser.devtools.panels.themeName` should work if you load the Chrome WebExtension into Firefox.

bvaughn: I'm actually not sure what that means 😁

bvaughn: Ah! Figured it out. Seems like I need to mimic the structure in shells/chrome for firefox, and then run [`web-ext run --firefox=/Applications/FirefoxNightly.app/Contents/MacOS/firefox-bin`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext) from the directory.

bvaughn:
Score! Running the Chrome extension in Firefox, but switching on FF's theme names:

firefox-theme

shells/firefox/panel/run.js
clarkbw: I looked away and you went all the way down the 🐰 hole! 😲

bvaughn:
Hey @clarkbw 😁  I just noticed that this PR has shifted in such a way as to overlap with #524

Seems like there's some issue for non-nightly builds (mentioned in the description) but otherwise it seems like the Chrome shell works fine in Firefox. Maybe we could combine forces and finish this effort?

Probably worth renaming `shells/chrome` to something else at some point.

shells/chrome/src/panel.js
```diff
@@ -17,12 +17,36 @@ var inject = require('./inject');
 
 import type {Props} from '../../../frontend/Panel';
 
-// chrome.devtools.panels added in Chrome 18.
-// chrome.devtools.panels.themeName added in Chrome 54.
-const browserName = 'Chrome';
-const themeName = (chrome.devtools.panels : any).themeName === 'dark'
-  ? 'ChromeDark'
-  : 'ChromeDefault';
+const IS_CHROME = navigator.userAgent.indexOf('Firefox') < 0;
```
bvaughn: Is this too clunky? Should I try to inject this parameter via the build script (or similar)?

clarkbw:
> Maybe we could combine forces and finish this effort?

💯 

> Probably worth renaming shells/chrome to something else at some point. But that could be done in a follow-up to avoid adding noise to this.

Agreed

bvaughn:
Don't suppose you have any insight into what's going on in FF <= 43?

screen-shot-2017-06-07-at-11 02 54-am

bvaughn:
Ah, nevermind. I think it's b'c we have a reference to `chrome.devtools.network` elsewhere that isn't being guarded against.

clarkbw:
> Testing in Chrome and Firefox Nightly, this seems to work great. Older versions of Firefox I have locally (47, 53) don't quite work though so I still need to look into that

[bug 1300590](https://bugzilla.mozilla.org/show_bug.cgi?id=1300590) only just landed in Firefox 55 and is required for features related to inspecting and console. And [bug 1300587](https://bugzilla.mozilla.org/show_bug.cgi?id=1300587) landed in 54 which allows for creating the panel.  We might need to set the min version at Fx 55

bvaughn:
Hm. Yeah. That seems reasonable I guess. We have too many references we'd have to guard against otherwise.

clarkbw:
> Ah, nevermind. I think it's b'c we have a reference to chrome.devtools.network elsewhere that isn't being guarded against.

If it's `chrome.devtools.network.onNavigated` that should also be working in Firefox.  I haven't tested it directly but according to [the docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/devtools.network/onNavigated) it is there and I haven't seen errors previously.  

I have been seeing errors within the `doublePipe` function which I suspect is causing problems.  It appears that one of the pipe references is being lost and I'm not sure why yet.

shells/chrome/src/panel.js
clarkbw: This is the same thing we did for the redux extension https://github.com/zalmoxisus/redux-devtools-extension/commit/0cf6830978dafdcb8c2a2d5d935e136f110a862f

bvaughn:
>> Ah, nevermind. I think it's b'c we have a reference to chrome.devtools.network elsewhere that isn't being guarded against.

> If it's chrome.devtools.network.onNavigated that should also be working in Firefox. I haven't tested it directly but according to the docs it is there and I haven't seen errors previously.

`chrome.devtools.network` is undefined for FF 53.

shells/chrome/src/panel.js
bvaughn: Oh, nice! 😁

clarkbw:
> The docs show it as being added in 54.

Oh right! Fx 54 starts rolling out next week.

clarkbw:
We'll need to figure out a way to use a separate manifest file for the Firefox version so we can set the min Fx version needed.

```json
  "applications": {
    "gecko": {
      "strict_min_version": "55"
    }
  }
```

More details [here](https://github.com/facebook/react-devtools/issues/524#issuecomment-298096071) and [here](https://github.com/facebook/react-devtools/issues/524#issuecomment-298403906). However, like you pointed out, that could be done with something like #632

I'd imagine the build steps are something like this:
* Build Chrome extension using the default build directory and zip it up as `chrome.zip`
* Build Firefox extension with the `application`attribute injected into the manifest and zip it up as `firefox.zip`

bvaughn:
I was thinking it might make sense to restructure things:

```
<root>
├───shells
│   ├───chrome
│   │   │   manifest.template.json
│   │   │   README.md
│   ├───firefox
│   │   │   manifest.template.json
│   │   │   README.md
│   └───webextension
│       │   (current shells/chrome content...)
```

Then use a build script (similar to the approach in #632) to package each of the extensions using their own manifests. This way `applicatinos.gecko.strict_min_version` doesn't leak into the Chrome extension and `minimum_chrome_version` doesn't leak into Firefox.

Or maybe it would be sufficient to just rename the `shells/chrome` directory to `shells/webextension` and add a separate Chrome and Firefox manifest template that the build script knows how to work with:
```
<root>
├── shells
│   ├── webextension
│   │   │   manifest.chrome.json
│   │   │   manifest.firefox.json
│   │   │   ...
```

bvaughn:
I've got the new folder structure and process I described above working locally.

#### Building Chrome
Chrome

#### Building Firefox
Firefox

#### Running the build script
building-devtools


The one limitation of the new build script is that you can't build both extensions at the same time. I could fix that though if we think it's important to support.

bvaughn:
I'm going to restructure the code slightly to enable parallel builds by using a temp dir for each build. This will also avoid possibly cluttering the workspace with copied manifests if the build script dies in the middle. Should be easy to do though.

clarkbw:
+1 on the use of `chaulk` 😄

shells/firefox/manifest.json
```diff
@@ -0,0 +1,55 @@
+{
+  "manifest_version": 2,
+  "name": "React Developer Tools",
+  "description": "Adds React debugging tools to the Chrome Developer Tools.",
+  "version": "2.3.1",
+
+  "applications": {
+    "gecko": {
+      "id": "extension@react.devtools",
+      "strict_min_version": "54.0"
```
clarkbw: I should check the versions here.  Things mostly work in 54 and definitely work in 55

shells/firefox/manifest.json
```diff
@@ -0,0 +1,55 @@
+{
+  "manifest_version": 2,
+  "name": "React Developer Tools",
+  "description": "Adds React debugging tools to the Chrome Developer Tools.",
```
clarkbw: s/Chrome/Firefox/ 😄

bvaughn: Whoops 😅

bvaughn:
Okay. The new build script feels pretty solid now. Based on my testing, it generates a useable extension for Chrome (58) and Firefox (54 and 55).

shells/firefox/build.js
```diff
@@ -15,11 +15,15 @@ const main = async () => {
...
-  console.log('\nThe Firefox extension has been built!');
+  console.log(chalk.green('\nThe Firefox extension has been built!'));
+  console.log(chalk.green('You can test it by following the steps below:'));
   console.log(chalk.gray('\n# Go to the unpacked directory:'));
   console.log(`cd ${unpackedPath}`);
   console.log(chalk.gray('\n# And launch Firefox with the extension enabled:'));
+  console.log('web-ext run');
+  console.log(chalk.gray('\n# You can also test against specific Firefox versions:'));
   console.log('web-ext run --firefox=nightly');
+  console.log('web-ext run --firefox=/Applications/Firefox54.app/Contents/MacOS/firefox-bin');
```
clarkbw: Perhaps in the Fx README if we link to the [Firefox releases](https://www.mozilla.org/en-US/firefox/channel/desktop/) and people download those then they can use `--firefox=firefox` or `--firefox=beta` or `--firefox=nightly` and it should just work.  I think `Firefox54.app` might be a custom location.  

Right now Firefox is 53, Beta is 54, Nightly is 55.  Of course next week those numbers will all bump up by one.

bvaughn: Good suggestion. Thanks!

bvaughn: I updated the README as well as the on-screen instructions.

screen shot 2017-06-08 at 8 35 01 am

bvaughn:
@gaearon Any thoughts on this one? 😄

bvaughn:
Firefox 54 is rolling out today. Going to test this extension in a fresh install once the final release is available and then I'd like to merge this branch. 😄

bvaughn:
I chatted a bit more with Bryan and it sounds like- after we merge this PR, if we decide to release an update for Firefox versions <=54, we could always release the old version, tag it with `"strict_max_version": "53"`, and the add-ons store would should the right version to the user based on their browser. (To be fair, we probably want to avoid doing this because it sounds like a hassle- but at least it's an option.)

shells/firefox/manifest.json
bvaughn: (Belated follow-up...) My own testing seemed to suggest that everything _except_ the theme worked in 54, and that degrades gracefully to the light theme.

bvaughn:
FYI I'm doing another round of testing this morning for this branch. Assuming I don't find any problems, I'm merging it.

PR #804, opened 2017-06-07
Bail out on fibers that returned false from shouldComponentUpdate

gaearon:
If https://github.com/facebook/react/pull/9887 gets in, this fixes https://github.com/facebook/react-devtools/issues/337.
Should also make DevTools faster on updates.

Will only work for Fiber.


backend/ReactTypeOfSideEffect.js
```diff
@@ -0,0 +1,26 @@
...
+  Bailout: 1, //            0b00000001
+  Placement: 2, //          0b00000010
+  Update: 4, //             0b00000100
+  PlacementAndUpdate: 6, // 0b00000110
+  Deletion: 8, //           0b00001000
+  ContentReset: 16, //      0b00010000
+  Callback: 32, //          0b00100000
+  Err: 64, //               0b01000000
+  Ref: 128, //              0b10000000
+};
```
bvaughn: This concerns me a bit. We might re-arranging or add side-effect types in the future and this could get out of sync easily (either in GitHub master, or in a specific deployed version of DevTools vs React).

Could we maybe pass these values in as a named parameter to `__REACT_DEVTOOLS_GLOBAL_HOOK__.inject` or something?

If we _have_ to hard-code, it seems better to limit it to only the `Bailout` value (which seems unlikely to change, assuming your inline comment is followed).

gaearon: I did this to match TypesOfWork which I previously inlined but if I recall correctly somebody asked me to un-inline it :-). I agree about just using the values directly.

gaearon: I'm not worried about Bailout number changing because I left a comment in React itself about it, and React repo logic also depends on it being 1 specifically.

I guess let's remove others, and add notes to React repo for constants we depend on externally (there's some other ones too outside DevTools). 

I'd avoid injection for this.

bvaughn: > I'm not worried about Bailout number changing because I left a comment in React itself about it, and React repo logic also depends on it being 1 specifically.

Agreed. 😄

gaearon:
We probably need to wait with landing this until FB www sync happens. Since we're now skip fibers that *don't* have `1` in effect tag, updates probably won't work with existing alphas.

gaearon:
Which also means we need an RN sync for this 😞

PR #805, opened 2017-06-07
[Electron] Fix problem of setDefaultThemeName override user setting

jhen0409:
Currently call `setDefaultThemeName(themeName)` will override the current user selected theme with new default, this is not expected. It happened since #751 and #779.

### DIFF
--- a/frontend/Themes/Store.js
+++ b/frontend/Themes/Store.js
@@ -31,7 +31,8 @@ class Store {
   }
 
   setDefaultTheme(defaultThemeName: ?string) {
-    this.setDefaultThemeName(defaultThemeName);
+    // Don't accept an invalid themeName as a default.
+    this.defaultThemeName = getSafeThemeName(defaultThemeName);
 
     // Don't restore an invalid themeName.
     // This guards against themes being removed or renamed.
@@ -75,11 +76,6 @@ class Store {
     this.theme = theme;
 
     setInLocalStorage(LOCAL_STORAGE_CUSTOM_THEME_KEY, serialize(theme));
-  }
-
-  setDefaultThemeName(defaultThemeName: ?string) {
-    // Don't accept an invalid themeName as a default.
-    this.defaultThemeName = getSafeThemeName(defaultThemeName);
   }
 }
 
--- a/frontend/Themes/Store.js
+++ b/frontend/Themes/Store.js
@@ -76,6 +76,12 @@ class Store {
   setDefaultThemeName(defaultThemeName: ?string) {
     // Don't accept an invalid themeName as a default.
     this.defaultThemeName = getSafeThemeName(defaultThemeName);
+    // Replace theme with new default if user hasn't select theme
+    const storedThemeKey = getFromLocalStorage(LOCAL_STORAGE_THEME_NAME_KEY);
+    if (storedThemeKey !== CUSTOM_THEME_NAME && !Themes.hasOwnProperty(storedThemeKey)) {
+      this.theme = Themes[this.defaultThemeName];
+      this.themeName = this.defaultThemeName;
+    }
   }
 }
 
--- a/frontend/Themes/Store.js
+++ b/frontend/Themes/Store.js
@@ -27,6 +27,10 @@ class Store {
   themes: { [key: string]: Theme };
 
   constructor(defaultThemeName: ?string) {
+    this.setDefaultTheme(defaultThemeName);
+  }
+
+  setDefaultTheme(defaultThemeName: ?string) {
     this.setDefaultThemeName(defaultThemeName);
 
     // Don't restore an invalid themeName.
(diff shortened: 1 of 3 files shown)


jhen0409:
Due to the [`store.changeTheme`](https://github.com/facebook/react-devtools/blob/master/frontend/Store.js#L343) will call [`themeStore.update`](https://github.com/facebook/react-devtools/blob/master/frontend/Themes/Store.js#L53), so it will forced update theme and save the default theme name to local storage.



Expected the method will not effect the user selected theme (https://github.com/facebook/react-devtools/pull/716#issuecomment-302584641), this PR fixed the problem:



cc @bvaughn

packages/react-devtools-core/src/standalone.js
```diff
@@ -191,10 +191,7 @@ var DevtoolsUI = {
       var {store} = panel.getChildContext();
       // Change default themeName if panel mounted
       store.setDefaultThemeName(themeName);
-      // Render devtools with the current theme.
-      // If user has selected a theme then ThemeStore will return it.
-      // If not the new default we set above will be used.
-      store.changeTheme(store.themeStore.defaultThemeName);
+      store.emit('theme');
```
glennreyes: Maybe we should explain here what's happening, like we had before with `changeTheme`.

frontend/Themes/Store.js
```diff
@@ -76,6 +76,12 @@ class Store {
   setDefaultThemeName(defaultThemeName: ?string) {
     // Don't accept an invalid themeName as a default.
     this.defaultThemeName = getSafeThemeName(defaultThemeName);
+    // Replace theme with new default if user hasn't select theme
+    const storedThemeKey = getFromLocalStorage(LOCAL_STORAGE_THEME_NAME_KEY);
+    if (storedThemeKey !== CUSTOM_THEME_NAME && !Themes.hasOwnProperty(storedThemeKey)) {
+      this.theme = Themes[this.defaultThemeName];
+      this.themeName = this.defaultThemeName;
+    }
```
bvaughn: I'm not sure if I like a method like `setDefaultThemeName` having this kind of side effect. Feels kind of unexpected.

I also wish we had less duplicate code between [these lines](https://github.com/jhen0409/react-devtools/blob/d186b6816be04e202e70b871b2551e1b8a552455/frontend/Themes/Store.js#L46-L50) and the newly-added ones.

Did you consider adding a new method (eg `setDefaultTheme`) that could be shared between the constructor and maybe also invoked for `src/standalone.js`? That way the intent is more clearly expressed and we have less duplication.

jhen0409: I just moved the constructor code to the new method. :)

packages/react-devtools-core/src/standalone.js
jhen0409: I added [this change](https://github.com/facebook/react-devtools/pull/805/files#diff-2f1a9061c5f923d07e5a5d2ddd5055ecR348), it may be clearly explained the use, without comments. :)

frontend/Themes/Store.js
bvaughn: I still have a small concern about the method name (now `setDefaultTheme`) doing a bit too much (eg loading custom theme). I'll just tweak this myself in a follow-up commit though so we don't have to keep going back and forth. I want to respect your time here. 😄 

Hopefully you're okay with this minor adjustment?
```diff
diff --git a/frontend/Themes/Store.js b/frontend/Themes/Store.js
index 27e6bcb..20754d4 100644
--- a/frontend/Themes/Store.js
+++ b/frontend/Themes/Store.js
@@ -27,6 +27,15 @@ class Store {
   themes: { [key: string]: Theme };
 
   constructor(defaultThemeName: ?string) {
+    this.themes = Themes;
+
+    // Load previous custom theme from localStorage.
+    // If there isn't one in local storage, start by cloning the default theme.
+    const customTheme = getFromLocalStorage(LOCAL_STORAGE_CUSTOM_THEME_KEY);
+    if (customTheme) {
+      this.customTheme = deserialize(customTheme);
+    }
+
     this.setDefaultTheme(defaultThemeName);
   }
 
@@ -41,18 +50,10 @@ class Store {
       this.defaultThemeName,
     );
 
-    // Load previous custom theme from localStorage.
-    // If there isn't one in local storage, start by cloning the default theme.
-    const customTheme = getFromLocalStorage(LOCAL_STORAGE_CUSTOM_THEME_KEY);
-    if (customTheme) {
-      this.customTheme = deserialize(customTheme);
-    }
-
     // The user's active theme is either their custom one,
     // Or one of the built-in sets (based on the default).
     this.theme = themeName === CUSTOM_THEME_NAME ? this.customTheme : Themes[themeName];
     this.themeName = themeName;
-    this.themes = Themes;
   }
 
   update(themeName: ?string) {
```

bvaughn:
Made a slight tweak in 3a0b82a as mentioned in my last comment. Thanks @jhen0409!

Issue #786, opened 2017-06-02
Follow-up keyboard nav fixes for Fiber


bvaughn:
There's one remaining keyboard nav issue (mentioned [here](https://github.com/facebook/react-devtools/pull/781#issuecomment-305832447) and [here](https://github.com/facebook/react-devtools/pull/781#issuecomment-305833600)): on facebook.com, click on any of the Like/Comment divs:
![screen shot 2017-06-02 at 9 06 29 am](https://cloud.githubusercontent.com/assets/29597/26734446/dbeb33ee-4772-11e7-9bfc-c2b59385b1b4.png)

Then use up/down arrows to step over the "clearfix" node:
![skip-null-2](https://cloud.githubusercontent.com/assets/29597/26734704/dd1098e4-4773-11e7-96db-c632d67b1b9e.gif)

The focus will disappear for one up/down press, showing "null" in the bottom breadcrumbs area, and then re-focus when you press up/down arrow again.

gaearon:
I’m still having this issue in UFI.

![](https://d2ppvlu71ri8gs.cloudfront.net/items/2y1B3P1M0E3219373O0D/Screen%20Recording%202017-06-09%20at%2004.57%20PM.gif?v=ab937276)

gaearon:
Fixed by #807.

PR #807, opened 2017-06-09
Skip nested empty array with keyboard

gaearon:
I think this fixes remaining issue in https://github.com/facebook/react-devtools/issues/786.
It was occurring because we didn't "jump over" a nested array which can be empty.

### DIFF
--- a/frontend/keyboardNav.js
+++ b/frontend/keyboardNav.js
@@ -178,12 +178,13 @@ function getNewSelection(dest: Dest, store: Store): ?ElementID {
   }
 
   // Children
+  var cid;
   if (dest === 'firstChild') {
     if (typeof children === 'string') {
       return getNewSelection('nextSibling', store);
     }
     for (var i = 0; i < children.length; i++) {
-      var cid = store.skipWrapper(children[i]);
+      cid = store.skipWrapper(children[i]);
       if (cid) {
         store.isBottomTagSelected = false;
         return cid;
@@ -194,7 +195,7 @@ function getNewSelection(dest: Dest, store: Store): ?ElementID {
     if (typeof children === 'string') {
       return getNewSelection('prevSibling', store);
     }
-    var cid = store.skipWrapper(children[children.length - 1], false, true);
+    cid = store.skipWrapper(children[children.length - 1], false, true);
     if (cid && !store.hasBottom(cid)) {
       store.isBottomTagSelected = false;
     }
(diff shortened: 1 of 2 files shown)


frontend/keyboardNav.js
```diff
@@ -182,8 +182,13 @@ function getNewSelection(dest: Dest, store: Store): ?ElementID {
     if (typeof children === 'string') {
       return getNewSelection('nextSibling', store);
     }
-    store.isBottomTagSelected = false;
-    return store.skipWrapper(children[0]);
+    for (var i = 0; i < children.length; i++) {
+      var cid = store.skipWrapper(children[i]);
```
gaearon: Find the first non-empty child rather than child #0.

frontend/keyboardNav.js
```diff
@@ -209,29 +214,32 @@ function getNewSelection(dest: Dest, store: Store): ?ElementID {
     pix = pchildren.indexOf(store._parents.get(id));
   }
   if (dest === 'prevSibling') {
-    if (pix === 0) {
```
gaearon: This looks like a messy diff, but I just moved this part *down*, and put the `else` branch into a `while` loop.

frontend/keyboardNav.js
```diff
@@ -209,29 +214,32 @@ function getNewSelection(dest: Dest, store: Store): ?ElementID {
...
-      const roots = store.searchRoots || store.roots;
-      if (roots.indexOf(store.getParent(id)) > -1) {
-        return getRootSelection(dest, store, id);
-      }
-      const childId = pchildren[pix];
+    while (pix > 0) {
+      const childId = pchildren[pix - 1];
```
gaearon: This code is copy paste from below, but I run it for every index one by one, looking for previous non-empty sibling before giving up.

frontend/keyboardNav.js
```diff
@@ -209,29 +214,32 @@ function getNewSelection(dest: Dest, store: Store): ?ElementID {
...
+      if (prevCid) {
+        if (store.hasBottom(prevCid)) {
+          store.isBottomTagSelected = true;
+        }
+        return prevCid;
       }
-      return getNewSelection('parent', store);
+      pix--;
+    }
+    const roots = store.searchRoots || store.roots;
```
gaearon: This code is copy paste from above and happens when we reached the zero.

PR #809, opened 2017-06-12
Add better comments to keyboard nav control

trueadm:
Adds some better inline comments to how the keyboard controls work on the dev tools view.


gaearon:
Could you also add comments to non-root paths?

trueadm:
@gaearon will pick up this up tomorrow as a follow up PR :)

Issue #798, opened 2017-06-06
Inconsistent left padding for RN style editor


gaearon:
The header is not aligned with other sections:

<img width="186" alt="screen shot 2017-06-06 at 4 01 23 pm" src="https://user-images.githubusercontent.com/810438/26835976-7a23bdae-4ad1-11e7-946f-b68991ee2065.png">

PR #816, opened 2017-06-16
Tweaked some right-panel paddings to better match Chrome

bvaughn:
Follow-up to #798 to match RN styles panel

screen shot 2017-06-16 at 10 15 58 am


gaearon:
Have you verified vertical mode is fine too? Just in case.

bvaughn:
Yeah, vertical mode is good! Thanks for checking 😄

We had inconsistent use of 0.25 and 0.5 before. This switched everything over to 0.25 b'c that's more like what Chrome uses and I made a judgement call on being space-conservative.

PR #824, opened 2017-06-29
Explicitly unset DevTools icon when URL changes

### LINKED ISSUE #823: Firefox: update icon when browsing away from a React site
This seems to be a separate problem from #650.

With the Firefox extension (2.3.3), a "React" tab is added to the developer tools and is shown for all pages, whether they use React or not. It’s a bit noisy if you’re not a React developer for 90% of your time. :)

Ideally, this tab would show up only when React is detected on the page. This seems to be how it works in Chrome.

If it requires a specific logic for Firefox, maybe it could be poached from what the [Vue.js DevTools](https://github.com/vuejs/vue-devtools) extension does.

bvaughn:
This prevents the Firefox extension from getting "stuck" showing a React icon when a user browses from a React site to a non-React site.

I don't _think_ this change impacts the behavior of the (current) Chrome extension but I'm not super familiar with the background script.

Resolves #823


bvaughn:
Whoops, just noticed that I need to rename `not-detected.html` to `disabled.html` to make the popup reset correctly too. I'm on it...

bvaughn:
@clarkbw I wonder if you might have some insight here.

Without this change, Firefox will permanently show an activated icon for React DevTools once a user has loaded a site with React (even if they browse away). By comparison, Chrome resets this state each time the user loads a new page.

With this change, Firefox functions correctly _but_ Chrome acts funny. (It seems like I keep getting `onUpdated` events after the initial page has loaded on eg facebook.com. Hash changes on sites like reactvirtualized.com also cause this to break for Chrome (but not for Firefox).

Any recommendations for a solid approach here?

clarkbw:
I'm not seeing why Chrome is behaving this way in general.  The [docs](https://developer.chrome.com/extensions/tabs#event-onUpdated) say it will reset when the tab is closed but it seems to be doing it when the URL changes as well.

bvaughn:
The odd thing to me, about Chrome, is the number of extra updates it’s dispatching relative to Firefox. Even after the page has finished loading, it dispatches a couple additional ones with status “loading”

digitarald:
Have you tried also checking `changeInfo.url`?

bvaughn:
Yeah, unfortunately Chrome seems to dispatches a lot of duplicate "loading" statuses with the same `changeInfo.url` so I wasn't able to determine a way to distinguish between them.

clarkbw:
I'll dig around into this a little more today and see if I can find anything else.

PR #830, opened 2017-07-06
Fix React detection on Firefox while navigating tab history

rpl:
While debugging the issue related to the react-devtools browserAction icon on Firefox (related to #823 and #824), I noticed that the browserAction is turned on when the tab has been navigated for the first time on a react app (e.g. one of the React todomvc demos), then it is turned off when navigating to a page that doesn't contain a react app, but once the tab navigates back to the React app again through the tab history, the browserAction is not turned on again as expected.

I compared the behavior on both Chrome and Firefox, and it seems that while in Chrome the content script is injected again in every page while navigating the tab history, in Firefox when the tab navigates the history, some of that documents can be still alive and the content scripts still active and not re-injected again (e.g. if the content script subscribes a listener on the pageshow/pagehide events and log something in the console, we can see the produces logs in the console when we navigate from/to that document in the history).

This PR contains a proposed workaround to make the content script able to replay the "React detection result" on pageshow, so that on Firefox the browserAction can be re-enabled as expected when the tab navigates back to a React app document which is still alive.

### DIFF
--- a/shells/webextension/src/GlobalHook.js
+++ b/shells/webextension/src/GlobalHook.js
@@ -18,6 +18,8 @@
 var installGlobalHook = require('../../../backend/installGlobalHook.js');
 var installRelayHook = require('../../../plugins/Relay/installRelayHook.js');
 
+var lastDetectionResult;
+
 // We want to detect when a renderer attaches, and notify the "background
 // page" (which is shared between tabs and can highlight the React icon).
 // Currently we are in "content script" context, so we can't listen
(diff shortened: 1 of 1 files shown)


rpl:
@bvaughn I did some debugging, and I noticed that a different behavior of the Firefox and Chrome content script injection when navigating the tab history is affecting the browserAction behavior on Firefox (I'm not completely sure if this is the same issue that we were discussing on twitter or a different one that exposes a similar behavior), let me know once you had the change to give it a try.

rpl:
@bvaughn sure, sounds great to me! \o/

I just tweaked the fix a little bit, only because I don't liked that in my previous version the "pageshow" listener was nested inside the "message" listener :-P

bvaughn:
Slick! Looks good to me 😁 Thanks again!

PR #835, opened 2017-07-10
Add NPM commands for quickly testing local extension

bvaughn:
Added new `test:chrome` and `test:firefox` Npm scripts to speed up local development. These scripts open a browser to a specific URL (the ReactJS homepage) with the local extension installed and enabled. Firefox also makes use of the default profile so as not to require users to re-log-into anything before testing. 🎸 

By default, build results push users toward these new NPM targets. The old, manual testing steps are still included though in case anyone encounters unexpected difficulty with the new scripts.

### Firefox build results
screen shot 2017-07-10 at 1 38 17 pm

### Chrome build results
screen shot 2017-07-10 at 2 00 02 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,8 @@
     "lint": "eslint .",
     "postinstall": "lerna bootstrap",
     "test": "node --harmony ./node_modules/.bin/jest",
+    "test:chrome": "node ./shells/chrome/test",
+    "test:firefox": "node ./shells/firefox/test",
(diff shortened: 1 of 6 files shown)


flarnie:
This looks pretty good - +1ing with the expectation that you've done manual testing. I'll manually test it later too. :)

bvaughn:
Thanks Flarnie! 😄

PR #837, opened 2017-07-10
Add "Copy element props" option to context menu

### LINKED ISSUE #833: Add support for copy props as JS
### Use case

I'm currently writing storybooks / jest snapshot tests for code that uses Relay, I'd like to be able to get the props that are passed to a component in prod into my source code so that I can reliably stub the networking. 

screen shot 2017-07-10 at 11 01 31

Functions can probably be stubbed functions that do nothing for example.

bvaughn:
As the gif below shows, elements with props (other than `children`) have a right-click option to copy props to clipboard. All props (excluding `children`) are copied as JSON. This may be helpful to people wanting to quickly copy "production" values out for testing/stubbing purposes.

Is this useful?

copy-element-props-

Resolves #833

### DIFF
--- a/utils/serializePropsForCopy.js
+++ b/utils/serializePropsForCopy.js
@@ -25,7 +25,7 @@ function stripFunctions(props: Object) {
   }
 }
 
-function serializeProps(props: Object): string {
+function serializePropsForCopy(props: Object): string {
   const cloned = Object.assign({}, props);
 
   // Don't try to copy 'children'; the data is probably not meaningful.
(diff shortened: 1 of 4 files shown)


alexkrolick:
This is really useful!

utils/serializePropsForCopy.js
```diff
@@ -0,0 +1,40 @@
...
+function serializePropsForCopy(props: Object): string {
+  const cloned = Object.assign({}, props);
+
+  // Don't try to copy 'children'; the data is probably not meaningful.
+  delete cloned.children;
+
+  // Convert functions to '[function]'
+  stripFunctions(cloned);
+
+  return JSON.stringify(cloned, null, 2);
```
jaredly: can you wrap this in a try/catch? there are some things that might throw

bvaughn: Definitely. That seems like a good safety step.

PR #839, opened 2017-07-11
Support case where functional component's displayName is a function

### LINKED ISSUE #803: Store version doesn't work on our site (fin.com) -- fixed version
For some reason, the store version of this extension breaks on fin.com, here's the fix that makes it work: https://github.com/facebook/react-devtools/compare/master...indigoviolet:master . 

Not sure how we're triggering this FB-specific hack -- if someone could give me a hint, we can try to fix our code. Would also be nice for the hack to not make it into the public version..

bvaughn:
Resolves  #803

### DIFF
--- a/backend/getDisplayName.js
+++ b/backend/getDisplayName.js
@@ -18,7 +18,23 @@ function getDisplayName(type: Function): string {
     return cachedDisplayNames.get(type);
   }
 
-  let displayName = type.displayName || type.name || 'Unknown';
+  let displayName = type.displayName || type.name;
+
+  // The displayName property might not be a string.
+  // If it's a custom function, attempt to use its return value.
+  // github.com/facebook/react-devtools/issues/803
+  if (typeof displayName === 'function') {
+    try {
+      displayName = displayName();
+    } catch (error) {}
+  }
+
+  // If we haven't got a display name (or we have an invalid one)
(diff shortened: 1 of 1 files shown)


bvaughn:
I'm going to move forward with this because I believe it will prevent the error reported in #803. Happy to iterate on it in a future release if there are suggestions for improvements. 😄

backend/getDisplayName.js
```diff
@@ -18,7 +18,23 @@ function getDisplayName(type: Function): string {
...
 
-  let displayName = type.displayName || type.name || 'Unknown';
+  let displayName = type.displayName || type.name;
+
+  // The displayName property might not be a string.
+  // If it's a custom function, attempt to use its return value.
+  // github.com/facebook/react-devtools/issues/803
+  if (typeof displayName === 'function') {
+    try {
+      displayName = displayName();
```
gaearon: Hmm, do we truly support this? I don't think we ever did. I think it's fine to just ignore it in this case—not sure we'd want to call it. For example we don't in the error messages.

bvaughn: Sure, that's fair. I can change this to fall back to `name || 'Unknown'` if `displayName` is a function rather than trying to be too clever about it.

bvaughn: Follow up: #842

PR #840, opened 2017-07-12
Change import to ES6 syntax

dePablobernat:


bvaughn:
Hi @dePablobernat!

The DevTools codebase uses `require` syntax for everything except Flow type imports. When we change this, we'll do a codemod for the entire project rather than for a single file. 😄

dePablobernat:
HI @bvaughn then why are you using es6 object spread destructuring.

Thanks!

Bernat

Issue #847, opened 2017-07-13
`yarn run test:firefox` fails with `Unhandled promise rejection`


flarnie:
Steps to reproduce;

1) Use `n` to run node v8.0.0.
2) run `yarn install` and `yarn build extension`
3) run `yarn run test:firefox`

Output:
```
Using globally installed version of Yarn
yarn run v0.28.1
$ node ./shells/firefox/test
(node:10470) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): ChildProcessError: Command failed: web-ext run --start-url=https://facebook.github.io/react/ --firefox-profile=/Users/flarnie/Library/Application\ Support/Firefox/Profiles/lyr8wuj6.default --keep-profile-changes --browser-console
/bin/sh: web-ext: command not found
 `web-ext run --start-url=https://facebook.github.io/react/ --firefox-profile=/Users/flarnie/Library/Application\ Support/Firefox/Profiles/lyr8wuj6.default --keep-profile-changes --browser-console` (exited with error code 127)
(node:10470) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

flarnie:
tagging @bvaughn since you have been helping with setting up these test scripts

flarnie:
`yarn run test:chrome` seems to hang for me.

I also tried switching to node v7.6.0 and got the same result. Going to do the manual testing steps for now and come back to this and investigate when I have time.

flarnie:
Now I'm thinking I might just need to install `web-ext` globally. The 'unhandled promise' looks unrelated.

flarnie:
That was it - ignore this.

bvaughn:
Sorry for the snag Flarnie. The [Firefox shell README](https://github.com/facebook/react-devtools/tree/master/shells/firefox#testing-in-firefox) mentions this as a pre-req but maybe the CLI output should also call it out? Thoughts?

clarkbw:
> The Firefox shell README mentions this as a pre-req but maybe the CLI output should also call it out? Thoughts?

What if we made `web-ext` one of the `devDependencies`?  This helps maintain a consistent version.  Lemme open a quick PR.

bvaughn:
I dig it

PR #848, opened 2017-07-13
Moves the web-ext utility to devDependencies

clarkbw:
Update the README and other docs to reflect the changes to how we use web-ext now.

Should improve the situation for issues like #847

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "deploy": "cd ./shells/theme-preview && ./build.sh && gh-pages -d .",
     "lint": "eslint .",
     "postinstall": "lerna bootstrap",
-    "start:firefox": "web-ext run -s shells/firefox/build/unpacked",
     "test": "node --harmony ./node_modules/.bin/jest",
     "test:chrome": "node ./shells/chrome/test",
     "test:firefox": "node ./shells/firefox/test",
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "deploy": "cd ./shells/theme-preview && ./build.sh && gh-pages -d .",
     "lint": "eslint .",
     "postinstall": "lerna bootstrap",
+    "start:firefox": "web-ext run -s shells/firefox/build/unpacked",
     "test": "node --harmony ./node_modules/.bin/jest",
     "test:chrome": "node ./shells/chrome/test",
     "test:firefox": "node ./shells/firefox/test",
@@ -69,6 +70,7 @@
   "devDependencies": {
     "chrome-launch": "^1.1.4",
     "firefox-profile": "^1.0.2",
-    "lerna": "2.0.0-beta.36"
+    "lerna": "2.0.0-beta.36",
+    "web-ext": "^1.10.1"
   }
 }
--- a/shells/chrome/README.md
+++ b/shells/chrome/README.md
@@ -2,11 +2,11 @@
 
 The source code for this extension has moved to `shells/webextension`.
 
-Modify the source code there and then rebuild this extension by running `node build.js` in this directory.
+Modify the source code there and then rebuild this extension by running `yarn run build:extension:chrome`.
 
 ## Testing in Chrome
 
 You can test a local build of the web extension like so:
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -44,6 +44,7 @@
     "deploy": "cd ./shells/theme-preview && ./build.sh && gh-pages -d .",
     "lint": "eslint .",
     "postinstall": "lerna bootstrap",
+    "start:firefox": "web-ext run -s shells/firefox/build/unpacked",
```
bvaughn: There's already a target below, `test:firefox` :)

shells/chrome/README.md
```diff
@@ -2,11 +2,11 @@
...
+Modify the source code there and then rebuild this extension by running `yarn run build:extension:chrome`.
 
 ## Testing in Chrome
 
 You can test a local build of the web extension like so:
 
- 1. Build the extension: `node build`
- 1. Follow the on-screen instructions.
\ No newline at end of file
+ 1. Build the extension: `yarn run build:extension:chrome`
```
bvaughn: nit: `node build` and `yarn run build:extension:chrome` are the same thing, except the latter needs to be run from the root directory. I don't mind changing the README but we should be clear I guess that the yarn command should not be run from the current dir.

shells/firefox/README.md
```diff
@@ -2,12 +2,11 @@
 
 The source code for this extension has moved to `shells/webextension`.
 
-Modify the source code there and then rebuild this extension by running `node build.js` in this directory.
+Modify the source code there and then rebuild this extension by running `yarn run build:extension:firefox`.
 
 ## Testing in Firefox
 
- 1. Install the [`web-ext` command line tool](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext#Installation).
```
bvaughn: Nice!

shells/firefox/build.js
```diff
@@ -25,13 +25,13 @@ const main = async () => {
   console.log(chalk.gray('\n# Go to the unpacked directory:'));
   console.log(`cd ${unpackedPath}`);
   console.log(chalk.gray('\n# And launch Firefox with the extension enabled:'));
-  console.log('web-ext run');
+  console.log('yarn run start:firefox');
```
bvaughn: Line 23 above already mentions the preferred testing route. These instructions just list out the manual steps for the curious.

clarkbw:
Thanks! I'll fix it up like you suggested

shells/chrome/README.md
clarkbw: Ok, I'm making this line use the `node build` method but I'll mention the other one in the sentence above.

bvaughn: Sure! 😄  I think, ideally, people can fully test this project from the root directory without having to bounce around between a bunch of sub-directories. That's why I've been adding more NPM scripts there.

shells/firefox/build.js
clarkbw: Oh right this is why I created the `start:firefox`.  I wasn't sure what to do here because now that we're installing web-ext this would need to change to be `../../node_modules/.bin/web-ext` or I can give instructions using the yarn script from the root directory; something like `(cd ../../ && yarn run start:firefox)`.  I think I'll give the location for `web-ext`.

bvaughn: Maybe it's okay for us to remove the manual `web-ext` instructions entirely in favor of just calling out the NPM script. What do you think? 😄

clarkbw: Yeah, that works.  Keeps it simpler. :)

I also discovered that I can use environment variables as well so I could keep the test.js file and just offer options like this `WEB_EXT_FIREFOX=nightly yarn run test:firefox`

bvaughn: I like it!

bvaughn:
This is excellent. Thanks @clarkbw !

clarkbw:
Thanks for all the feedback!

PR #853, opened 2017-07-14
Ignore 'null' internalInstance in React 15.*

### LINKED ISSUE #845: index undefined value
```
let serverList = []
for (let i in worldServers) {
    serverList.push(
        <option value={i}>{worldServers[i].name}{worldServers[i].delay}</option>
    )
}
<select>{serverList}</select>
```

if worldServers[i].delay is undefined, will throw an exception in backend.js:8688.

flarnie:
**what is the change?:**
In the <s>'walkRoots'</s> `getData` method used with React pre v16.0 we get 'null' as the
internalInstance in the following situation;
when a parent has a string child followed by '{null}'.

**Edit:** Instead of importing the whole `React` library in order to use
`React.Children.forEach`, we just hand-process the children into an
array, filter it, and then unwrap any single children.

**why make this change?:**
This was causing React devtools to throw in the backend.

<s>An alternative would be to figure out what kind of element we could construct using the 'null' internalInstance, but it seems to me that the correct thing to do here is ignore it.</s>

**Edit:** We found the place where we were pulling in the 'null' child from React and filtered it there.

**test plan:**
**Edit:** I beefed up my manual testing, rendering the following in a CRA:
```
class App extends Component {
  componentWillMount() {
  }

  componentDidMount() {
  }
  render() {
    const optionList = [];
    // possible children types that we can render in various combinations:
    const childValues = [
      {type: 'validChild', value: 'string child'},
      {type: 'afunction', value: () => 'returned string from fn'},
      {type: 'invalidNullChild', value: null},
    ];
    childValues.forEach((firstChild) => {
      childValues.forEach((secondChild) => {
        const childKey = firstChild.type + '-' + secondChild.type;
        optionList.push(
          <option key={childKey} value={childKey}>
            {firstChild.value}-{secondChild.value}
          </option>
        );

      });
    });
    optionList.push(<option key='empty' value='empty'></option>);
    return (
      <div className="App">
        <div className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
          <h2>Welcome to React</h2>
        </div>
        <p className="App-intro">
          To get started, edit <code>src/App.js</code> and save to reload.
        </p>
        <select>{optionList}</select>
      </div>
    );
  }
}
```
2. Run the CRA app and visit that url when testing the React devtools
   using `yarn test:chrome`
3. See that with this fix it actually renders the component tree, and
   without it the React devtools won't load.

**issue:**
fixes #845

### DIFF
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -59,7 +59,7 @@ function getData(internalInstance: Object): DataType {
     // Instead of pulling in the whole React library, we just copied over the
     // 'traverseAllChildrenImpl' method.
     // https://github.com/facebook/react/blob/240b84ed8e1db715d759afaae85033718a0b24e1/src/isomorphic/children/ReactChildren.js#L112-L158
-    let unfilteredChildren = internalInstance._currentElement.props.children;
+    const unfilteredChildren = internalInstance._currentElement.props.children;
     var filteredChildren = [];
     traverseAllChildrenImpl(
       unfilteredChildren,
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -10,7 +10,6 @@
  */
 'use strict';
 
-var React = require('React');
 import type {DataType} from './types';
 var copyWithSet = require('./copyWithSet');
 var getDisplayName = require('./getDisplayName');
@@ -56,14 +55,48 @@ function getData(internalInstance: Object): DataType {
     // prop is the unfiltered list of children.
     // This may include 'null' or even other invalid values, so we need to
     // filter it the same way that ReactDOM does.
-    const children = [];
-    React.Children.forEach(internalInstance._currentElement.props.children,
-      (child) => {
+    // We do a simplified filtering for valid children, instead of using
+    // `React.Children` so that we don't pull in the whole React library.
+    const REACT_ELEMENT_TYPE =
+      (
+        typeof Symbol === 'function'
+        && Symbol.for
+        && Symbol.for('react.element')
+      )
+      || 0xeac7;
+    let unfilteredChildren = internalInstance._currentElement.props.children;
+    const childrenType = typeof unfilteredChildren;
+    if (unfilteredChildren === null || childrenType === 'undefined') {
+      // only one child and it's invalid
+      unfilteredChildren = [];
+    } else if (childrenType === 'boolean'
+       || childrenType === 'string'
+       || childrenType === 'number'
+       || (childrenType === 'object' && unfilteredChildren.$$typeof === REACT_ELEMENT_TYPE)) {
+      // only one child, so let's wrap it in an array
+      unfilteredChildren = [unfilteredChildren];
+    }
+    var filteredChildren = [];
+    function gatherValidChildren(childrenArray) {
+      childrenArray.forEach((child, i) => {
         if (typeof child === 'string' || typeof child === 'number') {
-          children.push(child);
+          filteredChildren.push(child);
+        } else if (Array.isArray(child)) {
+          gatherValidChildren(child);
         }
-      },
-    );
+        // ignore 'null' or invalid children
+      });
+    }
+    gatherValidChildren(unfilteredChildren);
+    if (filteredChildren.length <= 1) {
+      // children must be an array of nodes or a string or undefined
+      // can't be an empty array
+      children = filteredChildren.length
+        ? String(filteredChildren[0])
+        : undefined;
+    } else {
+      children = filteredChildren;
+    }
   }
 
   if (!props && internalInstance._currentElement && internalInstance._currentElement.props) {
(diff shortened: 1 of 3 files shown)


gaearon:
It seems better to me to do this at the point we determine `children` rather than as we walk them. Since there may be more cases where we assume each of `children` exists.

gaearon:
Specifically we gather children [here](https://github.com/flarnie/react-devtools/blob/267f8f70620524944c3d7138c1c199ce6860fe53/backend/getData.js#L177-L181). So one of them is `null` in this case? This seems surprising to me because I would expect Stack to skip over null children and not add them to `_renderedChildren` in the first place.

gaearon:
I think [this](https://github.com/flarnie/react-devtools/blob/267f8f70620524944c3d7138c1c199ce6860fe53/backend/getData.js#L54-L57) is the right place to fix it.

> or (in the case of the <option>) a list of strings & numbers.

This is exactly the case we’re dealing with. Since we’re reading them from `props` rather from `_renderedChildren` the assumption that we’ve already filtered them is not true.

I think we should fix that branch to match [how ReactDOM handles them](https://github.com/facebook/react/blob/3b5e3b5a23069dd03feaa6979bccb64de4d76d09/src/renderers/dom/stack/client/wrappers/ReactDOMOption.js#L21). It uses `React.Children.forEach` to iterate over them, and only “takes” strings and numbers. So we should probably filter similarly.

flarnie:
Thanks @gaearon - I'm impressed with your knowledge of this codebase and will fix this up to follow your recommendation.

megalike

flarnie:
I had one doubt about my update - to iterate the `children` we normally use `React.Children`, and so I did that, but that means requiring `React` in the `backend` code. It looks like we don't do that so far in the `backend`, except in the injection code for specific versions of React. So I hope this doesn't add bloat or anything.

I can also just manually check if `...props.children` is an array and wrap it if it's not an array, if we want to avoid pulling in all of `React`.

gaearon:
Hmm that’s a good point. We probably don’t want to bring it in.

>I can also just manually check if ...props.children is an array and wrap it if it's not an array, if we want to avoid pulling in all of React.

This could work but probably wouldn’t for nested arrays. I’m not sure if nested arrays are even supported there, but if ReactDOM uses `React.Children` then it can probably handle those.

backend/getData.js
```diff
@@ -52,9 +52,51 @@ function getData(internalInstance: Object): DataType {
...
     // This is a native node without rendered children -- meaning the children
-    // prop is just a string or (in the case of the <option>) a list of
-    // strings & numbers.
-    children = internalInstance._currentElement.props.children;
+    // prop is the unfiltered list of children.
+    // This may include 'null' or even other invalid values, so we need to
+    // filter it the same way that ReactDOM does.
+    // We do a simplified filtering for valid children, instead of using
+    // `React.Children` so that we don't pull in the whole React library.
+    const REACT_ELEMENT_TYPE =
```
flarnie: The iteration logic is copied loosely from `traverseAllChildrenImpl` in `ReactChildren.js`:
https://github.com/facebook/react/blob/240b84ed8e1db715d759afaae85033718a0b24e1/src/isomorphic/children/ReactChildren.js#L112-L158

backend/getData.js
```diff
@@ -56,14 +55,48 @@ function getData(internalInstance: Object): DataType {
     // prop is the unfiltered list of children.
     // This may include 'null' or even other invalid values, so we need to
     // filter it the same way that ReactDOM does.
-    const children = [];
-    React.Children.forEach(internalInstance._currentElement.props.children,
-      (child) => {
+    // We do a simplified filtering for valid children, instead of using
+    // `React.Children` so that we don't pull in the whole React library.
+    const REACT_ELEMENT_TYPE =
```
gaearon: Let's hoist `REACT_ELEMENT_TYPE` outside the function.

gaearon:
Hmm. It is a bit hard to review because the code is complex. What do you think about inlining `traverseAllChildrenImpl` directly without changing its code? This might be a bit more generic than we need, but it gives us full confidence because this code has been tested for a long time.

flarnie:
I could try that - have some other things to work on before I come back to this though.

gaearon:
I could merge but it’s scary because we don’t have tests for this either.

flarnie:
Agreed - I would like to have a test for this logic. Going to work on this more as soon as I have time - thanks for your feedback and patience.

gaearon:
I'm fine with having no test if the logic is completely copy pasted. Because this is old code branch that will be removed when we drop support for Stack in DevTools (maybe in a couple of years) and it's completely localized.

flarnie:
From manual testing in Chrome:
screen shot 2017-07-27 at 9 18 36 am

gaearon:
This looks good to me. Let's get CI to pass and get it in. Thanks for following up!

PR #855, opened 2017-07-19
Remove deprecated proptypes

piuccio:
In react 15.5 React.Proptypes is deprecated and `prop-types` should be used instead

### DIFF
--- a/frontend/Breadcrumb.js
+++ b/frontend/Breadcrumb.js
@@ -16,6 +16,7 @@ import type {Theme} from './types';
 
 var {sansSerif} = require('./Themes/Fonts');
 var React = require('react');
+var PropTypes = require('prop-types');
 var decorate = require('./decorate');
 
 class Breadcrumb extends React.Component {
@@ -67,7 +68,7 @@ class Breadcrumb extends React.Component {
 }
 
 Breadcrumb.contextTypes = {
-  theme: React.PropTypes.object.isRequired,
+  theme: PropTypes.object.isRequired,
 };
 
 const containerStyle = (theme: Theme) => ({
--- a/frontend/ContextMenu.js
+++ b/frontend/ContextMenu.js
@@ -11,6 +11,7 @@
 'use strict';
 
 var React = require('react');
+var PropTypes = require('prop-types');
 var {sansSerif} = require('./Themes/Fonts');
 var HighlightHover = require('./HighlightHover');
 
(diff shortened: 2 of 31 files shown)


piuccio:
Not really sure how to test it other than `npm test`

bvaughn:
Couple of thoughts!

* DevTools currently uses [React version 15.4.2](https://github.com/facebook/react-devtools/blob/2bd9bdccfc5bfbdcab1e366b11de2b8e14382d80/package.json#L30) so this deprecation warning hasn't impacted the project yet. We could (and probably should) bump to version 15.6 though.
* DevTools has no dependency on the `prop-types` package yet. Until we add this, there's no `prop-types` lib to import.

Both of these are easily addressed by running the following:
```
yarn add react@^15.6.0 prop-types@^15.5.10
```

CI is failing on this branch, in part, because there are now 2 `PropTypes` variables defined in `SettingsPane.js` (lines 16 and 20).

Did you, by chance, use the [`React-PropTypes-to-prop-types` codemod](https://github.com/reactjs/react-codemod#react-proptypes-to-prop-types) from [reactjs/react-codemod](https://github.com/reactjs/react-codemod) to do this? I fixed a couple of small bugs with it just yesterday that may be related to the duplicate variable (eg reactjs/react-codemod/pull/160). Any interest in re-running with an updated version of this codemod?

piuccio:
Sorry didn't notice the CI fail. 
I've modified the package.json to include prop-types as a dependency but if you're using yarn I'll have to update the lock file as well. 

I don't think it's necessary to upgrade react, not in this PR at least, using external prop-types is compatible with both versions. 

I did all the changes by hand, last time I used the codemod it didn't work as expected, also all files have a mix of var / const, I'm not sure what the codemod will pick but I stuck with whatever was used in the file. 

I'm now at the airport and I'll be away for two weeks, not sure if I'll have time to address your comments. Feel free to take over if you want to push it through

piuccio:
@bvaughn I've pushed a new commit with changes to `SettingsPane` to remove duplicate `PropTypes` and updated `yarn.lock`.

There are a few changes on `yarn.lock`, I guess it's due to the `^` in package.json, anyway, let me know if this is fine

piuccio:
@bvaughn fixed that

piuccio:
@bvaughn gentle bump

kelset:
Any update on this?

BonnieMilian:
Is breaking my Build

bvaughn:
@BonnieMilian can you explain how a closed PR is breaking your build? 😕

BonnieMilian:
The yarn thing help me, but just in case review it if there is PropTypes issues.

PR #858, opened 2017-07-26
RN: Show computed properties for debugging flexbox layout

### LINKED ISSUE #515: [React Native][Feature Request] Display layout styles in a friendlier format for reading
### Problem

Currently, the layout styles are displayed exactly the way they were written. It'd be useful to do some calculations on them and display them in a friendlier format for reading.

For example, when trying to understand how a view got its size, the `flex` and `alignSelf` properties are very useful. However, in order to understand which axis these apply to, I have to look at the parent's `flexDirection` style. I also might want to look at the parent's `alignItems` property. This process is tedious and mechanical and the inspector can automate it.

### Solution

We might want to have a layout area in the right pane that shows this computed information for you so all the information you need is in one spot. If the parent's `flexDirection` is `row`, it might look like this:
  - `width: 100`
    - `flexBasis: 0`
    - `flexGrow: 1`
    - `flexShrink: 1 (from default)`
  - `height: 100`
    - `alignSelf: flex-start (from alignItems)`

Notice also that each value is annotated with where it came from (e.g. the value is the `default` value or the value came from `alignItems` of the parent). No annotation means the value was specified explicitly.

Another inspector feature that falls into this category is #330.

### Implementation

In order to implement this, the inspector code needs to be able to walk from the current node to its parent node and get the parent's style information. It doesn't look like nodes currently have pointers to their parents.

@gaearon I know you mentioned that React is being rewritten but I'd still be interested in some information about how to walk from a node to the parent node in the current React implementation for use in this feature.

rigdern:
Fixes #515.

The primary purpose of this change is to render a new section in the right pane useful for debugging flexbox layouts in React Native. It helps answer questions such as what values does the layout engine see for properties like `flexGrow`, `flexShrink`, and `alignSelf`? Which dimension are these properties influencing (e.g. width or height)? How did each property get its value? Was `flexGrow` set explicitly, is it getting its value from `flex`, or is it just using its default value?

As part of this change, I decided to create a new section in the right pane called "Layout". This section shows the new information I added for debugging flexbox. Additionally, I moved the box which shows the element's width, height, margin, and padding out of the "React Native Style Editor" section and into the "Layout" section. Initially, I tried putting the new flexbox debug UI into the "React Native Style Editor" section but it looked out of place. This is why I created the new "Layout" section.

Here's what it looks like when the inspected element doesn't have a style and it doesn't have a measure method:

image

Here's what the new flexbox debug information looks like:

image

Some things to note about the sceenshot:
  - `alignSelf` is under the width because it influences the width. Similarly, `flexBasis`, `flexGrow`, and `flexShrink` are under the height because they influence the height.
  - `alignSelf` is slightly faded out because its value is the default. Its value is neither explicitly set nor is it being set indirectly via another property (i.e. the parent's `alignItems` property).
  - `flexBasis`, `flexGrow` and `flexShrink` are fully opaque because their values are either explicitly set or being set indirectly via another property.
  - If you hover over a value, a tooltip appears indicating how the property got its value. The screenshot tells us that `flexGrow` inherited its value from the `flex` property.
  - The blue numbers next to the "Width" and "Height" headers indicate the measured value of that dimension.
  - The slightly faded out "width" and "height" properties represent the value the developer explicitly set for that dimension. In this case, they are slightly faded out and their values are both `undefined` indicating that the developer didn't explicitly set either dimension.

Adam Comella
Microsoft Corp.

### DIFF
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -222,7 +222,7 @@ class Panel extends React.Component {
 
       this._themeStore = new ThemeStore(this.state.themeName);
       this._store = new Store(this._bridge, this._themeStore);
-      
+
       var refresh = () => this.forceUpdate();
       this.plugins = [
         new RelayPlugin(this._store, this._bridge, refresh),
@@ -303,7 +303,7 @@ class Panel extends React.Component {
     var extraTabs = assign.apply(null, [{}].concat(this.plugins.map(p => p.tabs())));
     var extraPanes = [].concat(...this.plugins.map(p => p.panes()));
     if (this._store.capabilities.rnStyle) {
-      extraPanes.push(panelRNStyle(this._bridge, this._store.capabilities.rnStyleMeasure, theme));
+      extraPanes.push(panelRNStyle(this._store, this._bridge, this._store.capabilities.rnStyleMeasure, theme));
     }
     return (
       <Container
@@ -365,20 +365,15 @@ Panel.childContextTypes = {
   themes: React.PropTypes.object.isRequired,
 };
 
-var panelRNStyle = (bridge, supportsMeasure, theme) => (node, id) => {
-  var props = node.get('props');
-  if (!props || !props.style) {
-    return (
-      <div key="rnstyle" style={containerStyle(theme)}>
-        <strong>No style</strong>
-      </div>
-    );
-  }
+var panelRNStyle = (store, bridge, supportsMeasure, theme) => (node, id) => {
   return (
-    <div key="rnstyle" style={containerStyle(theme)}>
-      <strong>React Native Style Editor</strong>
-      <NativeStyler id={id} bridge={bridge} supportsMeasure={supportsMeasure} />
-    </div>
+    <NativeStyler
+      key="rnstyle"
+      id={id}
+      parentId={store.getParent(id)}
+      bridge={bridge}
+      supportsMeasure={supportsMeasure}
+    />
   );
 };
 
--- a/frontend/Panel.js
+++ b/frontend/Panel.js
@@ -377,12 +377,6 @@ var panelRNStyle = (store, bridge, supportsMeasure, theme) => (node, id) => {
   );
 };
 
-const containerStyle = (theme: Theme) => ({
-  borderTop: `1px solid ${theme.base01}`,
-  padding: '0.25rem',
-  marginBottom: '0.25rem',
-  flexShrink: 0,
-});
 const loadingStyle = (theme: Theme) => ({
   fontFamily: sansSerif.family,
   fontSize: sansSerif.sizes.large,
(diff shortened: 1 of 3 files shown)


gaearon:
Thanks for the PR! I’m sorry we didn’t respond. We’re trying to ship 16 and it’s been very busy few months.

I’m a bit worried about `resolve*` logic living here. Do we have to write this ourselves or can we use some library to do that for us? I’m not sure where this logic comes from (a spec?)

rigdern:
@gaearon Thanks for reviewing this. I think it's unlikely we'll find a JavaScript library which implements the `resolve*` functions. The `resolve*` functions are based on Yoga's behavior which isn't always consistent with how browser flexbox implementations behave.

I adapted some of the `resolve*` functions from Yoga's source code:
  - [`resolveFlexGrow`](https://github.com/facebook/yoga/blob/16052085d0c84e8e44c8fdb903fbb87f4fe4d636/yoga/Yoga.c#L530-L542)
  - [`resolveFlexShrink`](https://github.com/facebook/yoga/blob/16052085d0c84e8e44c8fdb903fbb87f4fe4d636/yoga/Yoga.c#L554-L567)
  - [`resolveFlexBasis`](https://github.com/facebook/yoga/blob/16052085d0c84e8e44c8fdb903fbb87f4fe4d636/yoga/Yoga.c#L569-L577)

`resolveAlignSelf` is based on the description of the relationship between `alignItems` and `alignSelf` from the Yoga documentation: https://facebook.github.io/yoga/docs/alignment/.

The implementations of `resolveWidth` and `resolveHeight` are just intended to indicate whether or not the user has explicitly set the `width` and `height` styles, respectively.

If you'd like, I can add comments to indicate where these implementations came from.

gaearon:
Is there any way we can use Yoga itself instead of inlining its logic?

rigdern:
@gaearon Here's the work that I think would be required to get this information directly from Yoga:
  - `Yoga.c` has to publicly expose `resolveFlexGrow`, `resolveFlexShrink`, and `resolveFlexBasis` (I'm not sure if Yoga has an equivalent of `resolveAlignSelf` and I think `resolveWidth` and `resolveHeight` are too simple to be worth plumbing through Yoga).
  - Each of the Yoga language projections (e.g. Java, C#) has to expose these methods and forward them to the C code.
  - The Yoga implementations have to be modified to expose the `source` of the data (e.g. for `flexGrow` is it explicitly set, is it inferred from `flex`, or is it the default value?). Currently the Yoga functions return the resolved number and that's it. We'd have to make sure returning this extra information wouldn't negatively impact Yoga's performance.
  - The inspector would have to be modified in a way so it can call the Yoga methods (similar to how the inspector has a way to get the styles from React Native today).
  - The JavaScript `resolve*` functions would become asynchronous so we'd have to modify the inspector code to take that into account.

@emilsjolander what are your thoughts on this?

emilsjolander:
I think this data would be useful for general debugging in many other tools so I would add it to Yoga. Add a YGDebugOutput struct to YGNode which contains all the resolved values and any other debug info you need. Then add apis to access these.

rigdern:
Additionally, Emil suggested populating the `YGDebugOutput` struct while the algorithm runs but only in debug mode. I will look into this.

bvaughn:

PR #876, opened 2017-08-30
Set the minimum Chrome version to 49

ngyikp:
On older versions of Chrome, the browser action icon doesn’t change properly and gives incorrect results. Even the React inspector itself might not work properly.

chrome minimum version

I tested multiple old Chrome versions and found that 49 is the minimum.

### DIFF
--- a/shells/chrome/manifest.json
+++ b/shells/chrome/manifest.json
@@ -4,7 +4,7 @@
   "description": "Adds React debugging tools to the Chrome Developer Tools.",
   "version": "2.5.0",
 
-  "minimum_chrome_version": "43",
+  "minimum_chrome_version": "49",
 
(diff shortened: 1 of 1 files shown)


gaearon:
Can we fix it up instead? Can you dig into why the error is happening?

bvaughn:
If the "fix" is easy- let's do it! If it's complicated, I'd vote for us to bump the minimum version.

The current `minimum_chrome_version` (43) was set 2 years ago (Sep 8, 2015). At that time, version 43 had been out for ~4 months. Given that at this time, version 49 has been released for 18 months, I think it's reasonable for us to bump the minimum.

gaearon:
Hmm. Sounds fine then 😛

PR #877, opened 2017-08-30
Make the browser action popup look better by enabling `browser_style`

ngyikp:
Enable `browser_style` which is available on Firefox 48+ (https://developer.mozilla.org/en-US/Add-ons/WebExtensions/manifest.json/browser_action#Syntax)
Since the extension already requires Firefox 54, compatibility risk is not a big issue.

Other changes:
- Set consistent body and text spacing
- Set the width of `<hr>` to 100% for Firefox compatibility
- Replace unnecessary link opener JavaScript with `target=“_blank”`

Here are the screenshots of the new popup on Chrome/Firefox and Windows/Mac/Ubuntu:

win7 chrome
win7 firefox
---
mac chrome
mac firefox
---
ubuntu chrome
ubuntu firefox

### DIFF
--- a/shells/firefox/manifest.json
+++ b/shells/firefox/manifest.json
@@ -26,7 +26,8 @@
       "128": "icons/128-disabled.png"
     },
 
-    "default_popup": "popups/disabled.html"
+    "default_popup": "popups/disabled.html",
+    "browser_style": true
   },
 
   "devtools_page": "main.html",
--- a/shells/webextension/popups/deadcode.html
+++ b/shells/webextension/popups/deadcode.html
@@ -1,9 +1,19 @@
 <script src="shared.js"></script>
 <style>
-  html, body {
+  body {
     font-size: 14px;
+    line-height: 1.5;
+    margin: 8px;
     min-width: 460px;
     min-height: 133px;
+  }
+
+  p {
+    margin: 0;
+  }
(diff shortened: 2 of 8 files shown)


gaearon:
Nice! Do you mind fixing up the conflicts? I'm sorry I didn't notice your PR before merging mine.

ngyikp:
Just fixed the conflicts, please take another look.

gaearon:
>Replace unnecessary link opener JavaScript with target=“_blank”

If I recall there was some specific reason why it was done that way. Do you know why it became unnecessary? I can't remember the exact reason :-(

gaearon:
It would be nice to have a little more separation between paragraphs here.

screen shot 2017-09-14 at 02 58 20

gaearon:
I also really want to figure out why we had that link workaround. Apparently it didn't work before: https://stackoverflow.com/questions/8915845/chrome-extension-open-a-link-from-popup-html-in-a-new-tab.

Is it possible that this only works with newest versions of Chrome? Can you check Chrome 39?

gaearon:
I pushed to undo non-essential changes for now until you can look at:

* Verifying links work in older Chrome
* Fixing the spacing between large paragraphs

I’ll take the rest now. Thank you!

PR #888, opened 2017-09-13
Add bad dead code elimination detection for React 16

gaearon:
Counterpart to https://github.com/facebook/react/pull/10702.
We need this because the hazard of shipping “two Reacts” is a new one due to flat bundles.

Our existing detection strategy doesn’t work here because it checks compiled code. But if you had bad DCE then you have *two* compiled versions, one of them dormant and never running but still present in your bundle.

So https://github.com/facebook/react/pull/10702 sidesteps this problem by calling DevTools in the CommonJS entry point.

screen shot 2017-09-13 at 23 30 04


shells/webextension/popups/development.html
```diff
@@ -2,7 +2,7 @@
 <style>
   html, body {
     font-size: 14px;
-    min-width: 430px;
+    min-width: 460px;
```
gaearon: This is unrelated but Chrome probably changed default fonts in popups because they started wrapping onto the next line a few versions ago. I made all popups 30px wider to make them look nice again.

gaearon:
Added the code to throw an error in `setTimeout`, just like we originally did in https://github.com/facebook/react/pull/10446.

backend/installGlobalHook.js
```diff
@@ -31,29 +30,17 @@ function installGlobalHook(window: Object) {
...
-        // By now we know that it's envified--but what if it's not minified?
-        // This can be bad too, as it means DEV code is still there.
-
-        // FIXME: this is fragile!
-        // We should replace this check with check on a specially passed
-        // function. This also doesn't detect lack of dead code elimination
-        // (although this is not a problem since flat bundles).
-        if (findFiberCode.indexOf('getClosestInstanceFromNode') !== -1) {
-          return 'unminified';
-        }
```
bvaughn: This change means DevTools won't quite support 16 alpha/beta/RC? I guess that's okay though.

backend/installGlobalHook.js
```diff
@@ -31,29 +30,17 @@ function installGlobalHook(window: Object) {
...
+        // React 16 uses flat bundles. If we report the bundle as production
+        // version, it means we also minified and envified it ourselves.
         return 'production';
+        // Note: There is still a risk that the CommonJS entry point has not
+        // been envified or uglified. In this case the user would have *both*
+        // development and production bundle, but only the prod one would run.
+        // This would be really bad. We have a separate check for this because
+        // it happens *outside* of the renderer injection. See `checkDCE` below.
       }
+      var toString = Function.prototype.toString;
```
bvaughn: Curious why we do this (vs `Function.prototype.toString.call`)?

backend/installGlobalHook.js
```diff
@@ -141,14 +128,45 @@ function installGlobalHook(window: Object) {
...
+    checkDCE: function(fn) {
+      // This runs for production versions of React.
+      // Needs to be super safe.
+      try {
+        var toString = Function.prototype.toString;
+        var code = toString.call(fn);
+        // This is a string embedded in the passed function under DEV-only
+        // condition. However the function executes only in PROD. Therefore,
+        // if we see it, dead code elimination did not work.
+        if (code.indexOf('^_^') > -1) {
```
bvaughn: Heh, should we put something slightly more descriptive? 😁
I guess this pretty unique and tiny which is nice. It's just so arbitrary.

**Edit** Looks like the corresponding change has already been merged to react so nevermind. :)

shells/webextension/popups/development.html
bvaughn: Thanks for the explanation. 👍

shells/webextension/popups/disabled.html
```diff
@@ -2,7 +2,7 @@
 <style>
   html, body {
     font-size: 14px;
-    min-width: 380px;
+    min-width: 410px;
```
bvaughn: I assume this is intentionally a different size from the others (given it was already that way) but just pointing it out.

backend/installGlobalHook.js
gaearon: It won't be as good at detecting unminified alphas. Which seems okay. It won't report false positives though.

backend/installGlobalHook.js
gaearon: No reason. It used to be used in more places but now it's not.

backend/installGlobalHook.js
gaearon: I figured I didn't want it to be long, but also didn't want it to be something that could accidentally appear in that function. Since then it would cause false positive.

shells/webextension/popups/disabled.html
gaearon: Yes, it's a different message that's less wide.

PR #892, opened 2017-09-19
Fix the DevTools warning URL

gaearon:
fburl.com is FB-only.
We probably meant fb.me but I'm not at the office right now and can't create it.

Filing this so we don't forget to release an update to DevTools with this before 16.
Ideally should change it to fb.me alias.


bvaughn:
Closing in favor of #894

Issue #908, opened 2017-10-06
Can't build on Windows


raulpineda:
When running `npm run build:extension` on Windows an error is raised stating

```
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:28:11)
    at relative (path.js:559:5)
    at relativePath (C:\tools\react-devtools\shells\webextension\build.js:24:71)
    at main (C:\tools\react-devtools\shells\webextension\build.js:98:31)
    at <anonymous>
```

It is caused by `process.env.PWD` is `undefined`  in this line of shells/webextension/build.js `const relativePath = path => relative(process.env.PWD, path);`

Hunch: It could be fixed by using `__dirname` instead, like shells/chrome/build.js does.

ngyikp:
Even with that PR fix, building on Windows will still fail:

```
PS C:\Users\User\Desktop\react-devtools-2.5.2> npm run build:extension:chrome

> @ build:extension:chrome C:\Users\User\Desktop\react-devtools-2.5.2
> node ./shells/chrome/build

✓ Preparing build
{ ChildProcessError: Command failed: NODE_ENV=production C:\Users\User\Desktop\react-devtools-2.5.2\node_modules\.bin\webpack --config webpack.config.js --output-path C:\Users\User\Desktop\react-devtools-2.5.2\shells\webextension\build\chrome\bin
'NODE_ENV' is not recognized as an internal or external command,
operable program or batch file.
 `NODE_ENV=production C:\Users\User\Desktop\react-devtools-2.5.2\node_modules\.bin\webpack --config webpack.config.js --output-path C:\Users\User\Desktop\react-devtools-2.5.2\shells\webextension\build\chrome\bin` (exited with error code 1)
    at callback (C:\Users\User\Desktop\react-devtools-2.5.2\node_modules\child-process-promise\lib\index.js:33:27)
    at ChildProcess.exithandler (child_process.js:278:5)
    at emitTwo (events.js:125:13)
    at ChildProcess.emit (events.js:213:7)
    at maybeClose (internal/child_process.js:927:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:211:5)
  name: 'ChildProcessError',
  code: 1,
  childProcess:
   { ChildProcess: { [Function: ChildProcess] super_: [Object] },
     fork: [Function],
     _forkChild: [Function],
     exec: [Function],
     execFile: [Function],
     spawn: [Function],
     spawnSync: [Function: spawnSync],
[...]

PR #909, opened 2017-10-06
use __dirname instead of process.env.PWD for build relative path

raulpineda:
Fixes https://github.com/facebook/react-devtools/issues/908

### DIFF
--- a/shells/webextension/build.js
+++ b/shells/webextension/build.js
@@ -21,7 +21,7 @@ const STATIC_FILES = [
   'panel.html',
 ];
 
-const relativePath = path => relative(process.env.PWD, path);
+const relativePath = path => relative(__dirname, path);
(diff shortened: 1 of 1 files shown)


shells/webextension/build.js
```diff
@@ -21,7 +21,7 @@ const STATIC_FILES = [
   'panel.html',
 ];
 
-const relativePath = path => relative(process.env.PWD, path);
+const relativePath = path => relative(__dirname, path);
```
gaearon: This is relative from `/shells/webextension/build`. Seems confusing.

Can we just use `process.cwd()`?

raulpineda: I had forgotten about this PR, but yes, that makes much more sense. Updated.

PR #911, opened 2017-10-11
Add a way to forceUpdate stateless components

### LINKED ISSUE #868: Read-only props for functional components
screen shot 2017-08-10 at 4 14 31 pm

It looks like props for functional components are being marked at "read-only". From digging around, 
it seems like this is because functional component instances don't have a `forceUpdate()` method. I definitely understand why this is, but I've seen this be confusing to developers. I can see two obvious options to help with this:

1. Update the label in the devtools to something more descriptive than "read-only". I'd suggest "functional component (read-only)".
2. Change the implementation to be able to edit props of all components (forceUpdate for functional component instances).

Option 1 is easier (and I'd be happy to provide a PR) but I think option 2 is the best long-term solution. From the POV of the React Core team trying to encourage people to use functional components, there probably shouldn't be any significant downsides to using them (refs aside). This would be more involved but it would lead to the devtools "just-working".

Happy to help out any way I can here.

zgotsch:
Previously, stateless components could not have their props or context updated using React devtools. This seems to stem from the check for a publicly available `forceUpdate` method on the instance, which doesn't exist for functional components. Functional components seem like they can still be `forceUpdate`d using `enqueueForceUpdate` on the instance updater. Not sure if this is the way stateless components should be updated through devtools, but it works for props and context in 15.4.2.

I expect Fiber uses a different codepath (in `backend/getDataFiber.js`), so it may not suffer from this issue, and if it does, this will not fix the problem.

Fixes #868.

### DIFF
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -127,12 +127,17 @@ function getData(internalInstance: Object): DataType {
 
   if (internalInstance._instance) {
     var inst = internalInstance._instance;
+
+    // A forceUpdate for stateless (functional) components.
+    var forceUpdate = inst.forceUpdate || function(cb) {
+      inst.updater.enqueueForceUpdate(this, cb, 'forceUpdate');
+    };
     updater = {
       setState: inst.setState && inst.setState.bind(inst),
-      forceUpdate: inst.forceUpdate && inst.forceUpdate.bind(inst),
-      setInProps: inst.forceUpdate && setInProps.bind(null, internalInstance),
+      forceUpdate: forceUpdate && forceUpdate.bind(inst),
+      setInProps: forceUpdate && setInProps.bind(null, internalInstance, forceUpdate),
       setInState: inst.forceUpdate && setInState.bind(null, inst),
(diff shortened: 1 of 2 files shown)


backend/getData.js
```diff
@@ -127,12 +127,17 @@ function getData(internalInstance: Object): DataType {
 
   if (internalInstance._instance) {
     var inst = internalInstance._instance;
+
+    // A forceUpdate for stateless (functional) components.
+    var forceUpdate = inst.forceUpdate || function(cb) {
+      inst.updater.enqueueForceUpdate(this, cb, 'forceUpdate');
```
jaredly: very cool. could you make it `inst.forceUpdate || (inst.updater.enqueueForceUpdate && function(cb) {....`?

chrislloyd:
Thank you! This is awesome!

jmogielnicki:
Awesome, thank you!  This is going to make development more seamless for us.  Much appreciated.

PR #921, opened 2017-10-13
Update react-devtools-core to ws@3.2.0

mostafaeweda:
Address the breaking changes from https://github.com/websockets/ws/releases/tag/3.0.0
Keeping the old default `perMessageDeflate: true` -- not sure if you need it & willing to take its perf cost.

### DIFF
--- a/packages/react-devtools-core/package.json
+++ b/packages/react-devtools-core/package.json
@@ -25,7 +25,7 @@
   "license": "BSD-3-Clause",
   "dependencies": {
     "shell-quote": "^1.6.1",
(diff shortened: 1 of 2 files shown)


gaearon:
Can you explain what would change if we removed `perMessageDeflate`?

cuibonobo:
This option adds `gzip` compression to every message from the server. This can save on bandwidth between the client and the server with the trade-off that the server will use more CPU/memory per message. I'm not sure why it was enabled by default at first, since it may not make sense for most projects (especially if the message size is small).

bvaughn:

PR #937, opened 2017-11-07
Extend React DevTools Core APIs

rsnara:
React DevTools always opens up its own WebSocket connection to the phone. At Facebook, we needed it to use another tool's WebSocket connection. Therefore, I've made the public APIs of React DevTools Core accept WebSocket objects. These objects don't have to be genuine `window.WebSocket` instances, which means that can be backed by a number of different communication channels.


packages/react-devtools-core/src/backend.js
```diff
@@ -59,7 +61,7 @@ function connectToDevTools(options: ?ConnectOptions) {
   var messageListeners = [];
   var closeListeners = [];
   var uri = 'ws://' + host + ':' + port;
-  var ws = new window.WebSocket(uri);
+  var ws = websocket ? websocket : new window.WebSocket(uri);
   ws.onclose = handleClose;
   ws.onerror = handleClose;
   ws.onmessage = handleMessage;
```
gaearon: It feels odd that we slap these properties onto the passed object. Would it be better if this function took `send` and *returned* these handlers? Then the calling code could choose what to do with them (including putting them on a websocket or some completely different interface).

PR #939, opened 2017-11-13
Always use UTF-8 charset when injecting backend

ngyikp:
Test example:
1. Go to https://untitled-52pbe26ac93b.runkit.sh/bad (source code: https://runkit.com/ngyikp/5a09b954735c280012e78536)
2. Inspect the React element

Expected result:
screen shot 2017-11-14 at 12 00 11 am

Actual result:
screen shot 2017-11-14 at 12 00 24 am

Why:
Webpages that use a different charset from UTF-8 may mangle the display of certain inspector elements on the page (e.g. the ‘×’ in ‘2px × 2px’)
Real life example would be https://www.deviantart.com

Solution is to force UTF-8 charset when injecting the script


gaearon:
Thanks!

PR #968, opened 2018-01-18
Prevent error if global hook couldn't run

jammaloo:
On pages with a CSP that prevents loading of scripts at self (e.g. github), the global
hook will fail. This is because we create a script tag and then inject
it into the page itself.

When this happens window.__REACT_DEVTOOLS_GLOBAL_HOOK__ will be
undefined, and our checks for react will throw an exception.


gaearon:
Thanks!

PR #977, opened 2018-01-31
Upgrade to Flow 66

wbinnssmith:
Hi friends! Hope this is a welcome change.

Quite a bit has changed since Flow 23, so I'm going to take this in stages.

I can continue adding commits to this PR, or I can land additional PRs. At the very least each commit should be isolated and fairly reviewable on its own. Whatever works for reviewing this big change in stages :P

To make this reviewable in chunks without breaking the build, I automated inserting $FlowFixMe using flow's `./flow/tool add-comments` to suppress new issues. I'll replace each with proper annotations and fixes over the next few commits.

Let me know if this works for you and if I can accomodate otherwise.

### DIFF
--- a/backend/getDisplayName.js
+++ b/backend/getDisplayName.js
@@ -15,6 +15,7 @@ const cachedDisplayNames = new WeakMap();
 
 function getDisplayName(type: Function): string {
   if (cachedDisplayNames.has(type)) {
+    // $FlowFixMe From the upgrade to Flow 64
     return cachedDisplayNames.get(type);
   }
 
--- a/backend/getDisplayName.js
+++ b/backend/getDisplayName.js
@@ -11,12 +11,12 @@
 'use strict';
 
 const FB_MODULE_RE = /^(.*) \[from (.*)\]$/;
-const cachedDisplayNames = new WeakMap();
+const cachedDisplayNames: WeakMap<Function, string> = new WeakMap();
 
 function getDisplayName(type: Function): string {
-  if (cachedDisplayNames.has(type)) {
-    // $FlowFixMe From the upgrade to Flow 64
-    return cachedDisplayNames.get(type);
+  const nameFromCache = cachedDisplayNames.get(type);
+  if (nameFromCache != null) {
+    return nameFromCache;
   }
 
   let displayName;
--- a/lerna.json
+++ b/lerna.json
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "3.1.0"
+  "version": "3.0.0"
 }
--- a/plugins/TraceUpdates/TraceUpdatesAbstractNodePresenter.js
+++ b/plugins/TraceUpdates/TraceUpdatesAbstractNodePresenter.js
@@ -15,6 +15,8 @@ const requestAnimationFrame = require('fbjs/lib/requestAnimationFrame');
 
 import type {
   Measurement,
+  MetaData as MetaDataType,
+  Presenter,
 } from './TraceUpdatesTypes';
 
 // How long the measurement should be presented for.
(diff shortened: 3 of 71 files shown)


wbinnssmith:
cc @flarnie @gaearon @bvaughn 👋

gaearon:
This sounds amazing! I'm subscribed.

backend/getDisplayName.js
```diff
@@ -11,12 +11,12 @@
...
 const FB_MODULE_RE = /^(.*) \[from (.*)\]$/;
-const cachedDisplayNames = new WeakMap();
+const cachedDisplayNames: WeakMap<Function, string> = new WeakMap();
 
 function getDisplayName(type: Function): string {
-  if (cachedDisplayNames.has(type)) {
-    // $FlowFixMe From the upgrade to Flow 64
-    return cachedDisplayNames.get(type);
+  const nameFromCache = cachedDisplayNames.get(type);
+  if (nameFromCache != null) {
```
gaearon: `typeof !== 'string'` check might be faster

wbinnssmith: hmm, looks like it might be a wash ¯\_(ツ)_/¯

```js
const Benchmark = require('benchmark');

function getCache() {
  const cachedDisplayNames = new WeakMap();
  cachedDisplayNames.set(function exists() {}, 'hello');
  return cachedDisplayNames;
}

function getSuite() {
  return new Benchmark.Suite()
    .on('cycle', function(event) {
      console.log(String(event.target));
    })
    .on('complete', function() {
      console.log('Fastest is ' + this.filter('fastest').map('name'));
    });
}


// add tests
getSuite()
  .add('typeof with cached name', function() {
    typeof getCache().get('exists') === 'string';
  })
  .add('!= null with cached name', function() {
    getCache().get('exists') != null;
  })
  .run();

console.log('------------')

getSuite()
  .add('typeof with cache miss', function() {
    getCache().get('dne') === 'string';
  })
  .add('!= null with cache miss', function() {
    getCache().get('dne') != null;
  })
  .run();
```

results in

```
typeof with cached name x 4,954,129 ops/sec ±1.78% (85 runs sampled)
!= null with cached name x 5,012,633 ops/sec ±1.68% (90 runs sampled)
Fastest is != null with cached name,typeof with cached name
------------
typeof with cache miss x 5,012,751 ops/sec ±1.60% (89 runs sampled)
!= null with cache miss x 4,928,582 ops/sec ±2.10% (88 runs sampled)
Fastest is typeof with cache miss,!= null with cache miss
```

wbinnssmith: Tested again without the overhead of creating the weakmap (just by comparing against a string value and undefined), and again they're within a percentage point.

I'm always happy to change this, especially if it's more in line with the rest of the code. This was habit on my part.

wbinnssmith:
Okay, just pushed up the remainder of fixes. I erred on the side of not making assumptions and left a few $FlowFixMes behind. Let me know what you think.

Test plan:
* `git rebase -i --exec "yarn typecheck" HEAD~11`
* `yarn build:extension:chrome` && `yarn test:chrome`
* `yarn build:standalone` && `yarn test:standalone`
* Verify the chrome extension detects React on the react homepage and displays its element tree.

wbinnssmith:
Updated to reflect Flow 66

lerna.json
```diff
@@ -3,5 +3,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "3.0.0"
+  "version": "3.1.0"
```
gaearon: Why do we need this?

plugins/TraceUpdates/TraceUpdatesAbstractNodePresenter.js
```diff
@@ -85,15 +90,15 @@ class TraceUpdatesAbstractNodePresenter {
 
     if (this._clearTimer) {
       clearTimeout(this._clearTimer);
-      this._clearTimer = 0;
+      this._clearTimer = undefined;
```
gaearon: Why sometimes null and sometimes undefined?

plugins/TraceUpdates/TraceUpdatesAbstractNodePresenter.js
```diff
@@ -131,7 +136,7 @@ class TraceUpdatesAbstractNodePresenter {
 
     this.drawImpl(this._pool);
 
-    if (this._pool.size > 0) {
+    if (this._pool.size > 0 && this._clearTimer != null) {
```
gaearon: I believe this introduced a bug. The `setTimeout` call inside should've executed regardless of whether `this._clearTimer` exists.

gaearon: fixed in https://github.com/facebook/react-devtools/pull/1005

PR #979, opened 2018-01-31
Devtools display AsyncMode, StrictMode, and Context components

bvaughn:
Add some display for new React components. We can do better later; this is just a first treatment.

# Browser Example
## JSX
```js
const {
  unstable_AsyncMode: AsyncMode,
  StrictMode
} = React;

const NumericContext = React.createContext(0);

<Application>
  <NumericContext.Provider value={123}>
    <h1>Example</h1>

    <AsyncMode>
      <div>Async subtree</div>
    </AsyncMode>

    <StrictMode>
      <div>Strict subtree</div>
    </StrictMode>
      
    <NumericContext.Consumer>
      {value => (
        <div>
          NumericContext value:{value}
        </div>
      )}
    </NumericContext.Consumer>
  </NumericContext.Provider>
</Application>
```

## DevTools
screen shot 2018-01-31 at 1 47 10 pm

# ReactNative Example

## JSX
```js
var React = require('react');
var ReactNative = require('react-native');
var {
  Text,
  View,
} = ReactNative;

const {unstable_AsyncMode: AsyncMode, StrictMode} = React;

const NumericContext = React.createContext(0);

class Playground extends React.Component<{}> {
  render() {
    return (
      <NumericContext.Provider value={123}>
        <View>
          <Text>Example</Text>
        </View>

        <AsyncMode>
          <View>
            <Text>Async subtree</Text>
          </View>
        </AsyncMode>

        <StrictMode>
          <View>
            <Text>Strict subtree</Text>
          </View>
        </StrictMode>

        <NumericContext.Consumer>
          {value => (
            <View>
              <Text>NumericContext value:{value}</Text>
            </View>
          )}
        </NumericContext.Consumer>
      </NumericContext.Provider>
    );
  }
}
```

## DevTools
screen shot 2018-02-01 at 9 48 40 pm


gaearon:
We should at least check what happens if `Symbol` isn't there (does RN have it?)

Ideally yes, we should support both Symbol values and their numeric fallbacks and treat them the same way.

bvaughn:
I don't think it would be too hard to support both. It just didn't seem likely for someone to be using DevTools _with a browser_ that doesn't support symbol.

bvaughn:
> does RN have it?

If RN is connected to an environment capable of running DevTools (eg Chrome) then `Symbol` support would be polyfilled by that environment.

I think everything _except_ IE supports it, which is why I thought we could ignore it in the devtools.

gaearon:
>If RN is connected to an environment capable of running DevTools (eg Chrome) then Symbol support would be polyfilled by that environment.

I'm not quite sure what you mean (RN never worked with DevTools in Chrome); I was referring to the standalone DevTools that work via Electron.

bvaughn:
> I'm not quite sure what you mean (RN never worked with DevTools in Chrome); I was referring to the standalone DevTools that work via Electron.

Sorry, yes, that makes sense. I guess I combined debugger+devtools in my head somehow, because I never use one without the other.

I will have to try this with RN+standalone and report back.

bvaughn:
I've forgotten how to test this locally. I thought all that was required was to [run a couple of watch commands and then `npm start`](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools#developing) but I seem to be getting a stale backend from core.

I've forgotten where to copy the `react-devtools-core` package to to have it actually picked up by the electron app.

bvaughn:
I tested ReactNative in open source, iOS and Android, and verified that the new component types render okay.

I'm not sure how to actually test without `Symbol` support.

bvaughn:
FWIW, I manually turned `Symbol` use off in my browser and verified that DevTools still pick up on the right types for the new components when they're numeric.

PR #980, opened 2018-02-01
Update Jest to v22

wbinnssmith:
This updates Jest and babel-jest to version 22. 
* `babel-jest` no longer needs explicit configuration
*  I found no modules ending in `.es6` to warrant the additional configuration. 
* Jest also appears to handle `.json` extensions as requireable modules now.
* Removed the calls to `dontMock()` as this is now the default behavior.
* It looks all the features of node we use don't require the `--harmony` in node these days, so that was removed in favor of simplifying the test script to simply `jest`.

Was there a reason we were using a -fb suffixed version number?

Test plan: `jest` and smoke check the chrome extension and standalone app.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ install:
   - npm install
 language: node_js
 node_js:
-  - '4'
+  - '8'
 script:
   - npm run lint
   - flow check
--- a/agent/__tests__/Agent-test.js
+++ b/agent/__tests__/Agent-test.js
@@ -9,7 +9,6 @@
  */
 'use strict';
 
-jest.dontMock('../Agent');
 var Agent = require('../Agent');
 
 describe('Agent', () => {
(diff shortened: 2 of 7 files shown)


wbinnssmith:
The test failure here is due to Travis using Node 4. It should disappear when it gets updated :)

gaearon:
Do you mean we should bump Travis config to 6?

wbinnssmith:
Yeah, or 8 or the latest LTS. I did this over in #978 since I updated the travis.yml to leverage the yarn lock as a cache key. Just can't express a dependency on that PR :\

Hopefully this is all right! I didn't see much of a reason to continue with 4 since we're likely not developing on that version.

bvaughn:
> I didn't see much of a reason to continue with 4 since we're likely not developing on that version.

👍

PR #991, opened 2018-02-20
deleted repeated JSX part

jcdby:
deleted repeated JSX part in render function in the TreeView Component.


gaearon:
Thanks, but we generally try to avoid stylistic refactorings that don't also fix some issues. In this case I'm not sure duplication is problematic so I'd prefer to keep it as is.

PR #997, opened 2018-03-09
Support hiding components in devtools with Symbol.for('react.devtools.hide')

Jessidhia:
This allows places such as HOC factory functions to define a `[Symbol.for('react.devtools.hide')] = true` property in their generated components in order to support hiding them in the devtools. 

The change done to `test/example/target.js`'s `wrapWithHoc` function for testing is such an example.

image

image

No attempt to make the settings window look nice was made, this is also untested in Fiber but if I understand the code correctly it _should_ work.

Guidance on how to test or how to write tests would be welcome.

### DIFF
--- a/backend/getData.js
+++ b/backend/getData.js
@@ -160,8 +160,10 @@ function getData(internalInstance: Object): DataType {
     };
   }
 
-  const hideSymbol = nodeType === 'Composite' && typeof Symbol === 'function' && type[Symbol.for('react.devtools.hide')];
-
+  const hideSymbol = nodeType === 'Composite' &&
+    typeof Symbol === 'function' &&
+    typeof type === 'function' &&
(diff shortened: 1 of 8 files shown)


Gvozd:
@Kovensky I rebase your PR on current master - at PR #1187
Also I add option for hide components with `displayName` property

eps1lon:
So the basic usage is supposed to be:
```js
function Component(props) { ... }
Component[Symbol.for('react.devtools.hide')] = true
```
Just wanted to add that this can cause trouble if that component is decorated with other HOCs that use the popular `hoist-non-react-statics` package (used in `react-redux`, `react-router`, `emotion` etc.). It will cause every HOC upwards to be hidden as well because that property will be hoisted.

I would therefore prefer #1076.

bvaughn:

PR #1000, opened 2018-03-15
Added forwardRef support to DevTools

bvaughn:
Added support for the upcoming `React.forwardRef` API (see [RFC](https://github.com/bvaughn/rfcs/blob/ref-forwarding/text/0000-ref-forwarding.md) and [implementation](https://github.com/facebook/react/pull/12346)).

This new component type displays a `ForwardRef(Name)` (where "Name" is the name of the render function passed to `React.forwardRef()` or "anonymous" if not named). For example:
```js
// ForwardRef(anonymous)
return React.forwardRef(
  (props, ref) => <ThemedComponent {...props} forwardedRef={ref} />
);

// ForwardRef(ThemeComponentRef)
return React.forwardRef(
  function ThemeComponentRef(props, ref) {
    return <ThemedComponent {...props} forwardedRef={ref} />
  }
);
```

I also decided to try changing the color for special node types `Context.Provider`, `Context.Consumer`, `AsyncMode`, `StrictMode`, and `ForwardRef` so they stood out from class components and host components. I'm not sure how I feel about this change.

Below are some screenshots showing how the following app renders:
```js
<StrictMode>
  <div>
    <AsyncMode>
      <>
        <FancyThemedButton
          label="Click me!"
          onClick={handleClick}
          ref={fancyButtonRef}
        />
        Hi
      </>
    </AsyncMode>
  </div>
</StrictMode>
```

### Chrome (light)
screen shot 2018-03-20 at 10 48 52 am

### Chrome (dark)
screen shot 2018-03-20 at 10 49 22 am

### Firefox (light)
screen shot 2018-03-20 at 10 49 40 am

### Firefox (dark)
screen shot 2018-03-20 at 10 50 14 am


bvaughn:
Friendly ping 😄

gaearon:
>or "anonymous" if not named

How do you feel about just keeping it `ForwardRef` in this case?

>I also decided to try changing the color for special node types Context.Provider, Context.Consumer, AsyncMode, StrictMode, and ForwardRef so they stood out from class components and host components. 

I think it's a good idea but we make them *less* prominent rather than just different?

bvaughn:
> How do you feel about just keeping it `ForwardRef` in this case?

I'm okay with that. 😄

> I think it's a good idea but we make them less prominent rather than just different?

That's a bit hard to do within the confines of themes. I'll try to pick a more similar color to class components I guess?

bvaughn:
Okay. I tried to lighten the contrast a little by using a different theme color.

PR #1005, opened 2018-03-23
Fix Highlight Updates

gaearon:
Two fixes:

* functional components and context consumers should get highlighted too (they were skipped before)
* fixes a regression from https://github.com/facebook/react-devtools/pull/977


gaearon:
I'll get this in but happy to follow up if you're prefer something changed.

Issue #1013, opened 2018-04-07
Fail to build chrome extension in windows cmd


lshgdut:
I have try to build chrome extension in windows cmd, but it fail to execute `NODE_ENV=production xxxx`.

The environment variable can't be set in one command line, so it have to been put into `env` variable in exec options

lshgdut:
i have already post a pr [1014](https://github.com/facebook/react-devtools/pull/1014) to fix this issue :)

PR #1014, opened 2018-04-07
fix: build failed in windows cmd

lshgdut:
Put `NODE_ENV` variable into `options.env` to resolve the issue #1013


gaearon:
Thanks

PR #1024, opened 2018-05-01
Use PerformedWork effect to tell if a context consumer re-rendered

gaearon:
This depends on https://github.com/facebook/react/pull/12729. It should avoid false positives in Highlight Updates handling of context consumers in an edge case when there's a referential bailout.

The idea is that we should just tag whatever re-rendered (and not just classes) with `PerformedWork`. I forgot about this flag when I added context consumer highlights in https://github.com/facebook/react-devtools/pull/1005.

The downside of this is that people who get this version of DevTools won't see highlights for context consumers for versions of React between 16.3.0 to 16.3.3. I think it's okay and it's better if the feature works consistently for future users.

I like that this lets us simplify the logic and remove the dependency on `updateQueue.hasForceUpdate`. If there was a forced update, *the work was performed* so we don't need to check for that specifically.

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -14,8 +14,10 @@ import type {Hook, ReactRenderer, Helpers} from './types';
 var getDataFiber = require('./getDataFiber');
 var {
   ClassComponent,
+  FunctionalComponent,
   ContextConsumer,
   HostRoot,
+  ForwardRef,
 } = require('./ReactTypeOfWork');
 
 // Inlined from ReactTypeOfSideEffect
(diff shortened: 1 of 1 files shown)


backend/attachRendererFiber.js
```diff
@@ -42,35 +44,26 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
...
   function hasDataChanged(prevFiber, nextFiber) {
-    if (prevFiber.tag === ClassComponent) {
-      // Skip if the class performed no work (shouldComponentUpdate bailout).
-      // eslint-disable-next-line no-bitwise
-      if ((nextFiber.effectTag & PerformedWork) !== PerformedWork) {
-        return false;
-      }
-
-      // Only classes have context.
-      if (prevFiber.stateNode.context !== nextFiber.stateNode.context) {
```
gaearon: I don't see how this could ever work.

backend/attachRendererFiber.js
```diff
@@ -42,35 +44,26 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
...
-        return true;
-      }
-      // Force updating won't update state or props.
-      if (nextFiber.updateQueue != null && nextFiber.updateQueue.hasForceUpdate) {
```
gaearon: Don't need this because "performed work" includes forced updates (as long as we ran render, we performed work).

backend/attachRendererFiber.js
```diff
@@ -42,35 +44,26 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
...
-        return true;
-      }
-    }
-    // Always treat context consumers as changed.
-    // This ensures we don't skip highlighting their updates.
-    if (prevFiber.tag === ContextConsumer) {
-      return true;
```
gaearon: This was too permissive and led to false positives.

backend/attachRendererFiber.js
```diff
@@ -42,35 +44,26 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
...
+          prevFiber.ref !== nextFiber.ref
+        );
     }
-    // Compare the fields that would result in observable changes in DevTools.
-    // We don't compare type, tag, index, and key, because these are known to match.
-    return (
-      prevFiber.memoizedProps !== nextFiber.memoizedProps ||
-      prevFiber.memoizedState !== nextFiber.memoizedState ||
-      prevFiber.ref !== nextFiber.ref ||
-      prevFiber._debugSource !== nextFiber._debugSource
```
gaearon: I wrote this initially but I don't think this was useful.

PR #1030, opened 2018-05-17
Added <Profiler> component type

bvaughn:
DevTools doesn't yet do anything interesting with this component type. Just shows the correct name. More to come!

cc @cyan33

screen shot 2018-05-17 at 8 58 28 am


cyan33:
Looks good! 😄

PR #1031, opened 2018-05-19
Show displayName for Context Provider and Consumer

nanovazquez:
This PR updates the code to use the _displayName_ for the new React 16.3 Context API components. It includes the following changes:

* _Creates ContextProvider and ContextConsumer node types_ for context providers and consumers. This way, we can identify these nodes by using their types, instead of their names.
    * _Adds logic to use displayName_ for context providers and consumers' name, if present.
    * _Updates logic to colorize context providers and consumers nodes_ with the same color used for Composite nodes.
* Updates `TraceUpdatesBackendManager` to highlight updates in Context Consumer nodes. For this, it checks the type instead of the name.
* _Updates the "plain" example_ by adding `React.createContext`, using one of [ThemeContext examples in React's documentation](https://reactjs.org/docs/context.html#when-to-use-context).

### Before
React Devtools wasn't picking up the `displayName` or colorizing these nodes.

image

### After
```js
const ThemeContext = React.createContext();
ThemeContext.Provider.displayName = 'ThemeContext.Provider';
ThemeContext.Consumer.displayName = 'ThemeContext.Consumer';
```
image

If there is no display name in the Context components, React Devtools falls back to default values (`<Context.Provider>` and `<Context.Consumer>` respectively).

image


benwiley4000:
I'd love for this to be merged - @nanovazquez are you able to solve the merge conflict?

test/example/target.js
```diff
@@ -397,29 +455,47 @@ for (var mCount = 200; mCount--;) {
...
+          <div style={styles.iframeWatermark}>this is an iframe</div>
+          {/* for testing highlighing in the presence of multiple scrolls
+          {long(long(long()))} {/* */}
+          <Todos/>
+          {/*<span thing={someVal}/>
+          <Target count={1}/>
+          <span awesome={2} thing={[1,2,3]} more={{2:3}}/>
+          <span val={null}/>
+          <span val={undefined}/>
+          <div>&lt;</div>*/}
```
benwiley4000: Did you mean to delete these comments?

nanovazquez:
@benwiley4000 done! 😄

bvaughn:
I have not thought about this much, but my initial thought is that this would be a preferable API:
```js
const ThemeContext = React.createContext();
ThemeContext.displayName = 'ThemeContext';
```

FWIW you can probably avoid most of the merge conflicts that require rebasing by not committing files from `test/example/build`. (Those can be regenerated easily using `yarn build:example` by the reviewer.)

benwiley4000:
@bvaughn I agree it would be nice to be able to not repeat yourself, but I think the least surprising thing would be to *at least* support putting the displayName directly on the consumer and the provider (which works for every other type of react component), and then maybe what you suggested.

bvaughn:
> but I think the least surprising thing would be to at least support putting the displayName directly on the consumer and the provider (which works for every other type of react component)

I don't think this is really true? It works for _user-defined_ function and class components– and the user-supplied render function passed to `forwardRef`– but you wouldn't expect to put a display name on e.g. `StrictMode` or `unstable_AsyncMode` types. And since these types are `Symbol`s, it wouldn't work if you tried.

benwiley4000:
@bvaughn ok sorry I did misspeak. I was really referring to user-defined components. The thing is that even though context components are created via a special factory, you (meaning I :) ) expect to be able to name them directly anyway because they feel like components *I* created. Unlike StrictMode or the like, Context.Consumer is specially instanced for my own purposes.

bvaughn:
Yeah, I can understand that point of view. I think my preference is still the same though. 😄

Another tiny reason why naming the context object feels preferable is that it avoids potentially mismatched names.

benwiley4000:
Yeah, that's a good point

bvaughn:
I'm going to plug in my own change and do a little testing. If it seems good, I'll merge.

bvaughn:
Thanks! I've merged this. As of the next release, this should be supported:
```js
const ThemeContext = React.createContext();
ThemeContext.displayName = 'ThemeContext';
```

benwiley4000:
Thanks!!

bvaughn:
Thanks for your patience 😄

bvaughn:
Released with 3.4.0

jaydenseric:
It doesn't feel intuitive that you can't set the display name separately on provider and consumer components; they are different components after all. A lot of libraries export a provider, called things like `ExampleProvider`, and don't necessarily export a consumer (it may be used internally in another exported component).

Here is an intuitive API:

```js
// example-lib.js

import React from 'react'

const {
  Provider: ExampleProvider,
  Consumer: ExampleConsumer
} = React.createContext()

ExampleProvider.displayName = 'ExampleProvider'
ExampleConsumer.displayName = 'ExampleConsumer'

export { ExampleProvider, ExampleConsumer }
```

```js
import React from 'react'
import { ExampleProvider } from 'example-lib'

const example = ''

const App = () => (
  <ExampleProvider value={example}>
    {/* Stuff */}
  </ExampleProvider>
)
```

It should display exactly as `ExampleProvider` in dev tools, as that is the name of the component users are importing and using.

If I were to rethink the context API to make this easier, perhaps it could be:

```js
import React from 'react'

const {
  ExampleProvider,
  ExampleConsumer
} = React.createContext(
  // Custom provider component name:
  'ExampleProvider',
  // Custom consumer component name:
  'ExampleProvider'
)
```

bvaughn:
`React.createContext()` already accepts a value– the default value for context. (This value also enables static type checking via Flow.) Customizing the display name of a component in DevTools is an advanced, nuanced case and I don't think it will ever be directly referenced as part of any of React's APIs.

I don't personally see why "ExampleProvider" and "ExampleConsumer" are inherently better than "Example.Provider" and "Example.Consumer". If anything, the latter provides a bit more of a visual cue that the two components are closely related as part of a single thing (the "Example" context).

benwiley4000:
Agree with @bvaughn's reasoning, especially because `ExampleProvider` is the name I end up giving to the wrapper component which holds all the "smarts". :)

armand1m:
@jaydenseric why would you want to give a different name to the provider and the consumer since they rely in the same context?

bisubus:
From this PR, I'm still unsure how does it work if we set `displayName` only once?
```
const ThemeContext = React.createContext();
ThemeContext.displayName = 'MyThemeContext';
```
I would expect that `<ThemeContext.Consumer>` will be shown as `<MyThemeContext.Consumer>` because `ThemeContext` is actually a consumer, `ThemeContext === ThemeContext.Consumer`>. And `<ThemeContext.Provider>` will be shown as `<MyThemeContext.Provider>`, because a consumer and a provider are closely related.

I supposed this is roughly the same as [this comment](https://github.com/facebook/react-devtools/pull/1031#issuecomment-423234994) says about.

PR #1046, opened 2018-06-16
Upgrade version of React used by DevTools (15.4 -> 16.4)

bvaughn:
* Upgrade `react` and `react-dom` dependencies (15.4 -> 16.4)
* Replace `React.createClass` with class components
* Replace `React.PropTypes` with `prop-types` package


frontend/Node.js
```diff
@@ -10,6 +10,8 @@
  */
 'use strict';
 
+const PropTypes = require('prop-types');
```
gaearon: const/var is inconsistent but I don't care

bvaughn: Yeah. This is the codemod's doing. It tries to match what's in the file (based on the presence of other `const`/`let` keywords). We're inconsistent in a lot of places in the DevTools codebase, unfortunately.

frontend/PropVal.js
```diff
@@ -149,49 +149,47 @@ function previewProp(val: any, nested: boolean, inverted: boolean, theme: Theme)
 }
 
 function previewArray(val, inverted, theme) {
-  var items = {};
+  var items = [];
   val.slice(0, 3).forEach((item, i) => {
-    items['n' + i] = <PropVal val={item} nested={true} inverted={inverted} theme={theme} />;
-    items['c' + i] = ', ';
+    if (i > 0) {
+      items.push(<span key={i+','}>{', '}</span>);
```
gaearon: Do we ever need a span here? Seems like just text would do.

bvaughn: Huh, at first I thought we did– to preserve white space– but I think you're right 😄

Issue #1039, opened 2018-05-29
nativeObjectCreate


LuisGonzaga:
Hello,
I am starting React & doing an "hello world", to my surprise I got the error described below, after searching the web (no similar results) I tested in FF and IE with no problems.
Then I disabled react-devtools 3.2.3 and everything was fine.

-> The script is very simple, just for the record: 
<script type="text/javascript">
      var helloWorld = React.createElement('h1', null, 'Hello world!')
      ReactDOM.render(helloWorld, document.getElementById('content'))    
</script>

-> error in Chrome  66.0.3359.181 (64 bits)
-> works fine in Firefox 60.0.1 (64-bit)

-> The error:
backend.js:708 Uncaught TypeError: Cannot read property 'nativeObjectCreate' of undefined
    at Object.<anonymous> (backend.js:708)
    at __webpack_require__ (backend.js:20)
    at Object.<anonymous> (backend.js:690)
    at __webpack_require__ (backend.js:20)
    at Object.<anonymous> (backend.js:59)
    at __webpack_require__ (backend.js:20)
    at backend.js:40
    at backend.js:43
(anonymous) @ backend.js:708
__webpack_require__ @ backend.js:20
(anonymous) @ backend.js:690
__webpack_require__ @ backend.js:20
(anonymous) @ backend.js:59
__webpack_require__ @ backend.js:20
(anonymous) @ backend.js:40
(anonymous) @ backend.js:43

madlabman:
I have similar error using React dev tools in Chrome but it's work fine in Firefox, check it out.

ashsearle:
I've seen the same, and think this is the minimum required to reproduce:

* Create an HTML file containing:
```html
<script src="https://unpkg.com/react@^16/umd/react.development.js"></script>
```

* Use "Open File..." in Chrome to open it using `file:///` protocol.
* Inspect page
* Switch to React tab

Observe 1 JavaScript error.  Switch to Console tab to view stack trace.

Serving the same file over http does not give any errors, just the "Waiting for roots to load..."  messages in React devtools.

LuisGonzaga:
Well I believe that this is a conflict with other Extensions.
In my particular case I've disabled Ember Inspector, Ghost Text, "Screenshot, Annotate, & Screen Video Recorder..." plus "Sencha & Ext.js" ... and it started to work.

ashsearle:
@LuisGonzaga I still see the same error even if I disable all other extensions.

[...]

PR #1061, opened 2018-07-17
Don't init modules until hook exists

gaearon:
Fixes https://github.com/facebook/react-devtools/issues/1039.

The issue was that we read [stuff like this](https://github.com/facebook/react-devtools/blob/faa4b630a8c055d5ab4ff51536f1e92604d5c09c/shells/webextension/webpack.backend.js#L35-L38) from the hook but this assumes the hook exists. The hook won't exist if we didn't get a permission, such as if we're loading a `file://` URL and our extension doesn't have access to them.

https://github.com/facebook/react-devtools/issues/1039#issuecomment-399901658 has exact reproduction instructions, but note that you'll need to disable `file://` access to reproduce.

Even with the fix, we still have a React tab that can't ever connect. I'm not sure how to fix this, but for now I added a link to troubleshooting instructions once you wait over 3 seconds. I think there were some other cases where this could happen, and this links to a FAQ.

### DIFF
--- a/flow/chrome.js
+++ b/flow/chrome.js
@@ -35,6 +35,7 @@ declare var chrome: {
     },
   },
   tabs: {
+    create: (options: Object) => void,
     executeScript: (tabId: number, options: Object, fn: () => void) => void,
     onUpdated: {
       addListener: (fn: (tabId: number, changeInfo: Object, tab: Object) => void) => void,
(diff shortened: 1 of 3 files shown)


gaearon:
I'll get this in for now because [this looks bad](https://mobile.twitter.com/ReactStudent/status/1018741404598538241), if there's a better way happy to follow up

PR #1063, opened 2018-07-18
Rename Timeout to Placeholder

cyan33:
I somehow suddenly remembered that we also need to do the renaming stuff in dev-tools as well after https://github.com/facebook/react/pull/13105. 😄


cyan33:
cc @bvaughn

Issue #992, opened 2018-02-20
Stay on Selected Component After Refreshing


xiao-vincent:
Is there a way to stay on the selected component after refreshing the browser?

It's a bit tedious to search for the component again after making changes to the code.

emil14:
+1

ultrox:
I was looking the same thing, its so annoying my filters gets cleared out every single time I refresh.

OrDuan:
Big pain point for me too.
Just opened a [PR](https://github.com/facebook/react-devtools/pull/1067)

BrentLabasan:
In addition, having whichever props and/or states that were expanded before, continue to be expanded after a page refresh, would be a huge time saver.

This would be such a huge time saver.

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #1067, opened 2018-07-26
Persist the search term between refreshes

OrDuan:
Saves the last searched text in local storage so when the user refresh the page(auto reload or manually) - the search text appears and shows the results.

Fixes:
- [992](https://github.com/facebook/react-devtools/issues/992)
- [295](https://github.com/facebook/react-devtools/issues/295)

Would be great to get feedback about:
- Is using the local storage is a bad practice?
- When should the text be cleared from the local storage?(if ever)

### DIFF
--- a/frontend/SettingsPane.js
+++ b/frontend/SettingsPane.js
@@ -18,7 +18,7 @@ const SvgIcon = require('./SvgIcon');
 const Icons = require('./Icons');
 const Input = require('./Input');
 const Hoverable = require('./Hoverable');
-
+const storage = require('../utils/storage');
 const decorate = require('./decorate');
 
 import type {Theme} from './types';
@@ -48,9 +48,9 @@ class SettingsPane extends React.Component {
     // capture=true is needed to prevent chrome devtools console popping up
     doc.addEventListener('keydown', this._key, true);
 
-    const lastSearch = localStorage.getItem('searchText') || '';
-    if (lastSearch !== '') {
-      this.props.onChangeSearch(lastSearch);
+    const lastSearchedTerm = storage.get('searchText', '');
+    if (lastSearchedTerm) {
+      this.props.onChangeSearch(lastSearchedTerm);
     }
   }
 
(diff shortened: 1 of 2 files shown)


frontend/SettingsPane.js
```diff
@@ -50,6 +50,11 @@ class SettingsPane extends React.Component {
     const doc = ReactDOM.findDOMNode(this).ownerDocument;
     // capture=true is needed to prevent chrome devtools console popping up
     doc.addEventListener('keydown', this._key, true);
+
+    const lastSearch = localStorage.getItem('searchText') || '';
+    if (lastSearch !== '') {
+      this.props.onChangeSearch(lastSearch);
```
bvaughn: I don't think this is the best place to do this 😄 We should re-initialize this value in `Store` so we can avoid rendering multiple times

OrDuan:
@bvaughn thanks for the feedback!
I updated the code to the the storage utility. 
I can not find an easy way to init the `Store` with the local storage data, I tried the contractor and the `_mountComponent` both called too early, while the  input's text has the last searched text, it doesn't trigger the search itself to reveal the nodes.

Where would you write the logic that fire the store's `changeSearch`?

bvaughn:
When the search text changes, [`SettingsPane` calls `props.onChangeSearch`](https://github.com/facebook/react-devtools/blob/e863aad2a4a049c59f754a53b16cab6cfa68e5fb/frontend/SettingsPane.js#L133) which is [connected to `store.changeSearch`](https://github.com/facebook/react-devtools/blob/e863aad2a4a049c59f754a53b16cab6cfa68e5fb/frontend/SettingsPane.js#L176).

So my initial thought would be to check in the `Store` constructor, and if a previous saved search exists, auto-filter by calling `changeSearch`– something like:

```js
const previousSearchText = storage.get(SEARCH_TEXT, ""); // Defaults to ""
if (previousSearchText !== "") {
  this.changeSearch(previousSearchText);
}
```

I also suggest using a constant (e.g. `SEARCH_TEXT`) where you get and set the text rather than an inline string, just to better keep them in sync.

bvaughn:
Thanks for elaborating.

I think we should try to dig in further and find out _why_ it didn't work. It still feels a bit gross to me to do this in a component's did-mount hook via a prop callback.

bvaughn:

PR #1069, opened 2018-07-27
Profiler plug-in for DevTools

bvaughn:
Builds on top of the [experimental Profiler API](https://github.com/reactjs/rfcs/pull/51) and the "interactions" POC (facebook/react/pull/13253).

### Blocking Dependencies
- [x] `Profiler` duration bugfix: https://github.com/facebook/react/pull/13313

### Non-blocking Dependencies
The profiler will degrade gracefully if the interaction-tracking features are not available at runtime.
- [ ] `interaction-tracking` package: https://github.com/facebook/react/pull/13234
- [ ] `Profiler` and `interaction-tracking` integration: https://github.com/facebook/react/pull/13253

### Follow Up Work
- [ ] The profiling detection capabilities I added to [the Profiler backend](https://github.com/bvaughn/react-devtools/blob/7958a96d962cfe1418dc9139d3a823ab0dbd89db/plugins/Profiler/backend.js#L19-L40)  are, I think, better than the approach used in [the Relay plugin](https://github.com/facebook/react-devtools/blob/4d3ae51dc31617bd813339fb7baa00e0154a0023/plugins/Relay/RelayPlugin.js#L37-L45). We might consider porting the approach over at some point.
- [ ] I've also added a new, lightweight [shared "storage" abstraction](https://github.com/bvaughn/react-devtools/blob/profiler-poc/utils/storage.js) that's used by the settings panel (to remember the theme) and by the new profiler plug-in (to remember the active chart). Based on [recent Twitter feedback](https://twitter.com/brian_d_vaughn/status/1025818107305054214) we might consider plugging this into other parts of DevTools to remember more things between sessions.

### Demo
Video walk throughs of profiler UI can be found [here](https://www.youtube.com/watch?v=-Ud_265QAwM) and [here](https://www.youtube.com/watch?v=vfm4018YBEY). Below are some screenshots.

##### No profiling data
screen shot 2018-08-09 at 1 51 47 pm

##### Recording
screen shot 2018-08-09 at 1 51 57 pm

##### No timing data for commit
screen shot 2018-08-09 at 1 52 15 pm

##### Flame chart
screen shot 2018-08-09 at 1 52 25 pm

##### Ranked chart
screen shot 2018-08-09 at 1 52 32 pm

##### Interactions
screen shot 2018-08-09 at 1 52 46 pm

##### No Interactions
screen shot 2018-08-09 at 1 53 21 pm

##### Fiber render chart
screen shot 2018-08-09 at 1 52 56 pm


bvaughn:
I don't think the profiler plug-in will be able to support "reload and start profiling" functionality like the browser's DevTools do, because React loads and starts running before DevTools has been initialized.

I did a little investigation into this though, so I'll leave a patch here in case we decide to pick it back up at some point:
```diff
diff --git a/plugins/Profiler/ProfilerBrowserApi.js b/plugins/Profiler/ProfilerBrowserApi.js
new file mode 100644
index 0000000..21b9510
--- /dev/null
+++ b/plugins/Profiler/ProfilerBrowserApi.js
@@ -0,0 +1,19 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ *
+ * @flow
+ */
+'use strict';
+
+// TODO (bvaughn) Move these functions to a more central location.
+
+export const isChromeApiAvailable =
+  chrome !== null &&
+  typeof chrome === 'object' &&
+  chrome.storage !== null &&
+  typeof chrome.storage === 'object';
diff --git a/plugins/Profiler/ProfilerStore.js b/plugins/Profiler/ProfilerStore.js
index caa2aa7..e0afa56 100644
--- a/plugins/Profiler/ProfilerStore.js
+++ b/plugins/Profiler/ProfilerStore.js
@@ -15,6 +15,7 @@ import type {Interaction, RootProfilerData, Snapshot} from './ProfilerTypes';
 
 const {List} = require('immutable');
 const {EventEmitter} = require('events');
+const {isChromeApiAvailable} = require('./ProfilerBrowserApi');
 
 class ProfilerStore extends EventEmitter {
   _bridge: Bridge;
@@ -36,6 +37,18 @@ class ProfilerStore extends EventEmitter {
     this._mainStore.on('roots', this.saveRoots);
     this._mainStore.on('selected', this.updateSelected);
     this._mainStore.on('storeSnapshot', this.storeSnapshot);
+
+    // If we're auto-profiling...
+    if (isChromeApiAvailable) {
+      try {
+        chrome.storage.sync.get(['recordOnStartup'], (data) => {
+          if (data && data.recordOnStartup) {
+            this.setIsRecording(true);
+          }
+          chrome.storage.sync.set({recordOnStartup: false});
+        });
+      } catch (error) {}
+    }
   }
 
   off() {
diff --git a/plugins/Profiler/views/ProfilerTabToolbar.js b/plugins/Profiler/views/ProfilerTabToolbar.js
index a16fbdf..b5e51c3 100644
--- a/plugins/Profiler/views/ProfilerTabToolbar.js
+++ b/plugins/Profiler/views/ProfilerTabToolbar.js
@@ -20,6 +20,7 @@ import Hoverable from '../../../frontend/Hoverable';
 import SvgIcon from '../../../frontend/SvgIcon';
 import Icons from '../../../frontend/Icons';
 import IconButton from './IconButton';
+import {isChromeApiAvailable} from '../ProfilerBrowserApi';
 
 const CHART_LABEL_WIDTH_THRESHOLD = 615;
 
@@ -99,6 +100,12 @@ const ProfilerTabToolbar = ({
       theme={theme}
     />
 
+    <ReloadAndRecord
+      isDisabled={!isChromeApiAvailable}
+      onClick={reloadAndRecord}
+      theme={theme}
+    />
+
     <HRule theme={theme} />
 
     {!isRecording && snapshots.length > 0 && (
@@ -253,3 +260,51 @@ const RecordButton = Hoverable(
     </button>
   )
 );
+
+const ReloadAndRecord = Hoverable(
+  ({ isDisabled, isHovered, onClick, onMouseEnter, onMouseLeave, theme }) => (
+    <button
+      disabled={isDisabled}
+      onClick={onClick}
+      onMouseEnter={onMouseEnter}
+      onMouseLeave={onMouseLeave}
+      style={{
+        background: 'none',
+        border: 'none',
+        outline: 'none',
+        cursor: 'pointer',
+        color: isHovered ? theme.state06 : theme.base05,
+        padding: '4px',
+        opacity: isDisabled ? 0.5 : 1,
+        pointerEvents: isDisabled ? 'none' : 'auto',
+      }}
+      title={'Start profiling and reload page'}
+    >
+      <SvgIcon
+        path={Icons.REFRESH}
+        style={{
+          flex: '0 0 1rem',
+          width: '1rem',
+          height: '1rem',
+          fill: 'currentColor',
+          display: 'inline',
+          verticalAlign: 'sub',
+        }}
+      />
+    </button>
+  )
+);
+
+const reloadAndRecord = () => {
+  if (isChromeApiAvailable) {
+    chrome.storage.sync.set({recordOnStartup: true},
+      () => {
+        // TODO (bvaughn) This only works if DevTools is docked to the current window.
+        chrome.tabs.query(
+          {active: true, lastFocusedWindow: true},
+          ([tab]) => tab && chrome.tabs.reload(tab.id)
+        );
+      }
+    );
+  }
+};
```

agent/inject.js
```diff
@@ -21,9 +21,11 @@ module.exports = function(hook: Hook, agent: Agent) {
       agent.setReactInternals(id, helpers);
       helpers.walkTree(agent.onMounted.bind(agent, id), agent.addRoot.bind(agent, id));
     }),
-    hook.sub('root', ({renderer, internalInstance}) => agent.addRoot(renderer, internalInstance)),
+    hook.sub('rootCommitted', ({renderer, internalInstance}) => agent.rootCommitted(renderer, internalInstance)),
     hook.sub('mount', ({renderer, internalInstance, data}) => agent.onMounted(renderer, internalInstance, data)),
+    hook.sub('root', ({renderer, internalInstance}) => agent.addRoot(renderer, internalInstance)),
     hook.sub('update', ({renderer, internalInstance, data}) => agent.onUpdated(internalInstance, data)),
+    hook.sub('updateProfileTimes', ({renderer, internalInstance, data}) => agent.onUpdatedProfileTimes(internalInstance, data)),
     hook.sub('unmount', ({renderer, internalInstance}) => agent.onUnmounted(internalInstance)),
```
gaearon: Is it worth keeping "basic" events necessary for the tree separate from "profiler" events?

plugins/Profiler/backend.js
```diff
@@ -0,0 +1,44 @@
...
+const emptyFunction = () => {};
+
+module.exports = (bridge: Bridge, agent: Agent, hook: Object) => {
+  const checkIfProfilingIsSupported = () => {
+    let profilingIsSupported = false;
+
+    // Feature detection for profiling mode.
+    // The presence of an "actualDuration" field signifies:
+    // 1) This is a new enough version of React
+    // 2) This is a profiling capable bundle (e.g. DEV or PROFILING)
```
gaearon: I forgot PROFILING is enabled in DEV too.

Is there any overhead in this implementation for DevTools running React in DEV mode while we're not recording?

plugins/Profiler/views/NoInteractionsMessage.js
```diff
@@ -0,0 +1,57 @@
...
+      }}>
+        {headerText}
+      </p>
+      {hasMultipleRoots && (
+        <p>
+          You may want to select a different root in the <strong>Elements</strong> panel.
+        </p>
+      )}
+      <p>
+        <a href="https://reactjs.org/docs/interaction-tracking.html" target="_blank">
```
gaearon: Can we make this an fb.me URL and then add a todo list to point it to the right thing together with the release that includes it?

agent/inject.js
bvaughn: Not sure what you're suggesting.

The added "updateProfileTimes" event was necessary to handle the case where a component bailed out– something DevTools traditionally didn't care about, but now needs to handle so that we can update the `actualDuration`

plugins/Profiler/backend.js
bvaughn: > Is there any overhead in this implementation for DevTools running React in DEV mode while we're not recording?

Yes and no. React will automatically opt roots into `ProfileMode` if DevTools is present. We have to do this in order to maintain accurate tree base times in the event that someone actually profiles.

But that behavior is in the React repo already, aside from this release. I don't think we have much new overhead in DevTools itself from this release _unless_ someone is actually profiling. (`ProfileCollector` doesn't gather data unless the user is recording.)

bvaughn:
> I see a lot of copying — wondering if you have any memory usage concern, and whether it's likely the integration will introduce more GC pressure to potentially skew the measurements.

It's possible, although I think the impact of this is minimal (compared to the immutable data structure we're already maintaining) _unless_ you actually _profile_ in which case... I still think we're maybe doing the best thing I could think of.

> I'd like to play with it in the coming days (maybe let's set up a video call)?

Sounds good to me!

plugins/Profiler/views/NoInteractionsMessage.js
bvaughn: Sure. Done.

Although I think fb.me URLs don't support being edited after they're created– so I'm not sure how much value this actually provides vs a direct reactjs.org link.

agent/inject.js
gaearon: Sorry, I proposed to separate 

```js
    // Basic functionality:
    hook.sub('mount', ({renderer, internalInstance, data}) => agent.onMounted(renderer, internalInstance, data)),	    hook.sub('mount', ({renderer, internalInstance, data}) => agent.onMounted(renderer, internalInstance, data)),
    hook.sub('update', ({renderer, internalInstance, data}) => agent.onUpdated(internalInstance, data)),	    hook.sub('update', ({renderer, internalInstance, data}) => agent.onUpdated(internalInstance, data)),
    hook.sub('unmount', ({renderer, internalInstance}) => agent.onUnmounted(internalInstance)),	    hook.sub('unmount', ({renderer, internalInstance}) => agent.onUnmounted(internalInstance)),

    // Necessary for profiler:
    hook.sub('root', ({renderer, internalInstance}) => agent.addRoot(renderer, internalInstance)),
    hook.sub('rootCommitted', ({renderer, internalInstance}) => agent.rootCommitted(renderer, internalInstance)),
    hook.sub('updateProfileTimes', ({renderer, internalInstance, data}) => agent.onUpdatedProfileTimes(internalInstance, data)),
```

So it's clearer the new set of events are additional and aren't necessary for the main part to work.

plugins/Profiler/views/NoInteractionsMessage.js
gaearon: They do support being edited.

bvaughn: I thought they used to! But the last two times I've looked for a way to edit one, it doesn't appear to exist anymore, and the tool specifically says that you shouldn't edit them because they might be cached externally. Maybe this changed recently?

bvaughn: Or maybe I just overlooked the dialog 😄 Hopefully that's all!

agent/inject.js
bvaughn: Ah! Yeah that makes sense 😄 I misunderstood.

plugins/Profiler/views/NoInteractionsMessage.js
gaearon: It's weird because there are two tools.

Go to the one that says "avoid" and then add `/create/` to its URL. You'll find the one that allows it (after clicking okay in a confirmation dialog).

bvaughn: Oh nice! Thanks

bvaughn:
Okay– I'm landing this beast. I'll rebase my other less-clutter PR once it's landed and fix up the two issues you mentioned on it.

Let's setup a time (at your convenience) to do a video chat about the profiler and if anything's uncovered, I'll do a follow up.

Thanks!

jrnail23:
@bvaughn, is this feature actually supposed to work as of devTools v3.3.1?  I see the Profiler tab, and I'm able to begin recording (or at least it looks that way according to the UI), but when I finish recording, I get nothing -- it looks as if no data has been collected.
FYI, this is with React v16.4.2, in development mode.
screen shot 2018-08-29 at 10 09 07 pm

gaearon:
I thought we didn't show Profiler with 16.4.2.. 🤔

bvaughn:
> @bvaughn, is this feature actually supposed to work as of devTools v3.3.1?

No, it isn't expected to work until we release the next minor. Sorry for the confusion.

> I thought we didn't show Profiler tab with 16.4.2

The profiler shows if it detects a profiling-capable version of React, which it does by checking for the presence of a specific profiling attribute that actually _does_ exist in 16.4. In hindsight, I should have checked for `treeBaseDuration` instead.

Regardless, the plan (based on feedback from Sophie) is to show the tab always and promote upgrading to a profiling-capable build of React if it's not supported. I was planning on making that change soon. As part of that, I'll change the feature detection to use `treeBaseDuration`.

bvaughn:
Btw @jrnail23, I've just published 3.3.2 to the Chrome and Firefox stores that should fix this issue. Thanks!

jrnail23:
Got it, @bvaughn, thanks again for being so responsive!

denniscual:
@bvaughn Awesome work! I tried to install the pre-build release of devtool but the Profiler tab is not showing. Did I miss something obvious in here? Or currently, the Profiler tab was purposely disabled?

bvaughn:
The prerelease page mentions this (but maybe not boldly enough) – the profiler will only work (show up) if you're running the latest React (from master) or one of the pre-built packages also on the page.

denniscual:
Sweet thanks!

Issue #746, opened 2017-05-26
Improve scrolling/centering of nodes as you browse trees


bvaughn:
Fix horizontal offset when nodes are selected. Selecting a node will scroll vertically devtools to ensure it's visible but not horizontally. Deep trees can be confusing in this way.

Clicking on a node in devtools that's near the boundary shouldn't jump the frame either. Only jump when an element is clicked outside.

The custom scroll to method is nice in that it maintains some padding around the top/bottom as we're scrolling, but...we can improve it?

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #1072, opened 2018-08-05
improve scrolling to element

stipsan:
Saw this tweet and thought I'd give it a go: https://twitter.com/aweary/status/1025789362607255552

Related issue: #746

kapture 2018-08-05 at 22 06 22

This PR adds support for horizontal scrolling and also ensures that vertical scrolling is always in view. There's room for improvements though:
* There's a setTimeout workaround for when the height of the elements scrolling frame changes because of very long multi-row breadcrumbs to ensure things are scrolled into view correctly.
* Some things like textNodes have a very wide width triggering unnecessary scrolling.
* There's little space between the edges and the node being scrolled to, but I'll be adding support for `scroll-margin` CSS in `scroll-into-view-if-needed` soon which will allow changing that.

Benefits to using the [`scroll-into-view-if-needed`](https://scroll-into-view-if-needed.netlify.com/) ponyfill, and later possibly [`scroll-margin`](https://developers.google.com/web/updates/2018/07/css-scroll-snap#scroll_padding_and_margin):
* Standards based approach that allows removing dependencies and custom logic in the future when browser support is good enough.
* Less code to maintain in this project.

Thoughts? 😃

### DIFF
--- a/frontend/TreeView.js
+++ b/frontend/TreeView.js
@@ -18,6 +18,7 @@ var SearchUtils = require('./SearchUtils');
 
 var decorate = require('./decorate');
 var {monospace, sansSerif} = require('./Themes/Fonts');
+const scrollIntoView = require('scroll-into-view-if-needed');
 
 import type {List} from 'immutable';
 import type {Theme} from './types';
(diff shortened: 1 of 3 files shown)


bvaughn:
Hi @stipsan! Thanks so much for the contribution!

Sorry for not seeing this PR sooner. 🙁  I fixed this in a slightly different way via #1091 by [using `scrollIntoView` rather than `scrollIntoViewIfNeeded`](https://github.com/facebook/react-devtools/blob/11987182206699a1fdf75e0c0850994ea6c34aa6/frontend/Node.js#L157-L168) since the container elements are wide enough to fool the browser into thinking they're in view when the content we want to see isn't actually.

So I think this PR can be closed! (Again, my apologies)

stipsan:
No worries at all @bvaughn! 

Awesome fix btw, less polyfills less problems 🎉

PR #1089, opened 2018-08-10
Improve Firefox extension development stability for other configurations

digitarald:
These fixes helped me run the `test` command locally.

- Provide `source-dir` as argument to `web-ext` (the env var used did not match the documentation)
- Handle `async` calls more gracefully, like getting the `default` profile (which my machine does not have)


bvaughn:
Whoa! Another PR from @digitarald! Awesome 😄

Sorry for not seeing this. Lemme take a look!

PR #1091, opened 2018-08-12
Make Elements panel feel less cluttered

### LINKED ISSUE #1079: Make Elements panel feel less cluttered
Based on some user recent feedback I gathered [via Twitter](https://twitter.com/brian_d_vaughn/status/1025776259056336897), one piece of common feedback is that the Elements panel feels too noisy/cluttered/crowded.

Ritesh [points out a few possible causes for this](https://twitter.com/ritz078/status/1026164993614524416) including "line height" and too many props. I personally agree with the props concern. I like the simplicity of the Vue DevTools panel, which only shows props for the currently selected node:
image

I think we should try doing the same for React DevTools. This would probably also have the side effect of making the DevTools render a little faster as well.

bvaughn:
### Do not merge before PR #1069

Resolves #1079

Includes the following subjective tweaks:
* Tweaks margins and paddings to feel less cluttered
* Re-styles header toolbar so search feels more prominent
* Moves "Highlight updates" and "Highlight search" options into settings popup
* Changes attribute display behavior to only show strings/numbers/booleans and hide the rest from the tree view
* Rather than forcing single-line elements, added improved wrapping behavior to make tree easier to see at a glance
* Forces breadcrumb strip to only take up a single vertical line rather than wrapping
* Improves scroll-into-view functionality to also horizontally scroll

### New search toolbar and settings
settings

### New element attribute wrapping behavior
wrap

### Improved breadcrumbs and scroll-to behavior
breadcrumbs-and-snapping


gaearon:
A few minor things based on screen recordings above:

* Styles appear wrong here (I mentioned on twitter but not sure if you have a chance to fix this yet)

screen shot 2018-08-14 at 6 09 09 pm

* I dislike the "lack" of input background. It's probably not great from accessibility standpoint (very low contrast) and in general I haven't seen this pattern elsewhere. Maybe it would be better if clicking transformed it into an input?

screen shot 2018-08-14 at 6 10 00 pm

bvaughn:
The first thing you pointed out was an edge-case when the tail was clicked but then the node collapsed. It's now fixed.

As for the search input style– I like it, but...I don't feel strong enough about it to fight for it 😄so I changed its color when focused like you mentioned.

kapture 2018-08-14 at 18 28 55

bvaughn:
❤️

Issue #1084, opened 2018-08-09
Element selection pop-up font is too damn big!


javajosh:
Hey guys, I love your tool and use it almost every day. However, I find myself consistently frustrated by the size of the descriptive popup when interactively selecting React elements! It's so big it gets in the way, and obscures a huge portion of the page!

Here is an example of selecting a form field, showing the relative size of the react-devtools popup : 
<img width="389" alt="screen shot 2018-08-09 at 15 29 25" src="https://user-images.githubusercontent.com/18854/43902253-de591fc0-9be9-11e8-8e2f-be6d3c04f6d9.png">

**_The default font size should be greatly decreased, and the size of the font should be user configurable_**.

Update: this is application specific. In dev mode we set a style `document.body.style.fontSize = '144px'` so that we catch unstyled text more easily. It's a style that is designed to be ignored, but the devtools don't ignore it. Not really sure what to do. I guess I could reset that size whenever starting react-devtools, but that seems like a lot of work. Alternatively, it might be nice to offer to run user code when starting the plugin.

bvaughn:
> Alternatively, it might be nice to offer to run user code when starting the plugin.

Unless I'm misunderstanding what you mean– this seems unlikely to happen.

Perhaps we could add an inline `font-size` style to this overlay rather than inheriting the body's size. Setting a body size of 144px seems pretty unexpected/unusual though 😄

PR #1092, opened 2018-08-13
Adjust overlay font size

ngyikp:
Fix https://github.com/facebook/react-devtools/issues/1084

`monospace.sizes.normal` is set to a number:
https://github.com/facebook/react-devtools/blob/4d3ae51dc31617bd813339fb7baa00e0154a0023/frontend/Themes/Fonts.js#L17

In React, setting a font-size to a number will automatically add `px` to the back:

```js
<div style={{fontSize: 20}}>ok</div>
```

However, this is not valid on regular JS:

```
document.body.style.fontSize = 20; // wrong
```


bvaughn:
Thanks!

Issue #1099, opened 2018-08-18
React Rally profiler feedback umbrella issue


bvaughn:
I did several live demos of the new DevTools profiler with people at React Rally. Some of them used [pre-built demo apps](https://react-devtools-profiler-prerelease.now.sh/) and others profiled local builds of their own apps. Over all the feedback was very positive, but I gathered several suggestions and observations from the tests. It's likely that we may not be able to implement all of them, but they are:

#### Flamegraph
- [x] 49b9e47, 5cf07b2: Flamegraph/ranked charts should enable you to deselect the current fiber so the right hand pane shows commit statistics again. Perhaps this should be a new button in the top of the panel. Clicking in an "empty" space of the flame graph should also deselect.
- [x] e03c7da: Show un-rounded durations on hover in tooltips for more info
- [x] Confirm that rounding isn't lopping off significant digits (all times ended up .00 in one of the demos)

#### Interactions
- [x] 35c1f2a: Hide the snapshot selector in this view since it's not interactive and it may cause confusion.
- [x] 09edccb: Tie the commits more to the chart views. Maybe use the colors (based on commit duration) for the circles in the interactions view?
- [x] 09edccb: Commits are shown as bars in the charts, but in the interactions tab they're circles. Maybe they should be little colored squares? At least in the "renders" side pane, rather than showing the triangle that fills in, show the colored squares?
- [x] 09edccb: Maybe use black for selected (like we do in the snapshot selector)

#### Horizontal snapshot selector
- [x] d5047c9, ea8e92c: Make the selected commit more prominent in the horizontal selector than just an alpha transform. Too many people didn't notice it. Maybe add an overlay pattern so it stands out more?
- [x] ad38466: Left/right arrow keys should step through commits/snapshots.
- [x] a2907c8: Make sure jumping from fiber-durations bar chart view to flame/ranked also horizontally scrolls the snapshot selector.
- [x] 0c79424: When dragging, don't leave the dragging state until you release the mouse. (Even if you mouse outside of the selector.)
- [x] ~~Scrollbar gets in the way– makes it hard to see the commits.~~

#### Right hand details pane
- [x] 5cf07b2: Make the right hand details pane buttons blue so they stand out more as buttons.
- [x] 6f075a5: Display "committed at" rather than "commit time".
[...]

PR #1100, opened 2018-08-18
React Rally profiler feedback

bvaughn:
Issue #1099

Demo build available at [react-devtools-profiler-prerelease.now.sh](https://react-devtools-profiler-prerelease.now.sh/)

#### Flamegraph
- 49b9e47, 5cf07b2: Flamegraph/ranked charts should enable you to deselect the current fiber so the right hand pane shows commit statistics again. Perhaps this should be a new button in the top of the panel. Clicking in an "empty" space of the flame graph should also deselect.
- e03c7da: Show un-rounded durations on hover in tooltips for more info

#### Horizontal snapshot selector
- d5047c9, ea8e92c: Make the selected commit more prominent in the horizontal selector than just an alpha transform. Too many people didn't notice it. Maybe add an overlay pattern so it stands out more?
- ad38466: Left/right arrow keys should step through commits/snapshots.
- a2907c8: Make sure jumping from fiber-durations bar chart view to flame/ranked also horizontally scrolls the snapshot selector.
- 0c79424: When dragging, don't leave the dragging state until you release the mouse. (Even if you mouse outside of the selector.)

#### Right hand details pane
- 5cf07b2: Make the right hand details pane buttons blue so they stand out more as buttons.
- 6f075a5: Display "committed at" rather than "commit time".

#### Component durations bar chart
- 5e7de4e: It's not obvious which commits are visible in the component durations bar chart. Perhaps we could dim the commits thate are not visible in the above selector?

#### Elements tree view
- e881623: Single-click on the tree view toggle arrow doesn't expand/collapse the tree anymore. (You have to double-click it.)

#### General
- 27e1f29: "Show native nodes" probably isn't that useful– maybe add a settings/gear icon there and have a profiling preferences popup?
- 802932b, e8c4ecb: Add a preference (in the new profiler preferences popup) to filter commits below a certain duration too (to reduce noise in commit selector).


bvaughn:
This doesn't resolve all of the requests in #1099, but I think it hits most of the low hanging fruit ones. I'll follow up on some of the others later.

PR #1101, opened 2018-08-20
Support multiple versions of React constants

gaearon:
This updates DevTools to work with constants that were changed in https://github.com/facebook/react/pull/13397.

I decided to pull all constants and Fiber-specific helpers in one closure that has access to the version. We might tweak more things over time, and I think it'll be easier to update if everything is in one place.

You might find reviewing individual commits easier.


backend/attachRendererFiber.js
```diff
@@ -90,7 +426,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
   function enqueueMount(fiber) {
     pendingEvents.push({
       internalInstance: getOpaqueNode(fiber),
-      data: getDataFiber(fiber, getOpaqueNode),
```
gaearon: This is now in the closure anyway.

bvaughn:
Got some lingering references to `ReactTypeOfWork.js` and `ReactSymbols.js` in `backend/getDataFiber.js` that are failing CI

bvaughn:
In general, I think the approach to forking based on version number is fine 👍

Both `attachRendererFiber` and `getDataFiber` need access to these constant values though, so we'll have to export them somehow.

gaearon:
Oops, I meant to delete that file.

gaearon:
>Both attachRendererFiber and getDataFiber need access to these constant values though, so we'll have to export them somehow.

I inlined `getDataFiber` into `attachRendererFiber` (right into the closure).

If there's still a `getDataFiber.js` file this sounds like a mistake. Will check now.

PR #1102, opened 2018-08-20
UI tweaks

bvaughn:
* 244f3b0: Fixed spacing before "== $r" for selected head node.
* 208074a: Fixed edge-case overflow issue in right/props panel

### DIFF
--- a/frontend/SplitPane.js
+++ b/frontend/SplitPane.js
@@ -140,6 +140,7 @@ const styles = {
   rightPane: {
     display: 'flex',
     width: '100%',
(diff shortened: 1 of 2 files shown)


frontend/SplitPane.js
```diff
@@ -140,6 +140,7 @@ const styles = {
   rightPane: {
     display: 'flex',
     width: '100%',
+    overflow: 'auto',
```
bvaughn: The relative positioned drag handle was causing weird overflow behavior in this component in some cases. This style fixes it though!

PR #1103, opened 2018-08-20
Update constants and support React.lazy

gaearon:
Updates constants to https://github.com/facebook/react/pull/13444 and adds support for lazy types.

You need to update `ReactVersion` in local bundle and enable suspense to test these.


backend/attachRendererFiber.js
```diff
@@ -148,11 +155,20 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
...
+    }
+
     switch (fiber.tag) {
       case FunctionalComponent:
+      case FunctionalComponentLazy:
       case ClassComponent:
+      case ClassComponentLazy:
         nodeType = 'Composite';
-        name = getDisplayName(fiber.type);
+        name = getDisplayName(resolvedType);
```
gaearon: These shouldn't appear in the tree before type is resolved. So I'm assuming it is here.

backend/attachRendererFiber.js
```diff
@@ -174,6 +190,13 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
         }
         children = [];
         break;
+      case ForwardRef:
+      case ForwardRefLazy:
```
gaearon: It was more convenient to use tag than $$typeof here.

gaearon:
I didn't test extensively (especially with old React). But we'll test it at least a bit anyway before next release.

PR #1112, opened 2018-08-24
Profiler UI/UX tweaks and bug-fixes

### LINKED ISSUE #1104: Don't scroll node into view if it's already fully in the view
See here:

scrolling

I wouldn't expect scroll to happen because all of these nodes are already into view. The jumping happens too often and I think it's a bit jarring.

Ideally we shouldn't jump if a node is already fully in the view.

bvaughn:
Resolves #1104 and umbrella issue #1099

#### Interactions
- 09edccb: Tie the commits more to the chart views. Maybe use the colors (based on commit duration) for the circles in the interactions view?
- 09edccb: Commits are shown as bars in the charts, but in the interactions tab they're circles. Maybe they should be little colored squares? At least in the "renders" side pane, rather than showing the triangle that fills in, show the colored squares?
- 09edccb: Maybe use black for selected (like we do in the snapshot selector)

#### Elements panel
- cca8b0c: Only scroll selected node if not already visible

#### General
- f984a56: Replace simple cache with LRU cache for charting data

### DIFF
--- a/frontend/Node.js
+++ b/frontend/Node.js
@@ -157,7 +157,9 @@ class Node extends React.Component<PropsType, StateType> {
   ensureInView() {
     var node = this.props.isBottomTagSelected ? this._tail : this._head;
     if (node != null) {
-      if (typeof node.scrollIntoView === 'function') {
+      if (typeof node.scrollIntoViewIfNeeded === 'function') {
+        node.scrollIntoViewIfNeeded();
+      } else if (typeof node.scrollIntoView === 'function') {
         node.scrollIntoView({
           // $FlowFixMe Flow does not realize block:"nearest" is a valid option
           block: 'nearest',
@@ -166,6 +168,13 @@ class Node extends React.Component<PropsType, StateType> {
       }
     }
   }
+
+  _setTailRef = tail => {
+    this._tail = tail;
+  };
+  _setHeadRef = head => {
+    this._head = head;
+  };
 
   render() {
     const {theme} = this.context;
@@ -251,7 +260,7 @@ class Node extends React.Component<PropsType, StateType> {
       }
       return (
         <div
-          ref={h => this._head = h}
+          ref={this._setHeadRef}
           style={sharedHeadStyle}
           {...headEvents}
         >
@@ -293,9 +302,9 @@ class Node extends React.Component<PropsType, StateType> {
       const isCollapsed = content === null || content === undefined;
       return (
         <div style={headWrapperStyle}>
-          <div ref={h => this._head = h} style={sharedHeadStyle} {...headEvents}>
+          <div style={sharedHeadStyle} {...headEvents}>
             &lt;
-            <span style={jsxSingleLineTagStyle}>{name}</span>
+            <span ref={this._setHeadRef} style={jsxSingleLineTagStyle}>{name}</span>
             {node.get('key') &&
               <Props key="key" props={{'key': node.get('key')}} inverted={inverted}/>
             }
@@ -323,7 +332,7 @@ class Node extends React.Component<PropsType, StateType> {
     const closeTag = (
       <Fragment>
         &lt;/
-        <span style={jsxCloseTagStyle}>{name}</span>
+        <span ref={this._setTailRef} style={jsxCloseTagStyle}>{name}</span>
         &gt;
         {selected && ((collapsed && !this.props.isBottomTagSelected) || this.props.isBottomTagSelected) &&
           <span style={dollarRStyle}>&nbsp;== $r</span>
@@ -335,7 +344,7 @@ class Node extends React.Component<PropsType, StateType> {
 
     const jsxOpenTagStyle = jsxTagStyle(inverted && (!isBottomTagSelected || collapsed), nodeType, theme);
     const head = (
-      <div ref={h => this._head = h} style={sharedHeadStyle} {...headEvents}>
+      <div style={sharedHeadStyle} {...headEvents}>
         <span
           onClick={onToggleCollapse}
           style={{
@@ -347,7 +356,7 @@ class Node extends React.Component<PropsType, StateType> {
           {collapsed ? '▶' : '▼'}
         </span>
         &lt;
-        <span style={jsxOpenTagStyle}>{name}</span>
+        <span ref={this._setHeadRef} style={jsxOpenTagStyle}>{name}</span>
         {node.get('key') &&
           <Props key="key" props={{'key': node.get('key')}} inverted={headInverted}/>
         }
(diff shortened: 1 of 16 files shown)


bvaughn:
Okedoke. This PR is ready for review.

It fixes #1104 (so we should probably do another browser release after it's merged) and wraps up the remaining low-hanging-fruit from the umbrella issue. I've split the remaining larger umbrella items off into separate issues [tagged with a new **plugin: profiler** label](https://github.com/facebook/react-devtools/issues?q=is%3Aissue+is%3Aopen+label%3A%22plugin%3A+profiler%22).

gaearon:
Do you mind updating the preview page so I could try it without building?

bvaughn:
Just pushed an update.

bvaughn:
(to https://react-devtools-profiler-prerelease.now.sh/)

gaearon:
Overall I'm not sure I dig black fill as the selected commit. I understand you wanted to make it more prominent, but it confuses me that we use color both to indicate the cost *and* to indicate what's selected, and selecting "hides" the cost hint from us.

Have you explored other options, e.g. a prominent black border while keeping the background as is?

bvaughn:
> Have you explored other options, e.g. a prominent black border while keeping the background as is?

Yes. I probably haven't explored them all, so maybe there's a good one I've not thought of. In general, theming makes this tricky– since it limits our color palette. 😦

The above is some good feedback, but most of it seems unrelated to this PR and it feels a bit hard to respond to since it's just a bullet list of comments. Do you mind filing them as individual issues with the profiler tag?

> there's a funny 1px white border at the right side

I also don't see what you're referring to with this one.

bvaughn:
> what happened in this commit? everything seems grey. were all renders insignificant? did grey things even render? in that case why are they grey? does "total renders" mean total since beginning or profiling? or total in this commit? probably since profiling but this didn't occur to me at first.

This is an interesting question. Any chance you could set a breakpoint and give me a `JSON.stringify` dump of that snapshot data so I could take a peek? A repro case is also fine.

bvaughn:
I'm going to just edit your original comment to add some inline responses. There's no other good way to respond to a big, multi-bullet-point comment like that.

gaearon:
>I also don't see what you're referring to with this one.

Zoom in on the image. You’ll see there’s a strange 1px white strip between black and grey. It looks odd like a glitch.

bvaughn:
I edited your original comment and added some inline responses. (I don't know of a better way to respond to a huge multi-bullet comment like that.) I'm going to push a couple tweaks for the very small items, but I think most things should be filed as follow up issues since this is a confusing format :D

bvaughn:
Pushed a few tweaks to low hanging fruit!

gaearon:
Sorry, I wasn’t tracking individual issues and which changes were associated with which PR so it’s more of a braindump from somebody trying this for the first time. I can later come back and file a single issue for each of these — it just felt excessive if most are small changes or not even necessarily fixable.

bvaughn:
Totally makes sense 😄 I responded inline with some comments and tagged a few as already completed.

gaearon:
>That's fair I guess, although I don't think we've been enforcing that level of scrutiny elsewhere in the DevTools "chrome" 😄 I don't see it as very important. Happy to add a style prop to the toolbar though!

I’m not scrutinizing just to nitpick :-) We didn’t have a tabbed UI before so I didn’t often click on different parts of the toolbar. Now that we have a tabbed interface, I click on it quite a bit and find myself accidentally selecting things on it. Seems like an easy fix so I flagged it. I can refrain from flagging small changes like this, but I selected the toolbar content two times by accident during a five minute testing period. And in general I think we shouldn’t avoid polish because some older parts aren’t polished.

bvaughn:
That was more of a disclaimer that "I don't think we've thought about this level of nuance in other part of DevTools" – I already made the change 😄

gaearon:
>Although I think this is unavoidable if the first commit measures a 0 duration due to it being so fast. Open to suggestions.

We could automatically select the first non-empty commit. We could also tweak the empty state text to suggest selecting a commit from the upper right corner or using arrow keys.

bvaughn:
> We could automatically select the first non-empty commit.

Maybe... I think this might be a bit disorienting though.

> We could also tweak the empty state text to suggest selecting a commit from the upper right corner or using arrow keys.

I like this!

bvaughn:
Actually, thinking more about that empty state– I want to look closer into this. I assume this is the suspense demo, the "Debugger" root?

gaearon:
>I'm still not seeing this, even after zooming in. Sorry

I’m talking about 1px white strip.

0eac7b43-c445-42e5-9cfd-f8ae3a4a3d56

gaearon:
>I'm not entirely sure what's happening here. I tried to repro with the "plain" test harness but couldn't.

I was using it on UFI in Workplace. When you say “couldn’t reproduce” what do you mean precisely? Do you mean that pressing left and right arrows to expand and collapse a single tag with many children always brings its start into view?

gaearon:
>Gets "stuck" how?

The circle in my GIF shows when the mouse is down. You can see that I release it outside the pane. When I enter the pane again, it still thinks the mouse is down (and scrolls through commits) even though I released it. The only way to stop it from thinking I’m holding he mouse down is to hold it down again and release.

This sounds like something that should be possible to solve. I just don’t remember how.

bvaughn:
> I’m talking about 1px white strip.

Oh! 😆 I see it now! Thanks for circling it.

> I was using it on UFI in Workplace. When you say “couldn’t reproduce” what do you mean precisely? Do you mean that pressing left and right arrows to expand and collapse a single tag with many children always brings its start into view?

Yes, using the "deeply nested" component in the example test app, I couldn't repro it. Seemed like I should have been able to.

> The circle in my GIF shows when the mouse is down. You can see that I release it outside the pane. When I enter the pane again, it still thinks the mouse is down (and scrolls through commits) even though I released it. The only way to stop it from thinking I’m holding he mouse down is to hold it down again and release.

Oh! The gif wasn't animating for me earlier so I wasn't seeing that. Now I do.

Hm... I listen for mouse-up on the window, but if you mouse out of the window entirely...hm. File a follow up issue for it. I don't know the fix off the top of my head.

bvaughn:
Okay! I just pushed a commit (and deployed to the preview URL) that improves the UI/UX for commits with a duration of 0. Try again? 😄

bvaughn:
I created a few spin off issues for remaining feedback that I didn't address already. Feel free to create more for anything I missed. 😄

bvaughn:
Anything blocking this from merge, @gaearon ?

Issue #1094, opened 2018-08-14
Highlight the owner tree when a node is focused


gaearon:
When "owner" information is available, filter the breadcrumbs strip to only show owners. When this info is not available, filter it to only show "Composite" components (e.g. no host nodes).

### Original description

I often want to debug why a prop is wrong. Unfortunately in deep trees it's very hard to say which component has passed which prop down (and at which point it got messed up).

We already have a concept of "owners" in React. We could surface that information by highlighting the owner tree whenever you focus on a node. It could be something subtle, like a small difference in opacity, or background color of owners, but I think could really help navigate large trees.

As a bonus I think it would make the "unrelated HOCs" issue less annoying, at least for the use case of debugging "why is this render result wrong" (which for me represents the majority of use cases for DevTools).

bvaughn:
Owner information is tracked in two places:
* It's stored transiently on a React element
* It's stored on Fiber's as `_debugOwner` (in DEV mode only)

The element is never exposed to DevTools directly, so only the `_debugOwner` attribute would be useful– and only in DEV mode. (Not sure off the top of my head if it's available, or in what form, for older majors.)

So what's the proposal for supporting this functionality?

gaearon:
Maybe we can use `_debugOwner` when available, and fall back to showing the parent.

In React 15, owner information is available from `_currentElement._owner` both in DEV and PROD.

bvaughn:
Gotcha.

So we could add a new "owner" property to the "data" object and use it to filter the breadcrumbs list when present. That makes sense.

I guess the behavior when "owner" is null (aka 16+ production build) would just be... an unfiltered list? like we currently have? Seems a bit awkward, but maybe it's okay since DevTools is already awkward in production mode with e.g. minified component names÷

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #1113, opened 2018-08-24
Make Breadcrumb resizable

miklosme:
react-dev-tools-resizable-breadcrumbs

I'm using Breadcrumb a lot, that is my primary view to understand the component tree.

### DIFF
--- a/frontend/Breadcrumb.js
+++ b/frontend/Breadcrumb.js
@@ -123,7 +123,6 @@ const containerStyle = (theme: Theme) => ({
   maxHeight: '80px',
   backgroundColor: theme.base01,
   borderTop: `1px solid ${theme.base03}`,
-  whiteSpace: 'nowrap',
   overflow: 'auto',
 });
 
--- a/frontend/Breadcrumb.js
+++ b/frontend/Breadcrumb.js
@@ -120,10 +120,11 @@ const containerStyle = (theme: Theme) => ({
   listStyle: 'none',
   padding: '0 0.5rem',
   margin: 0,
-  maxHeight: '80px',
   backgroundColor: theme.base01,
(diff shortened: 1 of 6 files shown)


bvaughn:
I appreciate the effort that's gone into this 😁 It looks cool!

But I think we're moving toward adding _fewer_ elements to the breadcrumbs strip (see #1094) which would make this change unnecessary. Ideally, we'll only show "owners" there. Worst case, we'll show only class/function components (no host nodes).

So unless someone else disagrees strongly, I think we should skip this enhancement 😄

miklosme:
This is how a representative breadcrumb looks like at my work:

screen shot 2018-08-27 at 12 58 23

I never use the treeview, when need to understand the component tree, I'm looking at the breadcrumb, when interested in the props, I use the right pane. The recent [change](https://github.com/facebook/react-devtools/blob/master/frontend/Breadcrumb.js#L126) in styling makes my way of working very difficult.

I might missing the point of treeview. I feel like the actual app is already a treeview, when I need to check siblings, I just use the selector tool.

bvaughn:
It's helpful to see an example of what the breadcrumb view looks like for your app 😄

First, all of the "div", "span", etc. elements will soon be removed– which would save a couple of rows. Once #1094 is implemented, we'll also try to remove everything but the "owners". Even for production mode, where we can't remove owners, we can probably add some special case handling to filter out HOCs (e.g. connect*). These changes should make your breadcrumb list much more manageable.

miklosme:
It's a bit vague to me how will it look like, but sounds very cool, looking forward to it.

Till then I'm using my special build for sure 😎

bvaughn:

PR #1130, opened 2018-08-28
add `getPromise` call to access removed `then` method of Deferred

kedromelon:
The `then` method of `Deferred` was removed (see D9308668 internally) -- the `RelayMutationRequest` (extending `Deferred`) `request` now must call `getPromise()` to access the underlying Promise methods

note that this change to fbjs/Deferred has not yet been synced to Github, so this only occurs internally -- error was reported in T33316897

### DIFF
--- a/plugins/Relay/installRelayHook.js
+++ b/plugins/Relay/installRelayHook.js
@@ -75,7 +75,7 @@ function installRelayHook(window: Object) {
     requestNumber: number,
   ) {
     var id = Math.random().toString(16).substr(2);
-    request.then(
+    request.getPromise().then(
       response => {
         emit('relay:success', {
           id: id,


bvaughn:
> note that this change to fbjs/Deferred has not yet been synced to Github

I think you're saying that removing `.then()` hasn't yet been synced, but the `getPromise()` API has already been release (and so is safe to switch to)?

kedromelon:
yup! the `getPromise` method was not added, it is already in the class 

https://github.com/facebook/fbjs/blob/cfd39964ba4b9ce351c314ed512e654ffa9cad26/packages/fbjs/src/core/Deferred.js#L35-L37

bvaughn:
Cool. Just checked and it looks like you're right that the [earliest version of `fbjs` we've pinned to](https://github.com/facebook/react-devtools/blob/6f6f5aec9d3878e0f433d51439be7bee417997e5/yarn.lock#L2806-L2828) does [include `getPromise()`](https://unpkg.com/fbjs@0.5.1/lib/Deferred.js) so... 👍

kedromelon:
thanks @bvaughn!

PR #1138, opened 2018-08-30
Pin Lerna version

Daniel15:
It would be good to convert this to use Yarn workspaces (given the project is already using Yarn), but for now let's at least pin the Lerna version to avoid unintentionally pulling in a new version with a different license.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
   "devDependencies": {
     "chrome-launch": "^1.1.4",
     "firefox-profile": "^1.0.2",
-    "lerna": "^2.8.0",
+    "lerna": "2.8.0",
     "web-ext": "^1.10.1"
   }
 }


Daniel15:
My mistake, I think you're right. I forgot about that - it's been a while since I've touched JS code. It looks like the Travis build isn't running using the `--frozen-lockfile` option, so it didn't catch the issue. I'll run `yarn install` locally and commit the updated lockfile.

vjeux:
The commit was reverted, should we still go ahead with this? I'd be happy to merge it in. https://github.com/lerna/lerna/pull/1633

nickmccurdy:
I believe you'd need to wait for the next release and pint that as a minimum version using `^`.

Daniel15:
Yeah this isn't as important now that the Lerna license change has been reverted.

PR #1139, opened 2018-08-30
Fix false positive feature-detection check

bvaughn:
Addresses the issue brought up with [this comment](https://github.com/facebook/react-devtools/pull/1069#issuecomment-417166634) by improving our feature-detection check to avoid false positive from 16.4

The `actualDuration` field was added by facebook/react/pull/12910 and was released in 16.4.

The `treeBaseDuration` field was renamed with facebook/react/pull/13156 in order to allow us to distinguish between the initial profiling release and a more stable one. I intended to switch DevTools to use this field but apparently forgot to. ☹️


bvaughn:
Merging without a review because this change is tiny and uncontroversial.

PR #1162, opened 2018-09-19
add `color by self time` option to ProfilerSettings

bennyhobart:
Hi,

I'd find it extremely useful to color components based on the time they took to run their render method, as opposed to how long the component and all its descendants took to render.

Here's a video showing off the feature https://vimeo.com/290768458

Thanks!

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -152,7 +152,6 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     var actualDuration = null;
     var actualStartTime = null;
     var treeBaseDuration = null;
-    var selfBaseDuration = null;
 
     var resolvedType = type;
     if (typeof type === 'object' && type !== null) {
@@ -317,7 +316,6 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
       actualDuration = fiber.actualDuration;
       actualStartTime = fiber.actualStartTime;
       treeBaseDuration = fiber.treeBaseDuration;
-      selfBaseDuration = fiber.selfBaseDuration;
     }
 
     // $FlowFixMe
@@ -340,7 +338,6 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
       actualDuration,
       actualStartTime,
       treeBaseDuration,
-      selfBaseDuration,
     };
   }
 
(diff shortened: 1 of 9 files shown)


bvaughn:
Firstly, this is really exciting. Thanks for contributing! 😁 I'm excited to play around with this some and see if it makes spotting perf problems easier (I think it will).

I do think there's a bit of an oddity about the current implementation though, captured in this GIF:
kapture 2018-09-19 at 14 49 13

I think we need to ignore self base time for native nodes if they've been filtered out of DevTools, else we risk showing a confusing commit like the GIF above shows. I think this– coupled with the potential for small discrepancies between `actualDuration` and `selfBaseDuration` might be reason enough to go with the approach of JIT determining the max "self time" via `committedNodes.forEach` rather than using the `_onMountOrUpdate` approach. (I suppose we _could_ just collect _two_ `_maxSelfBaseDuration` values– one included host nodes and one not, but that feels kind of janky.)

plugins/Profiler/views/SnapshotFlamegraph.js
```diff
@@ -257,16 +264,18 @@ class ListItem extends PureComponent<any, void> {
...
 
           return (
             <ChartNode
-              color={didRender ? getGradientColor(actualDuration / maxDuration) : didNotRender}
+              color={color}
               height={barHeight}
               isDimmed={index < focusedNodeIndex}
               key={id}
-              label={didRender ? `${name} (${actualDuration.toFixed(1)}ms)` : name}
+              label={didRender ? `${name} (${colorBySelfTime ? selfBaseDuration.toFixed(1) : actualDuration.toFixed(1)}ms)` : name}
```
bvaughn: I think I've changed my mind on the label after playing with it for a few minutes. For the self-time case, I think we should actually show "self time / tree time". Otherwise it feels confusing to me to see smaller values above larger values.

bvaughn: Showing `selfBaseDuration` / `actualDuration` makes the minor discrepancies I mentioned too obvious. I think this confirms that we should just calculate a "self duration" value.

bvaughn: I'm going to play with this a little locally and get back to you!

bennyhobart:
This reads really nicely,

I think it's better labeling both durations and having the color default to how much time is spent in the node, and there's less state being passed around everywhere which simplifies things.

I don't think there's much I'd change about it!

backend/attachRendererFiber.js
```diff
@@ -152,6 +152,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
     var actualDuration = null;
     var actualStartTime = null;
     var treeBaseDuration = null;
+    var selfBaseDuration = null;
```
bennyhobart: perhaps remove this as it's no longer being used, but it might be useful in the future?

bvaughn:
Thanks! I'm going to take one more pass at perf cleanup then it'll be ready for a final (code) review.

bvaughn:
Okay. Let's move forward with this. I think the UX is a big improvement. We can iterate on any technical issues or other concerns that come up.

Thanks! 😁

bvaughn:
Released with 3.4.0

PR #1165, opened 2018-09-23
Implement "Highlight Extra Props" option

wojtekmaj:
As React applications grow and change, it's common for developers to discover some props which are not used anymore. It contributes to a messy code and can potentially cause components to render more times than it is necessary.

With static code analysis like ESLint, it is very easy to find unused props within the component itself, and equally easy to find props which are not included in propTypes. However, it's next to impossible to find extra props that are passed to a component without it needing them.

This PR adds an option that highlights props in yellow in case:

1. propTypes for a given component were defined
2. among these propTypes, a given prop is not defined

Here it is in action:
obraz


test/example/target.js
```diff
@@ -479,6 +480,7 @@ class Wrap extends React.Component {
           <PropTester {...complexProps}/>
           <PropTester {...uninspectableProps}/>
           <PropTester massiveMap={massiveMap}/>
+          <PropTesterWithPropTypes validProp="test" invalidProp="test" propWithoutPropTypes="test" />
```
wojtekmaj: This is my concern here - it would be handy to have an invalid prop here for testing purposes, and it causes a propType error (duh). What do you think? Should I remove it for now?

bvaughn:

PR #1170, opened 2018-09-26
Adds guard for HostComponent name

jakubsadura:
name can be null, replace should be guarded

screen shot 2018-09-26 at 14 57 25

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -213,7 +213,9 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
 
         // TODO (bvaughn) we plan to remove this prefix anyway.
         // We can cut this special case out when it's gone.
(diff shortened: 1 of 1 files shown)


jakubsadura:
I have just signed Contributor License Agreement

bvaughn:
In what case would a host component name be null? Host components are just identified by strings, so a null value would fail to render anything?

jakubsadura:
Good question, I have an app with few React roots, I'm not able to reproduce the issue right now. As far as I remember I just tried to open React devtool quite soon after opening the app in Electron and came across error being thrown from `name.replace` - I quickly opened this PR and moved forward. I remember that commenting out `name.replace` helped to run the React devtools.

bvaughn:
I'd prefer to hold off on merging this until I better understand the cause. If you see it again, I'd love if you could share more info (or better yet a repro case) with me.

bvaughn:

PR #1171, opened 2018-09-26
AsyncMode -> ConcurrentMode

trueadm:
Follow up to https://github.com/facebook/react/pull/13732. Renames `AsyncMode` to `ConcurrentMode`.

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -73,6 +73,7 @@ function getInternalReactConstants(version) {
   ReactSymbols = {
     CONCURRENT_MODE_NUMBER: 0xeacf,
     CONCURRENT_MODE_SYMBOL_STRING: 'Symbol(react.concurrent_mode)',
+    DEPRECATED_ASYNC_MODE_SYMBOL_STRING: 'Symbol(react.async_mode)',
     CONTEXT_CONSUMER_NUMBER: 0xeace,
     CONTEXT_CONSUMER_SYMBOL_STRING: 'Symbol(react.context)',
     CONTEXT_PROVIDER_NUMBER: 0xeacd,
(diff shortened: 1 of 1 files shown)


trueadm:
@bvaughn I thought that was the point here? It's an unstable API so can break in any version, right? I'm happy to add support in on the devtools, but the PR to rename AsyncMode has already landed in React – did you want me to revert it?

bvaughn:
> @bvaughn I thought that was the point here? It's an unstable API so can break in any version, right? I'm happy to add support in on the devtools, but the PR to rename AsyncMode has already landed in React – did you want me to revert it?

Well, yes– the API can change in that we can e.g. rename the export with a new release of `react` _but_ we don't want a DevTools update to break compatibility with older releases. (DevTools tries to be compatible with all previous React versions.)

So in this case, we'll need the 16.x DevTools adapter to handle both `AsyncMode` and `ConcurrentMode` types. 😄

backend/attachRendererFiber.js
```diff
@@ -71,8 +71,9 @@ function getInternalReactConstants(version) {
     };
   }
   ReactSymbols = {
-    ASYNC_MODE_NUMBER: 0xeacf,
-    ASYNC_MODE_SYMBOL_STRING: 'Symbol(react.async_mode)',
+    CONCURRENT_MODE_NUMBER: 0xeacf,
+    CONCURRENT_MODE_SYMBOL_STRING: 'Symbol(react.concurrent_mode)',
+    DEPRECATED_ASYNC_MODE_SYMBOL_STRING: 'Symbol(react.async_mode)',
```
bvaughn: This isn't sufficient. We need the number and string.

trueadm: The number is the same though?

bvaughn: Haha, d'oh! Ok sorry.

probablyup:
It'd be cool if you reviewed my PR from 2 weeks ago that has its corresponding change already in react core and released. Thanks.

PR #1173, opened 2018-09-28
Support new pure HOC component

bvaughn:
Follow up to PR https://github.com/facebook/react/pull/13748

The following code:
```js
function Counter({count}) {
  return <small>{count}</small>;
}
const PureCounterA = React.pure(({count}) => <small>{count}</small>);
const PureCounterB = React.pure(Counter);
const PureCounterC = React.pure(Counter);
PureCounterC.displayName = 'PureCounter';
```

Will now render the following in DevTools:
screen shot 2018-09-28 at 1 28 13 pm


bvaughn:
I'll revert this or iterate on it later if there are concerns.

PR #1178, opened 2018-10-07
Pull memoizedInteractions from DataType instead of Fiber

marvinhagemeister:
I'm currently integrating the profiler into preactjs as an experimental feature and noticed, that the profiler reads from the `memoizedInteractions` property directly from a `fiber` instead of from `event.data` like it is done for all other properties.

The integration is nearly done and this is the only inconsistency I've come across. I'm really impressed by the profiler and how easy it was to add support for it. It's really amazing 💯 👍 

This PR fixes that by adding a `data` property to the `rootCommitted` event similar to how all other events work.

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -158,6 +158,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     var actualDuration = null;
     var actualStartTime = null;
     var treeBaseDuration = null;
+    var memoizedInteractions = null;
 
     var resolvedType = type;
     if (typeof type === 'object' && type !== null) {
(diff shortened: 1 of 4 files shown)


backend/attachRendererFiber.js
```diff
@@ -181,6 +182,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
           if (context && Object.keys(context).length === 0) {
             context = null;
           }
+          memoizedInteractions = publicInstance.memoizedInteractions;
```
bvaughn: We should be setting this for the `HostRoot` tag type actually. I can make this change.

bvaughn:
Merged via f109374

Thanks!

marvinhagemeister:
@bvaughn You're awesome! 🎉 Would've done the requested changes myself as the first thing tomorrow morning, but I'm happy that it's merged. Thanks again for looking at this PR 👍💯

bvaughn:
Thank you!

Just released as 3.4.2

Issue #1076, opened 2018-08-06
Provide a way to filter types of components from the Elements tree view


bvaughn:
Based on some user recent feedback I gathered [via Twitter](https://twitter.com/brian_d_vaughn/status/1025776259056336897), one common request is some sort of method of filtering elements from the tree to make it less noisy.

Some people have mentioned using a string/regex to filter, and that could work for hiding e.g. HOCs– but I'm picturing a panel that shows all of the types of elements (e.g. host nodes, fragments, context consumer/provider, etc) with checkboxes next to them to indicate whether they should be visible or not.

These settings should be persisted between page reloads through the use of a mechanism like `utils/storage` added in PR #1069.

markerikson:
Yep, definitely!  Prior existing issues:  #503, #604, #864, #997, and #1001 .

diegohaz:
I think there should be a way to hide those components from within the code. For example, setting a static property on the component type:
```js
Component.devTools = false; // example
```

Some good points of this approach I can see:

- It's very explicit!

- This configuration would be consistent within a team working on the same project.

- Library authors would be able to say which components are "useless" for the devtools (but the user should have a way to show them, perhaps combining it with the checkboxes solution, where the components hidden with the static property would be checked by default).

Well, just some thoughts. Thank you for the amazing work, guys.

bvaughn:
Proof of concept (hack) pushed to https://github.com/bvaughn/react-devtools/tree/hidden-types

![hide-elements-by-type](https://user-images.githubusercontent.com/29597/48806891-acb9a380-ecd0-11e8-9ad3-31c09b15a259.gif)

Has a couple of pain points to work out if we decide it's worth releasing:
* Keyboard navigation is a bit "broken" for expanding/collapsing nodes since children might be "hidden.
* Hiding native nodes looks a little weird at the moment when you drill down to the lowest function/class component, because `Node` expects there to be children.

dimensi:
Half a year has passed, is there news? This is the most important feature I've seen in devtools. I am very much tired of this information noise when working with react devtools. When you see tons of native elements

bvaughn:
If there was an update on this, we would have updated the GitHub issue. 😄 We're a small team and we maintain a large surface area. Things often don't move as fast as we'd like.

That being said, you may be interested in taking the new (still experimental) DevTools version for a test drive. It doesn't show native elements at all, so it may reduce some of the "noise" you're talking about. You can find early release builds in the top-right corner of the demo page:
[...]

PR #1187, opened 2018-10-15
[WIP] Support hiding components in devtools

Gvozd:
Add two options for hiding components
1) For components with `[Symbol.for('react.devtools.hide')] = true` - rebase of PR #997 
2) If component define `displayName` - I think that most popular case for HOC

I'm waiting for feedbacks
It is time to move forward on this issue after long discussions.

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -153,6 +153,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     var nodeType = null;
     var name = null;
     var text = null;
+    var hideSymbol = null;
 
     // Profiler data
     var actualDuration = null;
@@ -176,11 +177,15 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
         publicInstance = fiber.stateNode;
         props = fiber.memoizedProps;
         state = fiber.memoizedState;
+        hideSymbol = typeof Symbol === 'function' && fiber.type[Symbol.for('react.devtools.hide')];
         if (publicInstance != null) {
           context = publicInstance.context;
           if (context && Object.keys(context).length === 0) {
             context = null;
           }
+        }
+        if (name === 'HigherOrderComponent') {
+          console.log('fiber backend', fiber);
         }
         const inst = publicInstance;
         if (inst) {
@@ -360,6 +365,8 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
       actualDuration,
       actualStartTime,
       treeBaseDuration,
+
+      hideSymbol,
     };
   }
 
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -154,6 +154,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     var name = null;
     var text = null;
     var hideSymbol = null;
+    var hideDisplayNamed = null;
 
     // Profiler data
     var actualDuration = null;
@@ -178,6 +179,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
         props = fiber.memoizedProps;
         state = fiber.memoizedState;
         hideSymbol = typeof Symbol === 'function' && fiber.type[Symbol.for('react.devtools.hide')];
+        hideDisplayNamed = fiber.type.hasOwnProperty('displayName');
         if (publicInstance != null) {
           context = publicInstance.context;
           if (context && Object.keys(context).length === 0) {
@@ -358,6 +360,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
       updater,
       publicInstance,
       hideSymbol,
+      hideDisplayNamed,
 
       // Profiler data
       actualDuration,
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -179,8 +179,8 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     var nodeType = null;
     var name = null;
     var text = null;
-    var hideSymbol = null;
-    var hideDisplayNamed = null;
+    var needHideBySymbol = null;
+    var needHideByParensInName = null;
 
     // Profiler data
     var actualDuration = null;
@@ -205,8 +205,8 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
         publicInstance = fiber.stateNode;
         props = fiber.memoizedProps;
         state = fiber.memoizedState;
-        hideSymbol = typeof Symbol === 'function' && fiber.type[Symbol.for('react.devtools.hide')];
-        hideDisplayNamed = fiber.type.hasOwnProperty('displayName');
+        needHideBySymbol = typeof Symbol === 'function' && fiber.type[Symbol.for('react.devtools.hide')];
+        needHideByParensInName = /\(.*\)/.test(name);
         if (publicInstance != null) {
           context = publicInstance.context;
           if (context && Object.keys(context).length === 0) {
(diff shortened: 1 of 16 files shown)


Gvozd:
Call in the discussion of interested persons
@gaearon @sebmarkbage @Kovensky @zackargyle @kyleshevlin @jquense @jaredly @shriah @bvaughn

jaredly:
Looks good to me! I'm no longer a committer on this project, so I'll have to let someone else merge

backend/attachRendererFiber.js
```diff
@@ -176,6 +178,8 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
         publicInstance = fiber.stateNode;
         props = fiber.memoizedProps;
         state = fiber.memoizedState;
+        hideSymbol = typeof Symbol === 'function' && fiber.type[Symbol.for('react.devtools.hide')];
+        hideDisplayNamed = fiber.type.hasOwnProperty('displayName');
```
bvaughn: Sorry, I'm jet lagged at the moment so maybe this is a stupid question, but why is the "displayName" check here? I don't get how it's related.

frontend/Breadcrumb.js
```diff
@@ -163,7 +163,7 @@ function getBreadcrumbPath(store: Store): BreadcrumbPath {
 }
 
 module.exports = decorate({
-  listeners: () => ['breadcrumbHead', 'selected'],
+  listeners: () => ['breadcrumbHead', 'selected', 'hideSymbol', 'hideDisplayNamed'],
```
bvaughn: Why is `Breadcrumb` listening for `hideDisplayNamed` and `hideSymbol` when it doesn't use them?

frontend/PreferencesPanel.js
```diff
@@ -118,6 +122,26 @@ class PreferencesPanel extends React.Component<Props, State> {
...
+            </label>
+          </div>
+          <div>
+            <label>
+              <input
+                type="checkbox"
+                checked={hideDisplayNamed}
+                onChange={this._changeHideDisplayNamed}
+              />
+              Hide components with custom <code>displayName</code> property
```
bvaughn: I don't understand why you'd want this particular functionality.

backend/attachRendererFiber.js
```diff
@@ -176,6 +178,8 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
         publicInstance = fiber.stateNode;
         props = fiber.memoizedProps;
         state = fiber.memoizedState;
+        hideSymbol = typeof Symbol === 'function' && fiber.type[Symbol.for('react.devtools.hide')];
```
bvaughn: nit: `react-devtools.hide`?

bvaughn:
I'm not sure how I feel about the hidden symbol. I don't like the `hideDisplayNamed` logic on initial read though. It seems unintuitive to me. Maybe you can explain your thought process?

frontend/PreferencesPanel.js
Gvozd: Many HOC's creates wrapper components with custom displayName like `SomeWrapper(OriginalComponent)`
In contrast, regular components usually do not use dispalyName

I thought that such a simple rule is suitable for a draft, and better than nothing
This is the minimum that I would like to get in order to hide noisy HOC's
I can add/replace with other simple HOC recognition, if they work better.

asvetliakov: > Many HOC's creates wrapper components with custom displayName
>In contrast, regular components usually do not use dispalyName

There are many that are kind of like "presentation HOC" and assign displayName through babel plugin, i.e styled-components, emotion, jss, etc:
```jsx
const MyButton = styled.button`
   color: tomato;
`
// MyButton will have displayName: "MyButton" crafted by babel plugin
```
Although they are technically HOC i'd not want to hide them.

bvaughn: I think the `displayName` attribute is too likely to cause false positives, as @asvetliakov points out, due to popular plug-ins like [`babel-plugin-transform-react-display-name`](https://www.npmjs.com/package/babel-plugin-transform-react-display-name). It also feels too arbitrary to me.

bvaughn:
I believe I would also be more of a fan of that approach, as it would be more consistent with filtering other categories of components (as described in #1076)

FredyC:
This was so good looking until it wasn't. Would you please remove the second approach with `displayName`? It's too much opinionated and not that useful. Even though it's opt-in, it would be just confusing.

Gvozd:
Sorry for the break
Replaced `'Hide components with custom 'displayName' property` to `Hide HOC's with parens in name`
Fixed some bug with breadcrumbs, and selection
Corrected UI
image

In next plans:
- additional dimming mode
- optimize performance

FredyC:
@Gvozd thanks, much better.

I am wondering, won't this symbol approach be somewhat problematic in TypeScript or possibly even Flow? Generally attaching foreign properties to objects is rather painful with static typing. Any ideas how it could be solved except using `// @ts-ignore` above the assignment line?

Kovensky:
I wasn't thinking of attaching it to arbitrary functions but rather to functions/classes you generate yourself as part of the HOC factory function. `static [Symbol.for('react-devtools.hide')] = true`.

FredyC:
Well, for an example I would like to hide all `React.createContext` so I suppose I would have to do something like this? 

```ts
const context = React.createContext()
// @ts-ignore
context.Consumer[hideSymbol] = true
```

And also some stateless components are useless to be visible in the tree, so that's about attaching a symbol to existing function.

Gvozd:
1) I fixed the performance of many nested components
2) For custom static symbol/ordinary property
If they are approved by maintainers - need fix ts/flow lib definition for react
In this case, setting this property should work as well as `displayName`
It is also necessary to use an ordinary property, in order to properly support the flow
Offer `hideInDevtools` as the name of a static property

Gvozd:
> TLDR: There are no plans to remove classes from React.
The old HOCs will not disappear anywhere for a long time
Classes do not go anywhere, so the hooks will live for a long time in parallel with the existing APIs, classes, and therefore HOC/render props

It also takes time
- stabilize RFC
- while the most popular libraries will add hoke API
In addition, because hooks do not work with class components, and no one is going to refuse classes
- when (if) existing projects will be transferred to hooks, and exclusively functional components

This is the case if *all* developers accept hooks as more convenient, and decide to switch to them in full
Until then (an optimistic assessment for the next year or two), the HOC will be a frequent guest in many already existing projects

So I think that to hide the HOCis still necessary and useful.

Gvozd:
Upd: support dimming
At next - I fix bug with higlighting elements in page, when Preferences panel is opened
image

Gvozd:
I think I need to leave only one option of hiding/ darkening, for the HOC with parentheses in the name
It is also worth moving these options to SettingsPane, as in #503, only in the form of selectbox, so as not to take up much space
What you think?

Inve1951:
>I think I need to leave only one option of hiding/dimming, for the HOC with parentheses in the name

I think dimming should be the default. A toggle to hide instead would be usefull.
So basically, off = dimmed, on = hidden.

>It is also worth moving these options to SettingsPane, as in #503, only in the form of selectbox, so as not to take up much space

Good idea.

bvaughn:
I don't think the dimming is very useful. It's very subtle and the tree is just as cluttered as before. (I think the primary purpose of hiding components is that the element tree is too crowded.)

I also think it's confusing to have a checkbox that says, e.g. "Hide HOCs..." when they won't actually be hidden unless you select _one of three_ radio options above. I suppose I could imagine a UX where we have a global toggle on/off, but I don't think it's necessary.

I realize this PR has been pretty long running, and I feel bad about the inconvenience. If you'd like, I could pick up where this PR leaves off and make the additions I had planned for #1076 originally?

Gvozd:
@bvaughn, OK.

Inve1951:
>I don't think the dimming is very useful. It's very subtle and the tree is just as cluttered as before. (I think the primary purpose of hiding components is that the element tree is too crowded.)

I personally appreciate a dimming option for HOCs, as they are part of the virtual DOM and I'd like to view them as such, but at the same time want them to be visually distinguishable/faded since they are HOCs / less impactful than regular components.

>I also think it's confusing to have a checkbox that says, e.g. "Hide HOCs..." when they won't actually be hidden unless you select one of three radio options above. I suppose I could imagine a UX where we have a global toggle on/off, but I don't think it's necessary.

We just talked about this and I agree that the radio buttons are inconvenient. My suggestion was to make a checkbox/toggle for hiding. If dimming gets included in this feature then it could be the off-state of said toggle.

@bvaughn @Gvozd

PR #1190, opened 2018-10-22
Added new tag values for 16.6+

bvaughn:
* Reorganized tag constants to make it easier to identify the differences between the various 16.x targets.
* Removed the never-released "lazy" tags.
* Also added a TODO for supporting some upcoming tag-types in 16.6.

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -29,16 +29,13 @@ function getInternalReactConstants(version) {
   if (semver.gte(version, '16.6.0-beta.0')) {
     ReactTypeOfWork = {
       ClassComponent: 1,
-      ClassComponentLazy: -1, // Removed
       ContextConsumer: 9,
       ContextProvider: 10,
       CoroutineComponent: -1, // Removed
       CoroutineHandlerPhase: -1, // Removed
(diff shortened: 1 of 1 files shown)


backend/attachRendererFiber.js
```diff
@@ -23,51 +23,86 @@ function getInternalReactConstants(version) {
...
+      Mode: 8,
+      Profiler: 12,
+      SimpleMemoComponent: 15,
+      SuspenseComponent: 13,
+      YieldComponent: -1, // Removed
+    };
+  } else if (semver.gte(version, '16.4.3-alpha')) {
+    ReactTypeOfWork = {
       ClassComponent: 2,
       ClassComponentLazy: 3,
```
sebmarkbage: We didn't actually release these in public releases since they were behind the suspense flag. It seems like we can drop support for this tag now.  I don't think RN is using lazy yet neither and if when they start, we'll be on 16.6 anyway. We don't support alphas continuously neither.

bvaughn: Ah! I didn't know that the *Lazy tags never went out in a non-alpha. That's great, I'll remove them.

PR #1191, opened 2018-10-23
Fix showing component source when React Hot Loader is enabled

macku:
There is a problem with a `Show <Component> Source` feature when you are running React with [React Hot Loader](https://github.com/gaearon/react-hot-loader). After selecting the option from a context menu, you will be navigated to a React Hot Loader `ProxyFacade` component instead of the source of a selected component.

## How to reproduce?
1. Go to this page: https://jnwlwrm1qw.codesandbox.io/
2. Open React Devtools
3. Select `App` component on the Elements tab
4. From the context menu select `Show App Source`

## Current behavior
Devtools are activating Source tab and highlighting `ProxyFacade` that is coming from React Hot Loader.

## Expected behavior
Devtools should go to the Source tab and highlight the source of `App` component.

## Demo
rhl-bug

### DIFF
--- a/shells/webextension/src/panel.js
+++ b/shells/webextension/src/panel.js
@@ -77,7 +77,11 @@ var config: Props = {
         window.${globalPathToType}.prototype &&
         window.${globalPathToType}.prototype.isReactComponent
       ) {
-        inspect(window.${globalPathToInst}.render);
+        if (window.${globalPathToInst}.REACT_HOT_LOADER_RENDERED_GENERATION) {
+          inspect(Object.getPrototypeOf(window.${globalPathToType}));
+        } else {
+          inspect(window.${globalPathToInst}.render);
+        }
       } else {
         inspect(window.${globalPathToType});
       }


shells/webextension/src/panel.js
```diff
@@ -77,7 +77,11 @@ var config: Props = {
         window.${globalPathToType}.prototype &&
         window.${globalPathToType}.prototype.isReactComponent
       ) {
-        inspect(window.${globalPathToInst}.render);
+        if (window.${globalPathToInst}.REACT_HOT_LOADER_RENDERED_GENERATION) {
```
macku: The `REACT_HOT_LOADER_RENDERED_GENERATION` property is assigned by [React Hot Loader](https://github.com/gaearon/react-hot-loader/blob/acb748b4e410f3b72cd1079f741223f3606f2f23/src/reconciler/proxyAdapter.js#L16). Thanks to this property we can now that we are running RHL.

theKashey: It would not work for SFC, as long they are not in the prototype.
The only way to access "original" component is to use `window.${globalPathToInst}.__reactstandin__getCurrent`([link](https://github.com/gaearon/react-hot-loader/blob/master/src/proxy/createClassProxy.js#L306)), but this is something react-dev-tools should not rely on.

theKashey:
Would be fixed in Rect 4.5.0 via option `pureRender` (disabled by default, yet)

theKashey:
Probably you can close this issue now.
React-Hot-Loader 4.5.4 is going live tonight.

Issue #844, opened 2017-07-13
Persist "Highlight updates"-value when dev tools reloads


Fralleee:
This is useful when using 3rd party login, which forces react dev tools to reload and thus disabling highlight updates, or just simply when you are in "optimization"-mode and want the "highlight updates"-option enabled for a longer period of time.

sumnerevans:
@Fralleee, I can look into this. I've been looking through the codebase trying to figure out how to tackle this. Are there any mechanisms in place currently for
1. storing state in a way that is persistent across reloads; and
2. restoring state after a reload?

falinsky:
@gaearon @bvaughn Is there any chance to get a fix for this issue to be reviewed? May be is there any schedule regarding this? Thanks.

bvaughn:
This change will go out with the next DevTools release (probably sometime this week). Thanks @falinsky!

falinsky:
@bvaughn Thats a great news. Thank you.

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #1201, opened 2018-10-29
Persist "Highlight updates"-value when dev tools reloads (#844)

falinsky:
Implements #844

### DIFF
--- a/frontend/Store.js
+++ b/frontend/Store.js
@@ -11,7 +11,7 @@
 'use strict';
 
 var {EventEmitter} = require('events');
-var {Map, Set, List} = require('immutable');
+var {Map, Set, List, Record} = require('immutable');
 var assign = require('object-assign');
 var { copy } = require('clipboard-js');
 var nodeMatchesText = require('./nodeMatchesText');
@@ -20,6 +20,9 @@ var serializePropsForCopy = require('../utils/serializePropsForCopy');
 var invariant = require('./invariant');
 var SearchUtils = require('./SearchUtils');
 var ThemeStore = require('./Themes/Store');
+const {get, set} = require('../utils/storage');
+
+const LOCAL_STORAGE_TRACE_UPDATES_KEY = 'traceUpdates';
 
 import type Bridge from '../agent/Bridge';
 import type {ControlState, DOMEvent, ElementID, Theme} from './types';
@@ -569,6 +572,7 @@ class Store extends EventEmitter {
     this.emit('traceupdatesstatechange');
     invariant(state.toJS, 'state.toJS should exist');
     this._bridge.send('traceupdatesstatechange', state.toJS());
+    set(LOCAL_STORAGE_TRACE_UPDATES_KEY, state.enabled);
   }
 
   changeColorizer(state: ControlState) {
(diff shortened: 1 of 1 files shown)


frontend/Store.js
```diff
@@ -600,6 +604,13 @@ class Store extends EventEmitter {
       clearInterval(requestInt);
       this.capabilities = assign(this.capabilities, capabilities);
       this.emit('connected');
+
+      const traceUpdates = get(SESSION_STORAGE_TRACE_UPDATES_ENABLED_KEY, false);
```
bvaughn: Based on the Flow type:
```js
traceupdatesState: ?ControlState;
```
Shouldn't the default value either be `new Record({enabled: false})` or just `null`?

utils/storage.js
```diff
@@ -10,22 +10,34 @@
...
+    } catch (error) {
+      console.error('Could not write to storage.', error);
+    }
+    return false;
+  };
+
 }
+
+export const local = new WebStorage(localStorage);
+export const session = new WebStorage(sessionStorage);
```
bvaughn: Why session storage for this value vs local storage?

frontend/Store.js
```diff
@@ -20,6 +20,9 @@ var serializePropsForCopy = require('../utils/serializePropsForCopy');
 var invariant = require('./invariant');
 var SearchUtils = require('./SearchUtils');
 var ThemeStore = require('./Themes/Store');
+const {get, set} = require('../utils/storage').session;
+
+const SESSION_STORAGE_TRACE_UPDATES_ENABLED_KEY = 'traceUpdates:enabled';
```
bvaughn: nit: other keys are just named after the preference, e.g. `themeName` or the location+preference `profiler:commitThreshold`. I think ":enabled" is implied. 😄

utils/storage.js
falinsky: @bvaughn to have separate state for this value on different pages/browser tabs

frontend/Store.js
falinsky: @bvaughn OK, thanks, will change that

frontend/Store.js
falinsky: @bvaughn not sure if we need to mess up `boolean` data which comes from storage with an instance of `Record`.  So will change to be `null` here.

falinsky:
@bvaughn Changes provided, please review them.

utils/storage.js
bvaughn: Right, sorry. I understand that would be the result, but why is that the goal? I think that would be confusing if your DevTools settings panel had a value that was persisted differently for different pages. I think that would be a confusing UX.

falinsky: From my perspective it would be confusing to have one value shared between different applications for example, or different browser tabs. Also I believe there could be situations when the same application is opened in 2 different tabs and one has highlight updates enabled and other - disabled. In my mind that shouldn't be the same way as it's done for saving a theme in local storage which is shared between different instances.

But if that is not acceptable or indeed confusing - I can change implementation to use the local storage instead. Also if so - should I revert back changes regarding to introducing a wrapper for session storage and related refactoring as well?

bvaughn: > But if that is not acceptable or indeed confusing - I can change implementation to use the local storage instead. Also if so - should I revert back changes regarding to introducing a wrapper for session storage and related refactoring as well?

Yeah, I think so– since I don't think we want to use session storage.

falinsky: @bvaughn OK, thanks.

bvaughn:
Please don't ping multiple times for a review. We're spread pretty thin and will do our best to review changes as we get the chance.

falinsky:
@bvaughn Got it. Thank you.

PR #1203, opened 2018-10-29
Persist checkboxes values when open/close settings panel multiple times (#1202)

### LINKED ISSUE #1202: Persist checkboxes values when open/close settings panel multiple times
When toggling **Settings** (Customize React DevTools) panel multiple times - "Highlight Updates" and "Highlight Search" checkboxes become unchecked.

falinsky:
Fixes #1202

### DIFF
--- a/frontend/SettingsCheckbox.js
+++ b/frontend/SettingsCheckbox.js
@@ -45,8 +45,8 @@ class SettingsCheckbox extends React.Component<Props, State> {
   }
 
   componentDidMount(): void {
-    if (!this.props.state !== this._defaultState) {
-      this.props.onChange(this._defaultState);
+    if (this.props.state && (this.props.state !== this._defaultState)) {
+      this.props.onChange(this.props.state);
     }
   }
 


frontend/SettingsCheckbox.js
```diff
@@ -45,8 +45,8 @@ class SettingsCheckbox extends React.Component<Props, State> {
   }
 
   componentDidMount(): void {
-    if (!this.props.state !== this._defaultState) {
-      this.props.onChange(this._defaultState);
+    if (this.props.state && (this.props.state !== this._defaultState)) {
+      this.props.onChange(this.props.state);
```
bvaughn: This feels weird to me. If both `onChange` and `state` are coming from `props` (aka being passed in) then it doesn't seem right that our React component would need to tell the caller that props have "changed".

I'm going to think about this a little bit.

bvaughn:
Thanks a ton for diving in and submitting a PR for this!

I'm going to use this opportunity to cut a bit deeper and refactor this component, because it seems unnecessarily complicated to me. I've filed #1245 as an alternate fix to #1202. If you're interested in testing it to confirm on your end– that'd be great! 😄

falinsky:
@bvaughn Thanks for the information!

Issue #1193, opened 2018-10-24
memo()-wrapped components don't have names


selbekk:
CodeSandbox: https://codesandbox.io/s/1z1p679677
Reproduction: https://1z1p679677.codesandbox.io/

Open the React DevTools and you'll see something like this:

![image](https://user-images.githubusercontent.com/1307267/47413729-60318580-d76f-11e8-9df3-4bcf74e9cb7b.png)

I'd love to contribute, but I'm not sure where to begin. If you could please guide me in the right direction, I'll be more than happy to create a PR

chpio:
yeah, somehow related: the name is shown now, but no props

Finesse:
It's a progress though

Issue #1189, opened 2018-10-19
☂️ Suspense umbrella task


bvaughn:
## Misc
- [ ] Keyboard navigation: Arrow key navigation for selected element is a bit broken. You can use down/right arrow to drill down into a `Suspense` component, but the up/left arrows won't drill back up. (It gets stuck at the component beneath `Suspense`.)
![arrowkey](https://user-images.githubusercontent.com/29597/47247233-4f6fc100-d3b7-11e8-83a1-2601787e6867.gif)

## Hiding timed-out children

facebook/react/pull/13823 changed the behavior of timed-out suspense trees. In order to preserve state (of both React components and host components) the timed-out tree now stays mounted, but gets hidden (e.g. "display: none").

For the GIF below, I used `create-react-app` with the following changes: https://gist.github.com/bvaughn/aab2a855c5fa1daa9ca55b1fdfac20a5

- [ ] If the suspended subtree is expanded in the DevTools tree view, it gets auto-collapsed when it's hidden.
![collapse](https://user-images.githubusercontent.com/29597/47247260-6dd5bc80-d3b7-11e8-8a21-f179b290b537.gif)

- [ ] The Profiler's render counter gets reset for all components in the hidden subtree. This makes it look like it's a new tree and might be confusing to people.

## Testing 

**TODO** @bvaughn and @acdlite will fill this in later.

sw-yx:
would this be a nice to have? https://github.com/facebook/react-devtools/issues/1141

bvaughn:
Not sure. DevTools is already monstrously large. Don't know how I feel about adding more panels and increasing scope until/unless we tackle something like #1214

bvaughn:
Fixes went out with 3.4.3

PR #1207, opened 2018-10-30
Support React.memo and timed-out Suspense trees

bvaughn:
# Support `React.memo`

* **Resolves** #1193
* **Commit** 4d85b12

Remove `React.pure` (since it was never released in a stable) and add support for `React.memo`. Preserve previous custom `displayName` semantics.

#### Code
```js
function CounterFunction({count}) {
  return <small>{count}</small>;
}
const A = React.memo(({count}) => <small>{count}</small>);
const B = React.memo(CounterFunction);
const C = React.memo(CounterFunction);
C.displayName = 'MemoCounterFunction';
```

#### DevTools
screen shot 2018-10-29 at 11 16 27 am

# Support timed-out suspense subtrees

* **Resolves** #1189.
* **Commit** 34ed7be

**Backstory**: [A special case (hack)](https://github.com/facebook/react/blob/8c67bbf183cc5ae1cab15a8265c612daf80cd86f/packages/react-reconciler/src/ReactFiberBeginWork.js#L1107-L1132) was added to fiber to maintain state for components within timed-out suspended trees. In this event, React hides the original UI and shows a fallback UI. Rather than destroying the hidden UI, it holds onto it so the state isn't lost (e.g. component state, untracked input values).

DevTools mostly just worked with this hack, but [use cases like this one](https://gist.github.com/bvaughn/476d0e54bb57870e7e384680a7a15056) left the Elements panel in a broken state after the suspended render completed– due to an incorrect child->parent pointer.

I'm not sure of a great way to model this within [`attachRendererFiber`](https://github.com/facebook/react-devtools/blob/master/backend/attachRendererFiber.js) (where it probably belongs), but it seemed reasonably easy to handle as a generic reparenting check. So let's discuss? 😄 

# Misc

* **Commit** ef8c92d

Also fixed the Elements panel display name for `IndeterminateComponent` type components.

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -40,6 +40,7 @@ function getInternalReactConstants(version) {
       HostPortal: 4,
       HostRoot: 3,
       HostText: 6,
+      IncompleteClassComponent: 17,
       IndeterminateComponent: 2,
       LazyComponent: 16,
       MemoComponent: 14,
@@ -63,6 +64,7 @@ function getInternalReactConstants(version) {
       HostPortal: 6,
       HostRoot: 5,
       HostText: 8,
+      IncompleteClassComponent: -1, // Doesn't exist yet
       IndeterminateComponent: 4,
       LazyComponent: -1, // Doesn't exist yet
       MemoComponent: -1, // Doesn't exist yet
@@ -86,6 +88,7 @@ function getInternalReactConstants(version) {
       HostPortal: 4,
       HostRoot: 3,
       HostText: 6,
+      IncompleteClassComponent: -1, // Doesn't exist yet
       IndeterminateComponent: 0,
       LazyComponent: -1, // Doesn't exist yet
       MemoComponent: -1, // Doesn't exist yet
@@ -137,12 +140,14 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     FunctionalComponent,
     ClassComponent,
     ContextConsumer,
+    Fragment,
+    ForwardRef,
     HostRoot,
     HostPortal,
     HostComponent,
     HostText,
-    Fragment,
-    ForwardRef,
+    IncompleteClassComponent,
+    IndeterminateComponent,
     MemoComponent,
     SimpleMemoComponent,
   } = ReactTypeOfWork;
@@ -197,8 +202,10 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     // TODO: Add support for new tags LazyComponent, MemoComponent, and SimpleMemoComponent
 
     switch (fiber.tag) {
-      case FunctionalComponent:
       case ClassComponent:
+      case FunctionalComponent:
+      case IncompleteClassComponent:
+      case IndeterminateComponent:
         nodeType = 'Composite';
         name = getDisplayName(resolvedType);
         publicInstance = fiber.stateNode;
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -199,8 +199,6 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
       }
     }
 
-    // TODO: Add support for new tags LazyComponent, MemoComponent, and SimpleMemoComponent
-
     switch (fiber.tag) {
       case ClassComponent:
       case FunctionalComponent:
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -106,10 +106,10 @@ function getInternalReactConstants(version) {
     CONTEXT_PROVIDER_SYMBOL_STRING: 'Symbol(react.provider)',
     FORWARD_REF_NUMBER: 0xead0,
     FORWARD_REF_SYMBOL_STRING: 'Symbol(react.forward_ref)',
+    MEMO_NUMBER: 0xead3,
+    MEMO_SYMBOL_STRING: 'Symbol(react.memo)',
     PROFILER_NUMBER: 0xead2,
     PROFILER_SYMBOL_STRING: 'Symbol(react.profiler)',
-    PURE_NUMBER: 0xead3,
-    PURE_SYMBOL_STRING: 'Symbol(react.pure)',
     STRICT_MODE_NUMBER: 0xeacc,
     STRICT_MODE_SYMBOL_STRING: 'Symbol(react.strict_mode)',
     SUSPENSE_NUMBER: 0xead1,
@@ -143,6 +143,8 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     HostText,
     Fragment,
     ForwardRef,
+    MemoComponent,
+    SimpleMemoComponent,
   } = ReactTypeOfWork;
   var {
     CONCURRENT_MODE_NUMBER,
@@ -154,8 +156,6 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
     CONTEXT_PROVIDER_SYMBOL_STRING,
     PROFILER_NUMBER,
     PROFILER_SYMBOL_STRING,
-    PURE_NUMBER,
-    PURE_SYMBOL_STRING,
     STRICT_MODE_NUMBER,
     STRICT_MODE_SYMBOL_STRING,
     SUSPENSE_NUMBER,
@@ -166,6 +166,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
   // TODO: we might want to change the data structure
   // once we no longer suppport Stack versions of `getData`.
   function getDataFiber(fiber: Object): DataType {
+    var elementType = fiber.elementType;
     var type = fiber.type;
     var key = fiber.key;
     var ref = fiber.ref;
@@ -279,6 +280,17 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
         nodeType = 'Wrapper';
         children = [];
         break;
+      case MemoComponent:
+      case SimpleMemoComponent:
+        nodeType = 'Special';
+        if (elementType.displayName) {
+          name = elementType.displayName;
+        } else {
+          const displayName = type.displayName || type.name;
+          name = displayName ? `Memo(${displayName})` : 'Memo';
+        }
+        children = [];
+        break;
       default:
         const symbolOrNumber = typeof type === 'object' && type !== null
           ? type.$$typeof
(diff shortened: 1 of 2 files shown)


gaearon:
Looks good except https://github.com/facebook/react-devtools/commit/34ed7bee3f96d788b3b7f98c4992d76e956ef5d7 which worries me. I think this makes DevTools too loose about what is allowed to happen. Can we model it as mounting and unmounting of two different sets of children instead? You should be able to tell from the Suspense state whether it's timed out or not (just like React already does), and then branch somewhere in `updateFiber` in `attachRendererFiber`. This would likely mirror custom reconciliation logic. WDYT?

bvaughn:
Regarding 34ed7be, could you expand on your concern about making DevTools "too loose"? If this newly added code is ever executed, it means that React reparented a node– right? Is there any way it could trigger a false positive? (I'm not totally confident about this change either, just pushing back a little to make sure the initial concern is well founded.)

> Can we model it as mounting and unmounting of two different sets of children instead?

This is kind of what I wanted to do initially, but I struggled to think of a way to detect this case in [`attachRendererFiber`](https://github.com/facebook/react-devtools/blob/master/backend/attachRendererFiber.js), since it isn't stateful.

Maybe I could detect it with a combination of `fiber.tag` === suspense tag, plus `memoizedState.didTimeout`, plus `hasChildOrderChanged` but...that felt clunkier than my proposed solution in `Store`.

I'll give it another pass though.

bvaughn:
> Can we model it as mounting and unmounting of two different sets of children instead?

Another (small) downside of this is that unmounting and remounting would cause the profiler's "renders" count to be reset.

gaearon:
>If this newly added code is ever executed, it means that React reparented a node– right?

Or it could also mean that there is a bug in the `attachRendererFiber` traversal code. (I caught a few several times when I made changes in the past and I worry new behavior wouldn't have caught my bugs because it would just think we're doing reparenting.)

Since technically we don't support reparenting anywhere else I'm not sure it's worth it to add generic support for it. I'd rather see a targeted fix around this specific use case.

bvaughn:
Thinking about this more, I think it's actually _desirable_ to model this as a re-parent, since that's actually what's happening.

bvaughn:
I'm also a little anxious about adding custom logic in `updateFiber` for this specific Suspense case, since I'm under the impression that it's a hack that we may not always do.

Even though we don't happen to support re-parenting in the general case, it feels like less of a hack to detect this case in a more suspense-and-version-agnostic way.

bvaughn:
Related: facebook/react/pull/14065

bvaughn:
Here's a shitty, hacky way of not showing the timed out sub tree as thought it committed in the Profiler:
```diff
diff --git a/backend/attachRendererFiber.js b/backend/attachRendererFiber.js
index f3b72d5..3384b19 100644
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -170,7 +170,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
 
   // TODO: we might want to change the data structure
   // once we no longer suppport Stack versions of `getData`.
-  function getDataFiber(fiber: Object): DataType {
+  function getDataFiber(fiber: Object, isHidden: boolean = false): DataType {
     var elementType = fiber.elementType;
     var type = fiber.type;
     var key = fiber.key;
@@ -192,6 +192,10 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
     var treeBaseDuration = null;
     var memoizedInteractions = null;
 
+    // Flags a Suspense subtree as temporarily hidden.
+    // We do this so that it isn't shown in the Profiler.
+    var isTimedOutSuspense = false;
+
     var resolvedType = type;
     if (typeof type === 'object' && type !== null) {
       if (typeof type.then === 'function') {
@@ -287,7 +291,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
         break;
       case MemoComponent:
       case SimpleMemoComponent:
-        nodeType = 'Special';
+        nodeType = 'Composite';
         if (elementType.displayName) {
           name = elementType.displayName;
         } else {
@@ -342,6 +346,10 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
             name = 'Suspense';
             props = fiber.memoizedProps;
             children = [];
+
+            // If this is a timed-out Suspense component, React will have hidden the first subtree.
+            // We need to mark this, so that tree doesn't appear in Profiler as though it rendered.
+            isTimedOutSuspense = fiber.memoizedState != null && fiber.memoizedState.didTimeout;
             break;
           case PROFILER_NUMBER:
           case PROFILER_SYMBOL_STRING:
@@ -395,6 +403,10 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
       actualDuration,
       actualStartTime,
       treeBaseDuration,
+
+      // Suspense
+      isTimedOutSuspense,
+      isHidden,
     };
   }
 
@@ -491,10 +503,10 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
     }
   }
 
-  function enqueueMount(fiber) {
+  function enqueueMount(fiber, isHidden) {
     pendingEvents.push({
       internalInstance: getOpaqueNode(fiber),
-      data: getDataFiber(fiber),
+      data: getDataFiber(fiber, isHidden),
       renderer: rid,
       type: 'mount',
     });
@@ -502,14 +514,14 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
     const isRoot = fiber.tag === HostRoot;
     if (isRoot) {
       pendingEvents.push({
-        internalInstance: getOpaqueNode(fiber),
+        internalInstance: getOpaqueNode(fiber, isHidden),
         renderer: rid,
         type: 'root',
       });
     }
   }
 
-  function enqueueUpdateIfNecessary(fiber, hasChildOrderChanged) {
+  function enqueueUpdateIfNecessary(fiber, hasChildOrderChanged, isHidden = false) {
     if (
       !hasChildOrderChanged &&
       !hasDataChanged(fiber.alternate, fiber)
@@ -521,7 +533,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
       if (haveProfilerTimesChanged(fiber.alternate, fiber)) {
         pendingEvents.push({
           internalInstance: getOpaqueNode(fiber),
-          data: getDataFiber(fiber),
+          data: getDataFiber(fiber, isHidden),
           renderer: rid,
           type: 'updateProfileTimes',
         });
@@ -530,7 +542,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
     }
     pendingEvents.push({
       internalInstance: getOpaqueNode(fiber),
-      data: getDataFiber(fiber),
+      data: getDataFiber(fiber, isHidden),
       renderer: rid,
       type: 'update',
     });
@@ -565,7 +577,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
     });
   }
 
-  function mountFiber(fiber) {
+  function mountFiber(fiber, isHidden = false) {
     // Depth-first.
     // Logs mounting of children first, parents later.
     let node = fiber;
@@ -575,7 +587,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
         node = node.child;
         continue;
       }
-      enqueueMount(node);
+      enqueueMount(node, isHidden);
       if (node == fiber) {
         return;
       }
@@ -586,7 +598,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
       }
       while (node.return) {
         node = node.return;
-        enqueueMount(node);
+        enqueueMount(node, isHidden);
         if (node == fiber) {
           return;
         }
@@ -600,21 +612,32 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
     }
   }
 
-  function updateFiber(nextFiber, prevFiber) {
+  function updateFiber(nextFiber, prevFiber, isHidden) {
+    const isTimedOutSuspense = (
+      nextFiber.tag === ReactTypeOfWork.SuspenseComponent &&
+      nextFiber.memoizedState &&
+      nextFiber.memoizedState.didTimeout
+    );
+
     let hasChildOrderChanged = false;
     if (nextFiber.child !== prevFiber.child) {
+      let isFirstChild = true;
       // If the first child is different, we need to traverse them.
       // Each next child will be either a new child (mount) or an alternate (update).
       let nextChild = nextFiber.child;
       let prevChildAtSameIndex = prevFiber.child;
       while (nextChild) {
+        // React hides the first child of a timed-out Suspense component.
+        const isChildHidden = isHidden || isTimedOutSuspense && isFirstChild;
+        isFirstChild = false;
+
         // We already know children will be referentially different because
         // they are either new mounts or alternates of previous children.
         // Schedule updates and mounts depending on whether alternates exist.
         // We don't track deletions here because they are reported separately.
         if (nextChild.alternate) {
           const prevChild = nextChild.alternate;
-          updateFiber(nextChild, prevChild);
+          updateFiber(nextChild, prevChild, isChildHidden);
           // However we also keep track if the order of the children matches
           // the previous order. They are always different referentially, but
           // if the instances line up conceptually we'll want to know that.
@@ -622,7 +645,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
             hasChildOrderChanged = true;
           }
         } else {
-          mountFiber(nextChild);
+          mountFiber(nextChild, isChildHidden);
           if (!hasChildOrderChanged) {
             hasChildOrderChanged = true;
           }
@@ -640,7 +663,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
         hasChildOrderChanged = true;
       }
     }
-    enqueueUpdateIfNecessary(nextFiber, hasChildOrderChanged);
+    enqueueUpdateIfNecessary(nextFiber, hasChildOrderChanged, isHidden);
   }
 
   function walkTree() {
diff --git a/plugins/Profiler/ProfileCollector.js b/plugins/Profiler/ProfileCollector.js
index d802f08..b538d2e 100644
--- a/plugins/Profiler/ProfileCollector.js
+++ b/plugins/Profiler/ProfileCollector.js
@@ -91,7 +91,11 @@ class ProfileCollector {
       return;
     }
 
-    this._committedNodes.add(data.id);
+    // Don't count suspended, hidden components as part of the render.
+    if (!data.isHidden) {
+      this._committedNodes.add(data.id);
+    }
+
     this._maxActualDuration = Math.max(this._maxActualDuration, data.actualDuration);
   };
```

I'm going to hold off on committing things in hopes that something better will occur to me.

gaearon:
Note `fiber.memoizedState.didTimeout` was removed from state in https://github.com/facebook/react/pull/14083.

gaearon:
Related PR: https://github.com/notifications

bvaughn:
Thanks for the pointer to 14083, Dan. Will need to chat with @acdlite and figure out a new heuristic for this I guess.

bvaughn:
I think I'm going to go ahead and land this change as-is and circle back about the hidden subtree. I chatted with Andrew about the reparenting change. He's not familiar enough with DevTools to say definitively but didn't have any objections to the general concept.

0xdevalias:
Curious if we know when this will hit release? Seems it isn't in `3.4.2`?
* https://github.com/facebook/react-devtools/compare/3.4.2...master

Issue #1209, opened 2018-10-31
Improved DevTools release/regression testing


bvaughn:
Regression testing is an important part of updating and releasing DevTools, but it has gotten increasingly difficult as we've added advanced features to React (e.g. async, suspense, hooks). Even some of the recent tag constant changes (e.g. #1103, #1190) complicate this testing process, as they introduce potentially significant differences between minor releases of React that require explicit testing.

- [ ] **Improved automated tests**

Despite its large feature set, DevTools currently has only **4 test suites** containing a total of **16 tests**. We should invest some more effort here. Specifically, I'd like to see us add tests that protect against regressions. I think we could use a similar approach as with [reactjs/react-lifecycles-compat](https://github.com/reactjs/react-lifecycles-compat), where we run tests against [multiple versions of React](https://github.com/reactjs/react-lifecycles-compat/tree/master/react).

Specifically, I think there might be some low-effort value to be gained by writing a simple test for each version of React that feeds DevTools an element tree and then verifyies the expected return values (e.g. name, type, props, state).

- [x] #1210: **Improved manual test harness** 

Our ["plain test shell" example app](https://github.com/facebook/react-devtools/blob/master/test/example/target.js) can be great for testing a change to the DevTools UI, but it isn't all that helpful in terms of regression testing. The fact that it requires compilation, and that the compiled source code is committed to the Git repo, is also irritating.

I think we could create a couple of similar test harnesses, perhaps one for each React version 14, 15, and 16, _don't_ require compilation. Instead these test harnesses could load a UMD version of `react` and `react-dom` (maybe selectable via a drop-down that controls a URL parameter). Each test harness could progressively enhance its features based on the semver minor tag (e.g. 16.6.x+ would have a test for `.memo`).

This way we'd have a centralized place to sanity check changes prior to releasing a new version.

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #1210, opened 2018-10-31
Added manual test harnesses for React 16.0-16.7

bvaughn:
Related to issue #1209

Here's a proposal for basic manual test harness to regression test React minor releases for versions 16.0-16.7. (We can add support for older v14 and v15 versions too if we'd like.)

The v16 test app progressively adds React features based on the current minor version so that we can share code between various minor versions.

To use the test harness:
```sh
# Cross origin requests are not supported for the file protocol.
# Since the HTML files share a single script file, use http:
pushstate-server .

# View all of the test versions via iframes to test DevTools:
open http://localhost:9000/test/regression/
```

The test page will look like this:
screen shot 2018-11-01 at 9 20 15 am

And DevTools should show all of the versions loaded in their various iframes:
screen shot 2018-11-01 at 9 15 52 am


bvaughn:
Ooh, I think I can make this better using an iframe...

PR #1218, opened 2018-11-07
Resolve defaultProps for lazy etc

gaearon:
Related to https://github.com/facebook/react/pull/13902.

If `fiber.elementType` exists and differs from `fiber.type`, we need to respect the `defaultProps` defined on the `fiber.type`. Such default props wouldn't be kept in `fiber.memoizedProps`. This is why in the React code we also call `resolveDefaultProps()` in those cases.

I added a regression case to demonstrate it. It should display `<FooWithDefaultProps greeting="Hello" name="World" />`. In master, it is missing `name="World"` (that’s the issue I am fixing).

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -17,6 +17,22 @@ var semver = require('semver');
 var copyWithSet = require('./copyWithSet');
 var getDisplayName = require('./getDisplayName');
 
+// Taken from ReactElement.
+function resolveDefaultProps(Component: any, baseProps: Object): Object {
+  if (Component && Component.defaultProps) {
+    // Resolve default props. Taken from ReactElement
+    const props = Object.assign({}, baseProps);
+    const defaultProps = Component.defaultProps;
+    for (const propName in defaultProps) {
+      if (props[propName] === undefined) {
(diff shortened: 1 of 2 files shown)


backend/attachRendererFiber.js
```diff
@@ -354,6 +370,14 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
         break;
     }
 
+    if (
+      props !== null &&
+      typeof fiber.elementType !== undefined &&
+      fiber.type !== fiber.elementType
+    ) {
+      props = resolveDefaultProps(fiber.type, props);
```
gaearon: I could've put this into a more specific place, like `Function` + `Class` + `ForwardRef` + `Memo` or whichever ones support it. But I found this easier to read. The only way you could have a false positive here is if somebody assigned `defaultProps` to e.g. a context consumer. I think that's unlikely enough that it would be fine for DevTools to mistakingly resolve them for a type that doesn't support them — to keep the code readable.

bvaughn: This seems safe. I think any time `fiber.type !== fiber.elementType`, it's a lazy component.

PR #1219, opened 2018-11-07
Profiler should properly display hidden, suspended subtrees

bvaughn:
The behavior of timed-out `Suspense` trees is unique. Rather than unmount the timed out content (and possibly lose important state) React re-parents this content within a hidden `Fragment` while the fallback is showing.

This behavior can cause problems DevTools though. Specifically it messes with the profiler (showing components as though they rendered when they didn't) and the Elements tab (possibly losing selected/collapsed state for hidden components).

After trying several approaches (see commit history for this PR) the easiest fix seems to be to just strip out the intermediate `Fragment` fibers in the timed-out case. This is done in the `attachRendererFiber` file, to avoid any special case code leaking into other parts of DevTools.

### Expanded/selected state is maintained when a tree is hidden
kapture 2018-11-10 at 10 51 03

### Hidden trees don't look like they re-rendered in the profiler
kapture 2018-11-10 at 10 51 32

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -186,7 +186,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
 
   // TODO: we might want to change the data structure
   // once we no longer suppport Stack versions of `getData`.
-  function getDataFiber(fiber: Object): DataType {
+  function getDataFiber(fiber: Object, isHidden: boolean = false): DataType {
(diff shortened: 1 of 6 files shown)


bvaughn:
Chatting with @sebmarkbage about this problem yesterday, he suggested I skip over the `Fragment` nodes that wrapped around the hidden subtree and the fallback UI. After giving this a shot, I think it's probably a bit nicer than the previous two approaches.

PR #1220, opened 2018-11-07
Use request-supplied host and port in connectToDevTools

duncanbeevers:
### What changed
When using the standalone react devtools the `connectToDevtools` call passes no arguments. The fallback behavior in this case is for the devtools instance to try and connect at `localhost`.

This PR changes the behavior so that if a `host` header is supplied in the initial request, then the `connectToDevtools` call will use the header host and port values.

### Motivation
This simplifies development when the devtools server is running on one host and the client is running on another. Simply dropping in the devtools-provided script snippet should be sufficient to get the two halves communicating successfully.

### Caveats
The host-header checking code is pretty simplistic, and not robust to weird or malformed host headers. However, so long as execution passes those host-header checks, no regressions should be introduced to the current arg-less `connectToDevtools` behavior.


bvaughn:

PR #1224, opened 2018-11-09
Remove Relay classic plugin and any reference to Relay

avijohnson:
With the deprecation, of Relay Classic it no longer makes sense to maintain this plugin.
cc @bvaughn 
[Testing]
I follow the existing instructions for test in Chrome, Firefox and standalone, making sure a few smoke tests passed when using code that still uses Relay classic and React. Below are some screenshots from the console.

screen shot 2018-11-08 at 2 15 16 pm
screen shot 2018-11-08 at 2 42 51 pm
screen shot 2018-11-08 at 2 48 52 pm


bvaughn:
😍 

screen shot 2018-11-09 at 1 35 16 pm

bvaughn:
I've published packed versions of the Chrome and Firefox with support for Relay Classic (the last release that will have them) here https://github.com/facebook/react-devtools/releases/tag/3.4.2

PR #1226, opened 2018-11-12
Bug: handle "replacement" connections correctly

david-davidson:
Hi, and thanks for this amazing tool!

I noticed that the standalone package has [logic intended to handle cases where a new devtools connection "replaces" a previous one](https://github.com/facebook/react-devtools/blob/0228043b3b07a9dcc54269c859d11b6f0efeef9d/packages/react-devtools-core/src/standalone.js#L122-L128). However, it's not quite working right:

repro

(^ Two instances of Create React App pointing at the same `localhost:8097` standalone instance; second instance fails to connect)

### Problem

It looks like there are two things going on:

1.) The [onclose](https://github.com/facebook/react-devtools/blob/0228043b3b07a9dcc54269c859d11b6f0efeef9d/packages/react-devtools-core/src/standalone.js#L135) callback from the _previous_ (replaced) connection seems to fire asynchronously. That is, these steps take place in this order:
* We detect a replacement connection and [close it](https://github.com/facebook/react-devtools/blob/0228043b3b07a9dcc54269c859d11b6f0efeef9d/packages/react-devtools-core/src/standalone.js#L135)
* We [initialize](https://github.com/facebook/react-devtools/blob/0228043b3b07a9dcc54269c859d11b6f0efeef9d/packages/react-devtools-core/src/standalone.js#L140) the new connection, and [re-render](https://github.com/facebook/react-devtools/blob/0228043b3b07a9dcc54269c859d11b6f0efeef9d/packages/react-devtools-core/src/standalone.js#L46-L52)
* Then the `onclose` callback of the previous connection detects that it's been closed, and tries to shut down. But by now, `connected` is the _new_ connection (so we don't want to [set it to false](https://github.com/facebook/react-devtools/blob/0228043b3b07a9dcc54269c859d11b6f0efeef9d/packages/react-devtools-core/src/standalone.js#L136)), and we've already set up the DOM (so we don't want to [clear it with `onDisconnected`](https://github.com/facebook/react-devtools/blob/0228043b3b07a9dcc54269c859d11b6f0efeef9d/packages/react-devtools-core/src/standalone.js#L137))

Here's that sequence via breakpoints:

sequence

2.) Over in `backend.js`, we have logic to [schedule retries 2 seconds after a connection closes](https://github.com/facebook/react-devtools/blob/0228043b3b07a9dcc54269c859d11b6f0efeef9d/packages/react-devtools-core/src/backend.js#L90). If we solely make changes to accommodate 1.), we end up in a looping state where both connections keep retrying ad infinitum.

### Solution

In both cases, the code needs to be aware when a connection is being closed because of an incoming replacement. When it is, we can skip the usual `onClose` steps (the incoming connection will have already set everything up) and the scheduled retry.

This PR implements that approach by passing a string to the socket's `close` method (in cases of replacement) and checking it in both relevant event handlers.

### DIFF
--- a/packages/react-devtools-core/src/standalone.js
+++ b/packages/react-devtools-core/src/standalone.js
@@ -19,6 +19,7 @@ var Panel = require('../../../frontend/Panel');
 var launchEditor = require('./launchEditor');
 var React = require('react');
 var ReactDOM = require('react-dom');
+var { CONNECTION_REPLACED } = require('./backend');
 
 var node = null;
 var onStatusChange = function noop() {};
@@ -114,13 +115,14 @@ function connectToSocket(socket) {
   };
 }
 
+var WS_DEFAULT_CLOSE_CODE = 1000;
 function startServer(port = 8097) {
   var httpServer = require('http').createServer();
   var server = new ws.Server({server: httpServer});
   var connected = false;
   server.on('connection', (socket) => {
     if (connected) {
-      connected.close();
+      connected.close(WS_DEFAULT_CLOSE_CODE, CONNECTION_REPLACED);
       log.warn(
         'Only one connection allowed at a time.',
         'Closing the previous connection'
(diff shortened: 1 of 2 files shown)


packages/react-devtools-core/src/standalone.js
```diff
@@ -114,13 +115,14 @@ function connectToSocket(socket) {
   };
 }
 
+var WS_DEFAULT_CLOSE_CODE = 1000;
```
david-davidson: (per https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent#Properties)

bvaughn:

Issue #1214, opened 2018-11-02
Discussion: Re-architect DevTools


bvaughn:
## The problem
React DevTools currently support all "recent" versions of React (14+? 13+?). As the React API grows and changes, this will become increasingly difficult to maintain. Several recent changes to React core have broken parts of DevTools (e.g. the profiler) without us even being aware.

DevTools exists as a separate project because it supports multiple versions of React. However this makes automated testing and type-checking more difficult (see #1209).

## Possible solution
1. Extract most of the current "backend" and "frontend" code pieces into a new package, e.g. `react-devtools-interface`, and move this package to the React repo..
1. Release an initial version of `react-devtools-interface` that supports all of the currently supported versions of React (14+? 13+?).
1. Remove all legacy code from `react-devtools-interface` so that it only supports a single version of React (e.g. the version it corresponds to in Git).
1. Reduce the `react-devtools` project to be a basic shell that detects which version(s) of React are currently present, and loads the corresponding `react-devtools-interface` package(s) at runtime<sup>1</sup>.

<sup>1: Actual runtime loaded code seems to be frowned upon by both Google and Mozilla, but we could bundle the needed versions of `react-devtools-interface` along with the shell extension and have it swap between them based on the injected React version(s).</sup>

### Benefits
1. Size and complexity of DevTools drastically decreases.
1. The cost of refactoring the React API and its internals decreases as a result.
1. Moving `react-devtools-interface` to the React repo would us to have automated integration tests and Flow type checks.
1. Reducing complexity in the DevTools interface likely also results in improved performance.
1. This offers us a nice opportunity to address longstanding tech debt in some older sections of DevTools.

### Drawbacks
1. Using DevTools with multiple versions of React would result in a larger download size for the extension. (We should be able to easily remain below the [200 MB size limit](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/Distribution/Submitting_an_add-on) though.)

### Open questions
- [ ] How would we support bug fixes (or version-agnostic additions) to DevTools if this required multiple packages to be updated and released? (Could we automated this somehow?)
- [x] ~~How would we support multiple versions of React on a single page?~~ The shell wrapper would require some minimal UI for its loading state anyway. Maybe this UI could also handle multiple React versions by prompting you to select which version you want to inspect (and enable you to toggle this version later somehow)?
[...]

Issue #1090, opened 2018-08-10
Refactor theme to use CSS variables


bvaughn:
We currently use context to pass around the active `Theme` and is very heavy-handed. It forces use to use [factory functions for otherwise static CSS styles](https://github.com/facebook/react-devtools/blob/4d3ae51dc31617bd813339fb7baa00e0154a0023/frontend/TreeView.js#L144-L150) and to [force-deep-render the app when the current theme changes](https://github.com/facebook/react-devtools/blob/4d3ae51dc31617bd813339fb7baa00e0154a0023/frontend/Panel.js#L355).

In hindsight, I should have used [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables) for this. Fortunately, it should be easy to gradually migrate components over to using CSS variables instead of context. All we need to do is update the variables whenever a theme is loaded, and then we can gradually remove context references in favor of CSS variables over time.

infinnie:
What about the new context API combined with styled components?

bvaughn:
I don't think context is necessary. CSS variables would let us do this in a way that totally avoids the problem of re-rendering. It also simplifies components as they no longer need to worry about a `theme` param.

I'm not convinced that styled-components buys us anything that plain styles + CSS doesn't, and at least it used to be that the lib came with a significant performance cost. The performance thing might have changed. I've heard that the latest release made significant performance gains. But I don't personally find the API compelling in the case of DevTools.

Maybe you could put together a prototype that shows how you'd use SC in a specific DevTools panel– maybe showing a side by side comparison of inline styles versus SC– and make an argument for it?

bvaughn:
PR #1236 updates our infra to support CSS modules. They aren't strictly necessary for this change– (we could use CSS variables with inline styles)– but I think the combination of the two really cuts down on a lot of boilerplate and "feels" better.

bvaughn:
React DevTools has been rewritten and [recently launched a new version 4 UI](https://reactjs.org/blog/2019/08/15/new-react-devtools.html). The source code for this rewrite was done in a separate repository and now lives in the main React repo ([github.com/facebook/react](https://github.com/facebook/react/)).

Because version 4 was a total rewrite, and all issues in this repository are related to the old version 3 of the extension, I am closing all issues in this repository. If you can still reproduce this issue, or believe this feature request is still relevant, please open a new issue in the React repo: https://github.com/facebook/react/issues/new?labels=Component:%20Developer%20Tools

PR #1235, opened 2018-11-21
Upgrade dev build tooling (Babel 6 -> 7, Webpack 1 -> 4)

bvaughn:
Our build tooling was super outdated, which made maintenance kind of a chore. I think this step, while not totally necessary, will help with issues like #1214 and #1090.

## Build sizes
I also decided (at least unless someone knows of a good reason _not_ to do this) to stick with Webpack's default minification behavior– which is to say, do it for production builds. This resulted in the following built size reductions:

#### Standalone:
* backend.js: 215KB -> 97KB
* standalone.js: 1.3MB -> 601KB

#### Browser extension:
* packed and zipped extension: 1.8MB -> 840KB
* backend.js: 395KB -> 152KB
* panel.js: 1.3MB -> 606KB

## Build times
I timed `yarn build:standalone` and `yarn build:extension:chrome` in master and this branch and didn't notice any real difference in durations. (The duration varies between runs but seems to be about the same overall.)

## Testing
I've tested the example app, plain shell, Firefox and Chrome extensions, standalone DevTools, and the theme viewer. All seem okay after this change.

I also updated the backend/integration Webpack config, although I'm pretty sure that test hasn't been used in a long time– since the source has an invalid require (referencing a file that was moved a while back).


test/example/target.js
```diff
@@ -78,7 +78,7 @@ class Todos extends React.Component {
     todos.some(item => {
       if (item.id === id) {
         item.completed = completed;
-        text = item.text;
+        text = item.title;
```
bvaughn: This happened to be broken before. I noticed while testing the new config.

backend/integration/webpack.config.js
```diff
@@ -29,23 +32,14 @@ module.exports = {
...
-      test: /\.js$/,
-      loader:  'babel',
-      exclude: /node_modules/,
-    }],
+    rules: [
+      {
+        test: /\.js$/,
+        loader: 'babel-loader',
+        options: {
+          ...JSON.parse(readFileSync(resolve(__dirname, '../../.babelrc'))),
```
hugmanrique: Is this spread operator really needed? You could remove the object braces and set `options` to the result of JSON.parse

Maybe I'm missing something

packages/react-devtools-core/webpack.backend.js
```diff
@@ -9,38 +9,37 @@
...
-      test: /\.js$/,
-      loader: 'babel',
-      exclude: /node_modules/,
-    }],
+    rules: [
+      {
+        test: /\.js$/,
+        loader: 'babel-loader',
+        options: {
+          ...JSON.parse(readFileSync(resolve(__dirname, '../../.babelrc'))),
```
hugmanrique: Same thing as above

backend/integration/webpack.config.js
bvaughn: Originally, I thought I might add some custom options to individual configs but that turned out not to be necessary.

bvaughn:
I am going to land this. I'll take responsibility for anything that pops up as a result and fix it with a follow-up PR. I think the more interesting PR to get feedback on is #1236 which this soft blocks.

PR #1236, opened 2018-11-21
Css modules

bvaughn:
Prototype of #1090. Builds on top of #1235 to migrate two files (`SettingsPane` and `ChartNode`) away from using the "theme" `context` and to using CSS variables instead.

The `SettingsPane` component now uses CSS modules. `ChartNode` on the other hand continues to use inline styles. I don't know if I have a strong preference for one approach over the other– I just want to get rid of the theme context and thought it would be worth trying both approaches.

This PR shows a net increase in lines-of-code, but this is due to the `yarn.lock` file and additional CSS module Webpack loader config. As for the component code itself, **so much code can be removed by doing this** (compare [before](https://github.com/facebook/react-devtools/blob/master/frontend/SettingsPane.js) and [after](https://github.com/facebook/react-devtools/blob/8d7e380fbfa51192471b0c1bb0b8e74325195f5e/frontend/SettingsPane.js)).

I think we should gradually migrate views to use the new approach because it would dramatically simplify the process of updating and maintaining DevTools UI.

Incremental migration should be easy, since themes will continue to be passed down via context _as well as_ written to CSS variables. (A nice end goal would be to remove the context entirely. 😄)


frontend/SettingsPane.css
```diff
@@ -0,0 +1,87 @@
+.SettingsPane {
+  display: flex;
+  flex-wrap: wrap;
+  flex-shrink: 0;
+  align-items: center;
+  position: relative;
+  background-color: var(--base01);
+  border-bottom: 1px solid var(--base03);
```
bvaughn: Should these vars use a `theme` prefix (e.g. `--theme-based01` or `--theme-base-01`)?

frontend/SettingsPane.css
```diff
@@ -0,0 +1,87 @@
...
+.SearchInputWrapper {
+  position: relative;
+  display: flex;
+  flex-grow: 1;
+  flex-shrink: 0;
+  align-items: center;
+}
+
+.ClearSearchButton {
+  font-size: var(--sansSerif-size-large);
```
bvaughn: Is this a reasonable naming convention for our shared font values?

bvaughn:
This PR gives me a "Spring cleaning" feeling. It was kind of annoying updating the styles, but once it was done it felt **so nice**.

bvaughn:
I pushed one more example with da81414. This time I didn't use an external CSS file– but I did switch the inline styles for `ChartNode` to use CSS vars instead of the theme context, and was able to remove a lot unnecessary `theme` props from its ancestors.

I'm not totally set on using CSS modules for styling here. I just think we should git rid of the theme context.

bvaughn:
I feel good about this. I don't think it's pressing enough to drop everything and rip out the context theme, but I think it's something that would be nice to do here and there, iteratively.

PR #1248, opened 2018-12-12
Fix infinite retry loop

msand:
(Together with a pr to react-native https://github.com/facebook/react-native/pull/22625)
Fixes https://github.com/facebook/react-native/issues/21030

### DIFF
--- a/packages/react-devtools-core/src/backend.js
+++ b/packages/react-devtools-core/src/backend.js
@@ -66,7 +66,7 @@ function connectToDevTools(options: ?ConnectOptions) {
   // See D6251744.
   var ws = websocket ? websocket : new window.WebSocket(uri);
   ws.onclose = handleClose;
-  ws.onerror = handleClose;
+  ws.onerror = handleFailed;
   ws.onmessage = handleMessage;
   ws.onopen = function() {
     var wall = {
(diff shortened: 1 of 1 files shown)


msand:
This pr isn't intended to be merged as is, merely to point out where the logic needs to be improved and as a quick temporary override if people want to monkeypatch things.

msand:
Counting the number of consecutive errors without a single successful connection attempt and stopping after three or something, would probably already be better than a infinite retry loop.

msand:
@gaearon @bvaughn What do you think of this? This and https://github.com/facebook/react-native/pull/22625 are affecting the DX of react-native developers outside facebook quite significantly, as their logs inside xcode get several reconnection attempt failures per second, hiding any real errors in an ocean of noise. And, even making their computers crash (not just xcode) because of excessive memory use from it.

msand:
The other PR has been merged and released in react-native 0.58-stable (v0.58.6) and 0.59-stable (v0.59.0-rc.3). Could we get this merged as well. I've cherry picked this on top of 3.4.2 and 3.6.0 of react-devtools and made build commits available of react-devtools-core on github, and given instructions how to avoid the issue here: https://github.com/facebook/react-native/issues/21030
But getting this merged and possibly even released with react-native 0.59.0 would be ideal.
@bvaughn @gaearon Would you either one of you have time to look at this?

msand:
I think this issue is giving quite a bad impression for both newcomers to react-native, and long term developers. As the logs just keep filling up with errors about connection issues. Even if devtools and the inspector are working perfectly fine. Seems like the default setup for a clean react-native project shouldn't fill up the Xcode logs when developing for iOS. Or is there something missing in the configuration/documentation?

bvaughn:
3.6.1 has been published to NPM with this fix.

msand:
Excellent, thanks!

Issue #212, opened 2015-09-16
Editing props should trigger componentWillReceiveProps


terinjokes:
Hey react-devtools team:

I've noticed recently that modifying props in the DevTool's sidebar does not seems to respect the component lifecycle events: it instead sets the props directly and calls "render".

In particular, the lifecycle event "componentWillReceiveProps" is not called. This is officially documented:

> Use this as an opportunity to react to a prop transition before `render()` is called by updating the state using `this.setState()`.

This is easy to work around: I can wrap my component in another component that passes in all props, and interact with this wrapper from DevTools. Then my component will get notified of prop changes. However this breaks my concentration, and forces me to leave and go do another thing before I can continue debugging.

I'm curious why things are done the current way and if it will be possible to change DevTools to reflect the more expected outcome.

jaredly:
The current implementation was merely the simplest way possible :) You [can see here](https://github.com/facebook/react-devtools/blob/master/backend/getData.js#L110) what's happening -- it's just doing "forceUpdate" after modifying the props. I don't imagine it would be difficult to add in a "componentWillReceiveProps" call there, and perhaps that would be enough.
Do you want to try that out and see how it works?

terinjokes:
@jaredly yea, i'll give it a shot and let you know.

deanc:
Just got caught out by this. It would be a *great* addition to make it respect all parts of the lifecycle!

terinjokes:
For the record, I didn't get a chance to work on this.

sighrobot:
I will second the request for this to be fixed - I just spent over an hour trying to figure out why `componentWillReceiveProps` wasn't being called when I changed `props` in the React Dev Tool sidebar, and ended up having to use the wrapper approach described by @terinjokes. Unfortunately that approach is unwieldy for testing more complex components. You'd have to duplicate all child `props` as parent `props` or `state` in order to fully experiment with the child component.

Very confused why this isn't documented anywhere more obvious than within this issue, especially since the behavior of the plugin in this case deviates so substantially from one of the fundamental aspects of the library with which it interacts. Many people are going to needlessly scratch their heads on this one.

rolandfung:
Just ran into this myself, too. Fixing this would be great and make testing some of my more complex components easier.

gaearon:
There is no easy solution here IMO because “setting props” completely circumvents React lifecycle. I don’t think calling `componentWillReceiveProps()` from DevTools is appropriate here.

[...]

PR #1249, opened 2018-12-12
Support editable props for more component types

bvaughn:
Use new injected `overrideProps()` function (available in DEV builds only) to support editable props for function components, host elements, and special types like `memo` and `forwardRef`.

Move this update logic into React will also enable us to add more complex behavior in the future– such as the ability to override props in a way that won't get overridden unless the component is re-rendered with a _new_ props value.

## Demo
kapture 2018-12-12 at 15 40 18

Since there is no `forceUpdate` function for e.g. function components or host nodes, I also removed the `forceUpdate` attribute from the `CompositeUpdater` type. It wasn't being used for anything other than determining if a node supported editing, so I replaced it with a `canUpdate` boolean prop.

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -150,6 +150,7 @@ function getInternalReactConstants(version) {
 }
 
 function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): Helpers {
+  var {overrideProps} = renderer;
   var {ReactTypeOfWork, ReactSymbols, ReactTypeOfSideEffect} = getInternalReactConstants(renderer.version);
   var {PerformedWork} = ReactTypeOfSideEffect;
   var {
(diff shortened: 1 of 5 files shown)


bvaughn:
Published a pre-release build here https://react-devtools.now.sh/ if anyone is interested in giving it a spin

backend/attachRendererFiber.js
```diff
@@ -239,8 +250,8 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
         const inst = publicInstance;
         if (inst) {
           updater = {
+            canUpdate: true,
             setState: inst.setState && inst.setState.bind(inst),
-            forceUpdate: inst.forceUpdate && inst.forceUpdate.bind(inst),
             setInProps: inst.forceUpdate && setInProps.bind(null, fiber),
```
Kovensky: Would not overriding `setInProps` if `updater && updater.setInProps` fix https://github.com/facebook/react-devtools/issues/212 by using `scheduleWork` instead of `forceUpdate`?

bvaughn: Try and see?

PR #1256, opened 2018-12-30
Fix: Better display for string contexts

### LINKED ISSUE #1212: Better display for string contexts
When using `<SomeContext.Provider value="string">` and consuming that with the new `contextType` from React 16.6.0, the devtools displays the context value like an array:

```
0: "s"
1: "t"
2: "r"
3: "i"
4: "n"
5: "g"
```

It would be nicer if it was displayed something like this:

```
"string"
```

Test page:

```html
<!doctype html>
<meta charset="utf-8">
<title>React Context test</title>
<div id="container"></div>
<script src="https://unpkg.com/react@16.6.0/umd/react.development.js"></script>
<script src="https://unpkg.com/react-dom@16.6.0/umd/react-dom.development.js"></script>

<script>
  const LanguageContext = React.createContext();

  class Test extends React.Component {
    render() {
      return React.createElement("p", {}, `Language: ${this.context}`);
    }
  }

  Test.contextType = LanguageContext;

  ReactDOM.render(
    React.createElement(
      LanguageContext.Provider,
      { value: "en-us" },
      React.createElement(Test),
    ),
    container
  );
</script>
```

Screenshot:

screenshot from 2018-11-01 15-23-03

yifei-fu:
Closes #1212.
DataView now accepts any type of data (not just objects) and displays primitive values properly.
image

### DIFF
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -24,7 +24,7 @@ type Inspect = (path: Array<string>, cb: () => void) => void;
 type ShowMenu = boolean | (e: DOMEvent, val: any, path: Array<string>, name: string) => void;
 
 type DataViewProps = {
-  data: ?Object,
+  data: ?any,
   path: Array<string>,
   inspect: Inspect,
   showMenu: ShowMenu,
@@ -74,9 +74,20 @@ class DataView extends React.Component<DataViewProps> {
       return <div style={missingStyle(theme)}>null</div>;
     }
 
+    const dataType = typeof data;
+    var isPrimitive = dataType === 'number' || dataType === 'string' || dataType === 'boolean';
     var isArray = Array.isArray(data);
     var elements = [];
-    if (isArray) {
+    if (isPrimitive) {
+      elements.push(<DataItem
+          name={''}
+          hideName={true}
+          path={this.props.path}
+          inspect={this.props.inspect}
+          showMenu={this.props.showMenu}
+          readOnly={this.props.readOnly}
+          value={data} />);
+    } else if (isArray) {
       // Iterate over array, filling holes with special items
       var lastIndex = -1;
       data.forEach((item, i) => {
@@ -148,6 +159,7 @@ type Props = {
   noSort?: boolean,
   readOnly?: boolean,
   name: string,
+  hideName?: boolean,
   value: any,
 };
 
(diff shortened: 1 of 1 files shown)


frontend/DataView/DataView.js
```diff
@@ -74,9 +74,20 @@ class DataView extends React.Component<DataViewProps> {
       return <div style={missingStyle(theme)}>null</div>;
     }
 
+    const dataType = typeof data;
+    var isPrimitive = dataType === 'number' || dataType === 'string' || dataType === 'boolean';
```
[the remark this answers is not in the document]
yifei-fu: Found that as well. I'm pretty sure it's unrelated to the changes because numbers are not displayed when testing on the master branch. I'll probably open an issue for that.

frontend/DataView/DataView.js
```diff
@@ -74,9 +74,20 @@ class DataView extends React.Component<DataViewProps> {
...
     }
 
+    const dataType = typeof data;
+    var isSingleValue = dataType === 'number' || dataType === 'string' || dataType === 'boolean';
     var isArray = Array.isArray(data);
     var elements = [];
-    if (isArray) {
+    if (isSingleValue) {
+      elements.push(<DataItem
+          name={''}
```
bvaughn: Even though this is a single value, since we're adding it to an array– it needs a `key`.

frontend/DataView/DataView.js
```diff
@@ -74,9 +74,20 @@ class DataView extends React.Component<DataViewProps> {
...
+          hideName={true}
+          path={this.props.path}
+          inspect={this.props.inspect}
+          showMenu={this.props.showMenu}
+          readOnly={this.props.readOnly}
```
bvaughn: This case should be read-only, because we can't actually edit the context value without crawling upwards and mutating the ancestor context. (We _can_ override the value on the instance, but React will reset it when we call `forceUpdate` so the change won't actually stick.)

frontend/DataView/DataView.js
```diff
@@ -74,9 +74,20 @@ class DataView extends React.Component<DataViewProps> {
       return <div style={missingStyle(theme)}>null</div>;
     }
 
+    const dataType = typeof data;
+    var isSingleValue = dataType === 'number' || dataType === 'string' || dataType === 'boolean';
```
bvaughn: I kind of prefer "simple type" rather than "single value" but this is just a subject nit

Issue #1200, opened 2018-10-28
Large amount of data in component props causes devtools to peg CPU


vaughnkoch:
Problem:
If a React component has a large amount of data passed in props, even by reference, the CPU gets pegged after the component renders.

Repro
- Create an array or object structure with a large amount of data, for example 10-20MB. 
- Pass the array variable as the props of a component.
- Render the component.

This produces the following error:

```
Uncaught Error: Message length exceeded maximum allowed length.
    at PortImpl.postMessage (VM27 extensions::messaging:121)
    at Port.publicClassPrototype.(anonymous function) [as postMessage] (extensions::utils:138:26)
    at handleMessageFromPage (contentScript.js:24)
```

The bottom function is this:
```
    function handleMessageFromPage(evt) {
        evt.source === window && evt.data && "react-devtools-bridge" === evt.data.source && port.postMessage(evt.data.payload);
    }
```

Here's the current implementation of postMessage which is currently stringifying all the props for the component. Even if they're passed by reference, JSON.stringify still has to process all the data. 

```
  PortImpl.prototype.postMessage = function(msg) {
    if (!$Object.hasOwnProperty(ports, this.portId_))
      throw new Error(kPortClosedError);

    // JSON.stringify doesn't support a root object which is undefined.
    if (msg === undefined)
      msg = null;
    msg = $JSON.stringify(msg);
    if (msg === undefined) {
      // JSON.stringify can fail with unserializable objects. Log an error and
      // drop the message.
      //
      // TODO(kalman/mpcomplete): it would be better to do the same validation
      // here that we do for runtime.sendMessage (and variants), i.e. throw an
      // schema validation Error, but just maintain the old behavior until
      // there's a good reason not to (http://crbug.com/263077).
      console.error('Illegal argument to Port.postMessage');
      return;
    }
    var error = messagingNatives.PostMessage(this.portId_, msg);
    if (error)
      throw new Error(error);
  };
```

Notes/Thoughts
- A large amount of data in props passed by ref can be a legitimate use case, for example using Vega for dataviz. Usually this isn't a problem since they're passed by ref.
- Are there any other ways to serialize arbitrary JS objects instead of JSON.stringify, for example using Apache Arrow?
- If the props are only used for display to the user in the inspect or profiler ui, would it make sense to just send the first 10-20 elements in an array and tell the user that the data was clipped? Not sure how you'd handle this with arbitrarily large objects though.

vaughnkoch:
[...]

PR #1258, opened 2019-01-02
Serialize nested attributes less aggressively

bvaughn:
Related to #1200 

In the current version of DevTools, the bridge serializes object attributes for a few levels– after which, attributes that are more deeply nested are lazily serialized when the frontend requests them.

This PR makes the bridge _lazier_ by reducing the nested threshold from 2 to 1– enough to drill into props (e.g. `props.object`) but no further (e.g. `props.object.foo`) unless/until a user clicks to expand the value.

In my testing, this improves performance _a lot_ for some of the more extreme poor performance test harnesses. For example, here are the perf profiles for opening the DevTools extension on a fresh tab of [the treeherder repro](https://github.com/facebook/react-devtools/issues/1200#issuecomment-437346433)):

#### Before
screen shot 2019-01-02 at 11 02 28 am

#### After
screen shot 2019-01-02 at 11 03 25 am

The numbers above vary a bit by run, but the extension consistently feels much more responsive after these changes.

### DIFF
--- a/agent/dehydrate.js
+++ b/agent/dehydrate.js
@@ -10,6 +10,8 @@
  */
 'use strict';
 
+const LEVEL_THRESHOLD = 1;
+
 /**
  * Get a enhanced/artificial type string based on the object instance
  */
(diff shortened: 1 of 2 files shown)


agent/dehydrate.js
```diff
@@ -10,6 +10,8 @@
  */
 'use strict';
 
+const LEVEL_THRESHOLD = 1;
```
gaearon: Maybe leave a comment here about perf implications? e.g. "This used to be set to 2 but it caused too much overhead in large apps and we decided to set it to 1 instead"

bvaughn:
One downside of this change (that I just realized) is that this threshold indirectly determines how far you are able to drill into props for a commit within the Profiler.

I think it's still the right change to make though. That constraint was already in place, and although this makes it "worse" it improves performance of the DevTools overall.

bvaughn:
Unfortunately I think I'm going to have to back this change out because of #1262 – at least until/unless I can think of a way to serialize more aggressively for profiling data, which would require architectural changes since all bridge traffic is treated equal.

PR #1260, opened 2019-01-02
Disable props/state inspection for Profiler

bvaughn:
Follow up for something I noticed while testing #1258

Inspecting props/state within the Profiler never worked right because React mutates fibers, so older "snapshots" of props/state _can't_ be inspected. However the UI seemed to suggest that props _could_ be expanded– which was confusing.

This changes things to make the UI more intuitive. For example, the `state.object` object below doesn't show a carrot or mouse pointer cursor for inspecting.

Here's the Elements panel style:
screen shot 2019-01-02 at 2 53 57 pm

And here's the Profiler panel style:
screen shot 2019-01-02 at 2 54 10 pm

I also changed the `DataView` component displayed inside of the profiler to expand objects by default up until the point where they're dehydrated, to offset the fact that we don't allow you to toggle them expanded/collapsed any more.

### DIFF
--- a/frontend/DataView/DataView.js
+++ b/frontend/DataView/DataView.js
@@ -26,7 +26,7 @@ type ShowMenu = boolean | (e: DOMEvent, val: any, path: Array<string>, name: str
 type DataViewProps = {
   data: ?any,
   path: Array<string>,
-  inspect: Inspect,
+  inspect: Inspect | null,
   showMenu: ShowMenu,
   startOpen?: boolean,
   noSort?: boolean,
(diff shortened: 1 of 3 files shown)


frontend/DataView/DataView.js
```diff
@@ -195,25 +195,31 @@ class DataItem extends React.Component<Props, State> {
   }
 
   inspect() {
+    if (this.props.inspect === null) {
+      return;
```
gaearon: Might be worth commenting why this is useful. Someone later might look at this code and think: "I'll just simplify this by passing an empty function!"

PR #1261, opened 2019-01-03
Fix Context.Consumer displayName for v16.6+

### LINKED ISSUE #1251: Inconsistent displayName configuration for context Consumer/Provider
Due to this change after the original PR:

https://github.com/facebook/react-devtools/commit/c2db8dd5c13edd29444c72714b49cdb1073a1a44

This is now needed to set the displayName for both as for `Consumer` we do not look up the `displayName` on `_context` like we do for `Provider`:

```js
TestContext.displayName = 'TestContext';
TestContext.Consumer.displayName = 'TestContext';
```

Was this intentional or can I submit a PR to fix it?

bvaughn:
Resolves #1251

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -358,9 +358,14 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
           case CONTEXT_CONSUMER_SYMBOL_STRING:
             nodeType = 'Special';
             props = fiber.memoizedProps;
+
+            // v16.3-16.5 read from fiber.type (because the Consumer is the full context object)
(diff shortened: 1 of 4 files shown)


backend/attachRendererFiber.js
```diff
@@ -358,9 +358,14 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
           case CONTEXT_CONSUMER_SYMBOL_STRING:
             nodeType = 'Special';
             props = fiber.memoizedProps;
+
+            // v16.3-16.5 read from fiber.type (because the Consumer is the full context object)
+            // v16.6+ should read from fiber.type._context (because Consumer is a separate object)
```
gaearon: "is a separate object in DEV" maybe

PR #1263, opened 2019-01-03
Increase Bridge serialization threshold from 1 -> 2

### LINKED ISSUE #1262: Interaction labels are broken in master
While testing the upcoming release of DevTools, I discovered that interaction labels in the Profiler are broken in master:
screen shot 2019-01-03 at 10 27 58 am

For comparison, here's the latest release:
screen shot 2019-01-03 at 10 28 23 am

bvaughn:
This basically reverts #1258. It is being done to fix interaction names within the Profiler.

Resolves #1262

### DIFF
--- a/agent/__tests__/dehydrate-test.js
+++ b/agent/__tests__/dehydrate-test.js
@@ -33,31 +33,31 @@ describe('dehydrate', () => {
     var object = {a: {b: {c: {d: 4}}}};
     var cleaned = [];
     var result = dehydrate(object, cleaned);
-    expect(cleaned).toEqual([['a', 'b']]);
-    expect(result.a.b).toEqual({type: 'object', name: '', meta: {}});
-    expect(result.a.b.c).toBeUndefined(); // Dehydrated
+    expect(cleaned).toEqual([['a', 'b', 'c']]);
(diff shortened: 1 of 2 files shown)


gaearon:
Can we fix just the profiler stuff alone somehow?

bvaughn:
> Can we fix just the profiler stuff alone somehow?

Requests sent through the Bridge are batched and treated pretty generically.

Maybe the profiler could proactively "hydrate" interactions during the commit phase so that we serialize a little more information about them? This would slow down the commit phase though.

I can give it a little more thought before merging this I guess.

bvaughn:
After thinking about this a little more, I don't think there's a great alternative (that would not involve making more significant changes to the bridge architecture).

I considered trying to explicitly hydrate the root fiber's `memoizedInteractions` set during the commit phase (when profiling), but I think this wouldn't work– since the `ProfilerStore` assumes things can be committed _synchronously_ but bridge hydration is _asynchronous_.

I also briefly considered trying to special case the serialization threshold for root fibers and/or the interactions `Set` somehow but I'm not sure how that could be cleanly accomplished with the current layers of abstraction.

PR #1264, opened 2019-01-03
Fix createContext runtime error for 16.3.0

### LINKED ISSUE #1168: Cannot read property 'displayName' of undefined
Hi guys! I upgrade react-devtools to 3.4.0 version and catch an error:

**Cannot read property 'displayName' of undefined in backend.js** 
image
Before 3.4.0 I had 3.3.2 and it was ok.
So, I can't use tab react-devtools, because that have waiting status:

**Waiting for roots to load...**

Browser: Ya.Browser v 18.9.0.3363 (64-bit)

bvaughn:
Resolves #1168


bvaughn:
> 😐

I know 😦

Issue #1266, opened 2019-01-04
undefined is not a function(near .....reactDevTools.connectToDevTools. ) issue


shravanteegala:
![screen shot 2019-01-04 at 2 19 06 pm](https://user-images.githubusercontent.com/13646977/50679978-c36f4c00-102b-11e9-94f3-0d19716ce95c.png)

gokulnathperiasamy:
Any alternate way or workaround ?

shravanteegala:
i fix using below command.
npm install --save-dev react-devtools-core@3.4.2 Resolved the issue

herodrigues:
This was broken in latest release `v3.5.0`. Roll back to `v3.4.3` and everything should be fine.

bvaughn:
Hi all 👋 Sorry for the trouble 😦

I believe this was something I broke when upgrading the React DevTools project's Babel and Webpack versions in https://github.com/facebook/react-devtools/pull/1235, and we're just now encountering it because we haven't released the `react-devtools-core` package since then.

I just published `react-devtools-core` v3.5.1. Can anyone let me know if this resolves the issue? (I'm unable to repro the original reported issue so it's hard for me to verify the fix.)

bvaughn:
Someone on the other thread (https://github.com/facebook/react-native/issues/22863) reported that the 3.5.1 release has fixed their problem, so I'm going to close this issue for now!

herodrigues:
@bvaughn It's working now :+1:

Issue #1265, opened 2019-01-04
react-devtools 3.5.0 breaks react-native 0.57


sunnylqm:
https://github.com/facebook/react-native/issues/22863

bharatidudhjiya:
I have same issue with react native 0.57

herodrigues:
It was fixed in the latest patch release `v3.5.1`

PR #1267, opened 2019-01-04
Fix commonjs exports

vovkasm:
Fixed: #1266 and facebook/react-native#22863

### DIFF
--- a/packages/react-devtools-core/webpack.backend.js
+++ b/packages/react-devtools-core/webpack.backend.js
@@ -24,6 +24,7 @@ module.exports = {
   output: {
     path: __dirname + '/build',
     filename: '[name].js',
+    libraryTarget: 'commonjs2',
   },
   plugins: __DEV__ ? [] : [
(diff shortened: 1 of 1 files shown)


vovkasm:
Also same issue: #1265

vovkasm:
Also the tests passed I'm not sure is it right fix in common. Yes it fix RN issue, but should this module be available to direct inclusion from browser environment? May be better solution to split builds and make distinct builds for various targets (node, browser, es modules) and set appropriate entry points in package.json?
What people think? If this is the case, I can rework patch.

bvaughn:
Hm, it seems like this was something I missed/broke when upgrading Webpack 1 -> 4 (PR #1235) and we're just noticing it now because we haven't released `react-devtools-core` since then.

The `libraryTarget` used to be "umd" before that change.

bvaughn:
Just published v3.5.1 with the "umd" change– which seems to fix this issue. So thanks again!

Issue #1269, opened 2019-01-04
react-devtools 3.5.1 breaks standalone mode for Chrome Extensions


jamaljsr:
I upgraded to v3.5.1 today and now I cannot inspect my Browser Extension through Chrome as I did previously. See [usage-with-react-dom](https://github.com/facebook/react-devtools/tree/master/packages/react-devtools#usage-with-react-dom)

The console displays the error:
```
Uncaught ReferenceError: ReactDevToolsBackend is not defined
```

I've found what appears to be the problem. `ReactDevToolsBackend` is hard coded in [standalone.js](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools-core/src/standalone.js#L154). The global exported by webpack was changed to `backend` in the recent commits v3.5.0 [28d36a22](https://github.com/facebook/react-devtools/commit/28d36a228de7ff8d72665f3c6a1b8f11ad7e94b6#diff-0da7ccbc75d9b6430b9a287fe1232b1b) and v3.5.1 [bcf6b1aa](https://github.com/facebook/react-devtools/commit/bcf6b1aaf3e15b67b5c797f95c639c810c13db32#diff-0da7ccbc75d9b6430b9a287fe1232b1b).

To fix this locally, I updated [standalone.js](https://github.com/facebook/react-devtools/blob/master/packages/react-devtools-core/src/standalone.js#L154) replacing
```
res.end(backendFile + '\n;ReactDevToolsBackend.connectToDevTools();');
```
with
```
res.end(backendFile + '\n;backend.connectToDevTools();');
```

bvaughn:
Hm...I am also unable to launch connect the standalone DevTools to Chrome, but not for the reason you mention. The error I'm getting is:

> Error: ENOENT: no such file or directory, open '/build/backend.js'

bvaughn:
> I upgraded to v3.5.1 today

Could you clarify _what_ you upgraded? Was it `react-devtools` or `react-devtools-core` (or both)?

bvaughn:
Gotcha! Thanks for confirming.

jamaljsr:
I upgraded ‘react-devtools’ to v3.5.1

bvaughn:
Yeah, makes sense now that I know you were also seeing the other error. Was confusing before. 😄

bvaughn:
A fix has been published to NPM as 3.5.2.

bvaughn:
[...]

PR #1271, opened 2019-01-07
Fixed standalone+DOM integration

bvaughn:
Resolves issue #1269

I guess this was broken by the recent Webpack upgrade (#1235) and we didn't notice because it's not a common use case.

Things I tested:
* Local standalone (`yarn build:standalone && yarn test:standalone`)
* NPM install (although this required editing the nested `node_modules/react-devtools-core` to copy over local changes)
* Standalone + RN (shouldn't be impacted by this change)

### DIFF
--- a/packages/react-devtools-core/src/standalone.js
+++ b/packages/react-devtools-core/src/standalone.js
@@ -11,7 +11,6 @@
 
 var ws = require('ws');
 var fs = require('fs');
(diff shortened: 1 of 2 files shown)


bvaughn:
Since this is already pretty broken 😄 I'm just going to merge and release the fix. Happy to make follow up adjustments later.

PR #1272, opened 2019-01-09
MVP hooks support

bvaughn:
Support (read-only) inspectable hooks.

kapture 2019-01-11 at 11 07 44-hooks

# Open questions

### How should we handle production mode?
How should we handle hooks inspection in production mode? The `react-debug-hooks` package relies on parsing function names to identify custom hooks. If these names are mangled, they won't be identifiable, and as a result– the hooks tree will look strange.

#### Development:
screen shot 2019-01-11 at 8 58 00 am

#### Production:
screen shot 2019-01-11 at 8 58 09 am


bvaughn:
No, we aren't currently planning on surfacing that info in DevTools.

TryingToImprove:
Given I have component like below:

```
const C = () => {
   const [person, setPerson] = useState({ name: 'test' });
   const [schools, setSchools] = useState(['test', 'test2]);

   return (<div>{person} went to {schools.join(',')}</div>)
```

In devtools it will display

```
- Hooks
   - State: person
   - State: schools
```

So there will be mutiple "state"-objects in the list if I do not use "custom hooks".

Is what I am describing correct?

bvaughn:
Yes, because you are using multiple state hooks in that example. They'll be listed in the order you've used them in, and they'll show the values currently returned.

Saifadin:
@TryingToImprove most likely it will look like this

```
- Hooks
  - State: {
      name: 'test'
    }
  - State: [
      'test', 'test2'
    ]
```

bvaughn:
I should clarify my previous comment.

Yes, you will see multiple "State" hooks– because you are calling `useState` multiple times in that example. They'll be listed in the order you've used them in, and they'll show the values currently returned.

So with the code example you provided, it would actually look like this:
```
Hooks
  State: {…}
  State: Array[2]
```

bvaughn:
Okay, I think this is ready for review.

agent/Agent.js
```diff
@@ -96,22 +98,34 @@ class Agent extends EventEmitter {
...
+          this.global.$r = data.publicInstance;
+          lastSelected = data.publicInstance;
+        }
+        if (data.containsHooks) {
+          inspectedHooks = this.updateHooksTree(id);
+        }
+      }
+      if (this._prevInspectedHooks !== inspectedHooks) {
+        this._prevInspectedHooks = inspectedHooks;
+        this.emit('inspectedHooks', inspectedHooks);
```
bvaughn: Doing this (storing data, having Fiber-specific code) in the Agent sucks. I'm open to suggestions for a better place for this logic to live, but at the same time– I didn't stress about it too much because I think this version of DevTools is short lived (without either a rewrite or a refactor).

agent/Agent.js
```diff
@@ -431,6 +467,37 @@ class Agent extends EventEmitter {
...
+    const data = this.elementData.get(id);
+    const internalInstance = this.internalInstancesById.get(id);
+    if (internalInstance) {
+      const rendererID = this.rendererIdsByInternalInstance.get(internalInstance);
+      if (rendererID) {
+        const internals = this.reactInternals[rendererID].renderer;
+        if (internals && internals.currentDispatcherRef) {
+          // HACK: This leaks Fiber-specific logic into the Agent which is not ideal.
+          // $FlowFixMe
+          const currentFiber = data.state === internalInstance.memoizedState ? internalInstance : internalInstance.alternate;
```
bvaughn: This is gross for obvious reasons, but it needs to live somewhere. Happy to move it out of the Agent if anyone thinks it's too gross even for a temporary solution.

gaearon:
Does it still show "undefined" for Hooks that don't have values? Is that necessary/useful? Can be just empty?

bvaughn:
> Does it still show "undefined" for Hooks that don't have values? Is that necessary/useful? Can be just empty?

Yeah, but I can probably just change it to empty string for custom hooks without labels. That seems nicer :)

backend/attachRendererFiber.js
```diff
@@ -215,6 +216,13 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer):
     // Suspense
     var isTimedOutSuspense = false;
 
+    // Hooks inspection
+    var containsHooks =
+      (tag === FunctionComponent ||
+        tag === SimpleMemoComponent ||
+        tag == ForwardRef) &&
```
ksaldana1: Are you wanting to use `==` for the `FowardRef` comparison here?

bvaughn: Typo. Doesn't actually matter in this case but I'll fix it.

frontend/PropState.js
```diff
@@ -251,4 +270,9 @@ const noPropsStateStyle = (theme: Theme) => ({
   padding: '0.5rem',
 });
 
+const hooksTreeLoadingStyle = {
```
threepointone: any reason this isn't in a stylesheet? (low pri, ignore)

bvaughn: Just following the (super gross) convention within the file. Old DevTools are pretty gross.

bvaughn:
Whoops, I just realized that I need to update the profiler props panel to make sure we're displaying hooks correctly there too. 😅 Gimme a few.

gaearon:
>Good call. I added a guard for this.

Just to confirm. If I specifically call `useDebugValue(undefined)` I'd want to see it though. There's a difference between wanting to set `undefined` as a value, and not having a value.

bvaughn:
> Just to confirm. If I specifically call useDebugValue(undefined) I'd want to see it though.

There is no way for DevTools to differentiate between the two states of "I explicitly called `useDebugValue(undefined)`" and "I didn't call `useDebugValue`" so I chose to optimize for the case I think is far more common.

bvaughn:
I moved the fiber+hooks specific hacks inside of `Agent` into a new `HooksInspector` plugin because they felt pretty dirty, but with that... I think we're good enough for an MVP. Anything else that crops up– I'll fix it in the future! :)

Issue #1142, opened 2018-09-02
$r doesn't highlight Stateless Functional Components


mgol:
Typing `$r` in the console prints the nearest ancestor component that is not a stateless functional one.

Test case: https://github.com/mgol/react-tic-tac-toe

When highlighted in the React DevTools, only the top-level stateful `Game` component is displayed when `$r` is typed, no matter which one is selected.

bvaughn:
`$r` is a convenience pointer that will refer to either the currently selected DOM element _or_ the currently selected component _instance_. Since functional components have no instance, there's nothing for it to point to.

So basically, to me, it sounds like this is working as designed.

mgol:
Thanks for the explanation.

One issue is that React DevTools display ` == $r` next to the selected JSX node, whether it's a component instance or a DOM element. Having `$r` returns something completely different to what the UI claims is confusing. Especially that now I see it doesn't return the ancestor component but just whatever "working" thing `$r` pointed to previously. I guess this is the same what #929 reports. Wouldn't returning an object like:
```js
{
	displayName: 'Board',
	props: {/* Board props */}
}
```
be a good idea, even if it has no direct internal presence? If you want to keep existing behavior I would keep the ` == $r` mark next to the last selected element so that it's accurate.

Issue #929, opened 2017-10-24
$r points to wrong component, doesn't update on select


skidding:
**Minimal reproducing project: https://github.com/react-cosmos/react-cosmos.** Not minimal in number of deps, sorry! It will take a minute or two for yarn to bootstrap deps. But minimal in number of steps: **Run `yarn start`.**

Before diving in, here is what I'm experiencing

![devtools-bug](https://user-images.githubusercontent.com/250750/31947419-9203f6ca-b8d4-11e7-8a25-7f7c291508f3.gif)

`$r` always points to the same component (not a root nor a leaf), regardless of what I select.

Possible area of conflict: The Cosmos Playground has two frames, the top one with the menu and the iframe with the loaded components. They come from different bundles with different copies of React in each, and the inner frame attaches `__REACT_DEVTOOLS_GLOBAL_HOOK__` to the parent window, which is why they both show up in the React tab. As far as the React tab is concerned, components from both bundles show up and their correct info is shown when selected. Only problem is `$r` doesn't point to the selected component in the Console.

Let me know if there's anything else I can provide or if you have suggestions for what to debug on my end. Thanks!

RocketPuppy:
I'm seeing this as well on an app developed locally, not in Cosmos. `$r` always points to the first thing selected and does not update. React 16, React Dev Tools extension 2.5.2, chrome 61.0.3163.79

RocketPuppy:
Digging into the __REACT_DEVTOOLS_GLOBAL_HOOK__ it seems as if very few of my components have a public instance. This runs when a component is selected in the React tab:
```
	    _this.on('selected', function (id) {
	      var data = _this.elementData.get(id);
	      if (data && data.publicInstance && _this.global.$r === lastSelected) {
	        _this.global.$r = data.publicInstance;
	        lastSelected = data.publicInstance;
	      }
	    });
```
And `data.publicInstance` always seems to be `null` for components I've written, but is present for native DOM components and library components. I'm also using Typescript.

michelmattos:
It seems to happen only with functional components. If with change your component to be a stateful one, then you will be able to select it (and `$r` will work fine). Maybe `data.publicInstance` is being set to `null` because functional components don't have an instance?

RocketPuppy:
That makes sense @michelmattos! My project is almost all functional components.

bvaughn:
Closed by #1281

Issue #645, opened 2017-04-28
How does $r work with functional components?


gaearon:
I noticed an oddity: accessing the `($r in the console)` message is displayed for functional components too, and I get an internal representation as a result:

<img width="955" alt="screen shot 2017-04-28 at 1 45 46 pm" src="https://cloud.githubusercontent.com/assets/810438/25529197/01e458d4-2c19-11e7-98a2-0ac3417a0c3f.png">

This will likely break with Fiber.

Should we just expose a dummy object with `props` in this case?
I’m not sure.

fletcher91:
This difference between components seems to have caused some confusion (https://github.com/facebook/react-devtools/issues/929 https://github.com/facebook/react-devtools/issues/1142 https://github.com/facebook/react-devtools/issues/366#issuecomment-280312516), so it'd seem useful from a community perspective.

Our developer tools use this feature as well, to display extra information (e.g. all available data of a resource) based on the currently selected component.

Upgrading those components to use hooks broke those, so I can see the confusion grow once hooks reach stable and get wide-spread deployment.

gaearon:
I think currently `$r` is ignored for functions at all which is super annoying. Ideally it should maybe show props? Or something like `{ props, hooks }` where `hooks` is the introspected Hooks state.

fletcher91:
The latter seems more consistent to me, imagine anticipating whether to type `$r.propName` or `$r.props.propName` depending on the component.

It'd be extensible as well, with hooks as you say, either literal or via some mapping between hooks and their (virtual, flat?) `state` and `context` counterparts (though that might be hard since they're essentially nameless)

gaearon:
For now I think we'll set `$r` to elements like we did for host nodes (https://github.com/facebook/react-devtools/pull/619). This should be a quick fix. Later we can reconsider and figure something out for state.

bvaughn:
Closed by #1281

PR #1281, opened 2019-01-17
Expose meaningful $r as fallback for function components

gaearon:
Fixes https://github.com/facebook/react-devtools/issues/1142, https://github.com/facebook/react-devtools/issues/645, https://github.com/facebook/react-devtools/issues/929. This is a minimally invasive fix.

Before:

screen shot 2019-01-17 at 9 55 52 pm

screen shot 2019-01-17 at 9 55 49 pm

After:

screen shot 2019-01-17 at 9 56 24 pm

Kinda works with Hooks too:

screen shot 2019-01-17 at 9 56 49 pm

For Hooks state, this exposes our internal implementation but I think it's better than nothing if you're debugging. We can later improve it to flatten the list into an array or even show a tree representation.

I made it a fallback and not just for functions in case you want to inspect props of some exotic type. I see nothing wrong with that.

(I also used this as an opportunity to make it work for text nodes too.)

### DIFF
--- a/backend/attachRendererFiber.js
+++ b/backend/attachRendererFiber.js
@@ -322,6 +322,7 @@ function attachRendererFiber(hook: Hook, rid: string, renderer: ReactRenderer): 
         break;
       case HostText:
         nodeType = 'Text';
+        publicInstance = fiber.stateNode;
         text = fiber.memoizedProps;
         break;
       case Fragment:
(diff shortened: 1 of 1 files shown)


Jessidhia:
How come I missed this this is brilliant :tada:

Gone will be the days where I literally go spelunking around __reactInternalFiber's `children` / `next` / `return`s to be able to look at a component's relevant fiber state

gaearon:
You and me, @Jessidhia, you and me.

gaearon:
(On the other hand, isn't it a nice primer to React internal data structure? 😆)

yfain:
I'm using version 3.6 of this extension in Chrome, but the issue is still there, e.g. $r is not defined for functional components. Is there a newer build of this extension that includes this fix?

bvaughn:
@yfain This version of the extension isn't getting a lot of focus right now, because we're trying to finish up and release the new one. I'd suggest just hopping on the beta train for that one 😄 

https://github.com/bvaughn/react-devtools-experimental#installation

PR #1299, opened 2019-02-09
Stop sending profile data when recording is off

ValentinH:
In https://github.com/facebook/react/issues/14574, I have reported an issue where the devtools can get really slow.
It happens when a lot of profiling data are sent via the bridge to the frontend.

This PR disables sending the profile data to the frontend when the profiler isn't recording.

I wasn't sure where to prevent the data from being sent and how to do it. Also, I'm not sure if not sending this data when not recording can break other features.

### DIFF
--- a/agent/Agent.js
+++ b/agent/Agent.js
@@ -96,6 +96,7 @@ class Agent extends EventEmitter {
   capabilities: {[key: string]: boolean};
   _updateScroll: () => void;
   _inspectEnabled: boolean;
+  _isRecordingProfile: boolean
 
   constructor(global: Object, capabilities?: Object) {
     super();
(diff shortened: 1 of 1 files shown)


bvaughn:
I think this change would break the profiler. Times are intentionally updated, per commit, for each Fiber that's changed. The profiler relies on the (previously-sent) times to be accurate when generating its graphs. (This is how the profiler can accurately draw trees even for fibers that didn't commit at all during a given profiler session.)

It's _mostly_ enough for React to gather profiling data so it has accurate base times, but if we stopped sending that data while the profiler wasn't active, we'd have to send it all in one bit dump (rather than just the delta) when the profiler was first started. If you'd be interested in making this change, I could review it.

The [DevTools rewrite I'm working on](https://github.com/bvaughn/react-devtools-experimental) keeps the source of truth for things on the page context (rather than in the frontend `Store` object) so it doesn't need to be sent across a bridge at all. This should make optimizations like this unnecessary. (It wouldn't be sending profiling data unless the frontend requested it.) Unfortunately it's still a work in progress.

ValentinH:
Thanks for this complete answer. Indeed, I didn't try to profile an app where only a small part of the tree changes. In my case, only sending the data when the profiler is recording was fine. I'll test more but your explanation makes sense (and I'm really new to this project so I don't have the overview).

Do you have any idea how to solve the issue I linked above?

bvaughn:
> Do you have any idea how to solve the issue I linked above?

A rewrite (or major refactor) that reduces the amount of bridge traffic. That's one of the major focuses of the rewrite I recently started.

I haven't had a chance to dig into the issue you linked to yet because I'm on the road this weekend visiting family.

ValentinH:
I've just tried this PR on an app where I have 3 components being rendered in an interval and another one (a material-ui Button) that isn't updated at all: I still see everything in the flamegraph. Here the example I used: https://2p9zwo598j.codesandbox.io.

Did I misunderstand something in your first message?

ValentinH:
https://github.com/bvaughn/react-devtools-experimental is quite advanced now and will fix this issue :)

PR #1329, opened 2019-04-12
Prevent invocation of getters/setters during Object inspection

saintmarina:
Currently, the code is using `val.__proto__[name]` to retrieve prototype values during Object inspection. Unfortunately, when dealing with getter and setters, this invokes the function itself, instead of referring to it. This can have side-effects, including crashes.

By using the property descriptors, we have clean access to the prototype values. Note that the code could be expanded to provide on-demand getter inspections similar to the Chrome devtools inspections.

Note: it would be preferable to use getOwnPropertyDescriptors() instead of getOwnPropertyNames() + getOwnPropertyDescriptor(), but flowcheck does not have the getOwnPropertyDescriptors function registered.

### DIFF
--- a/agent/Bridge.js
+++ b/agent/Bridge.js
@@ -436,8 +436,9 @@ class Bridge {
         if (isFn && (name === 'arguments' || name === 'callee' || name === 'caller')) {
           return;
         }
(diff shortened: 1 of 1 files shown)


bvaughn:

PR #1332, opened 2019-04-13
Fix crash when selected fiber isn't in snapshot

### LINKED ISSUE #1313: Moving between renders crashes if selected element doesn't exist
Moving between renders while an element is selected will crash the devtools if the selected element doesn't exist in the next render.

## Reproduction

1. Go to this sandbox: https://xz27s.codesandbox.io/ (forkable [here](https://codesandbox.io/s/react-dev-tools-issue-1313-xz27s)).
2. Start a React Devtools Profiler recording.
3. Click the <kbd>Toggle Popup</kbd> button two or more times (need multiple renders to move between).
4. Stop the recording.
5. Select the `Popup` (you need to be in a render where there is a `Popup` element).
6. Click an arrow to move to another render
7. :boom: React Devtools crashes :boom:

pshrmn:
Fixes #1313.

The extension was crashing because the `ProfilerFiberDetailPane` component was attempting to call `snapshotFiber.get` when `snapshotFiber` is `null`.

This PR moves the `snapshotFiber.get` calls and the content rendered when a `snapshotFiber` exists to its own component.

I tested this by running the plain shell both on `master` and with this change. The application crashed on `master` and worked in this branch.

### DIFF
--- a/plugins/Profiler/views/ProfilerFiberDetailPane.js
+++ b/plugins/Profiler/views/ProfilerFiberDetailPane.js
@@ -14,14 +14,10 @@ import type {Theme} from '../../../frontend/types';
 import type {Snapshot} from '../ProfilerTypes';
 
 import React, {Fragment} from 'react';
-import {monospace, sansSerif} from '../../../frontend/Themes/Fonts';
-import DataView from '../../../frontend/DataView/DataView';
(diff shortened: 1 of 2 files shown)


pshrmn:
Closing because I don't think that this version of the dev tools is receiving any updates and the issue is removed in the next version.

